PR #1, opened 2016-12-06
Added a basic travis configuration based on rust-everywhere.

SirVer:


SirVer:
This is the base: https://github.com/japaric/rust-everywhere. A standard starting point for CI integration for rust project. It has a ton of cool features (for example building release packages), but most cannot be used for our purposes.

PR #10, opened 2016-12-13
Ship compressed points to the client.

SirVer:
This required updating three, since only newer version support normalized buffers.

### DIFF
--- a/client/octree_viewer.ts
+++ b/client/octree_viewer.ts
@@ -20,22 +20,26 @@ const KEY_L = 'L'.charCodeAt(0);
 
 let VERTEX_SHADER = `
 uniform float size;
-attribute vec4 color;
+uniform float edgeLength;
+uniform vec3 min;
 
-varying vec4 v_color;  // 'varying' vars are passed to the fragment shader
+attribute vec3 color;
+
+varying vec3 v_color;
 
 void main() {
   v_color = color / 255.;   // pass the color to the fragment shader
-  gl_Position = projectionMatrix * modelViewMatrix * vec4(position, 1.0);
+  gl_Position = projectionMatrix * modelViewMatrix *
+    vec4(position * edgeLength + min, 1.0);
   gl_PointSize = size;
 }
 `;
 
 let FRAGMENT_SHADER = `
-varying vec4 v_color;
+varying vec3 v_color;
 
 void main() {
-  gl_FragColor = v_color;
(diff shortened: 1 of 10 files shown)


client/octree_viewer.ts
```diff
@@ -86,78 +95,146 @@ class NodeLoader {
...
         numBytesRead += 4;
-        const BytesPerPoint = 16;
-        const numPoints = numBytesInBlob / BytesPerPoint;
-        let points = new NodePoints(
-            new Float32Array(data, numBytesRead, numPoints * 3),
-            new Uint8Array(
-                data, numBytesRead + numPoints * 12, numPoints * 4));
+
+        const bytesPerCoordinate = view.getUint8(numBytesRead);
+        numBytesRead += 1;
```
wohe: Aren't these 4 lines equivalent to `numBytesRead += 4;`?

client/octree_viewer.ts
```diff
@@ -86,78 +95,146 @@ class NodeLoader {
...
+        new THREE.Vector3(
+            nodeRenderData.min.x + nodeRenderData.edgeLength,
+            nodeRenderData.min.y + nodeRenderData.edgeLength,
+            nodeRenderData.min.z + nodeRenderData.edgeLength));
+    geometry.boundingSphere = new THREE.Sphere(
+        new THREE.Vector3(
+            nodeRenderData.min.x + nodeRenderData.edgeLength / 2,
+            nodeRenderData.min.y + nodeRenderData.edgeLength / 2,
+            nodeRenderData.min.z + nodeRenderData.edgeLength / 2),
+        nodeRenderData.edgeLength / 2
```
wohe: This is a sphere contained inside the bounding box, right? And is this okay? Or should it contain the bounding box instead?

client/octree_viewer.ts
```diff
@@ -86,78 +95,146 @@ class NodeLoader {
...
+          case 1:
+            position = new Uint8Array(data, numBytesRead, numPoints * 3);
+            normalizePosition = true;
+            break;
+
+          default:
+            console.log("Invalid bytesPerCoordinate: ", bytesPerCoordinate);
+        }
+        numBytesRead += numPoints * bytesPerCoordinate * 3;
+        if (numBytesRead % 4 != 0) {
```
wohe: Isn't this padding unnecessary? Should the format be changed?

PR #11, opened 2016-12-15
Support slightly more of the PLY format.

### LINKED ISSUE #8: PLY parsing assumes (x,y,z,r,g,b) in binary format
It will happily consume other files, but run out of data to nom before getting points.

cc @yutakaoka

SirVer:
- Get rid of nom as parser, since I did not figure out how to properly
  parse the PLY header with it.
- Parse without copying. This buys ~40% performance.
- Support PLYs with different data types and optional color. This costs ~50%
  performance due to the parsing having more branches.

All in all, parsing PLY is now ~10% slower than before, but more robust
and has more features.

Fixes #8.


src/ply.rs
```diff
@@ -14,22 +14,90 @@
...
+            DataType::Float32 => 4,
+            DataType::Float64 => 8,
+        }
+    }
+
+    fn read<B: ByteOrder>(&self, slice: &[u8]) -> f32 {
+        match *self {
+            DataType::Int8 => (slice[0] as i8) as f32,
+            DataType::Uint8 => slice[0] as f32,
+            DataType::Int16 => B::read_u16(slice) as f32,
```
wohe: I think you meant `read_i16` here and `read_u16` below?

src/ply.rs
```diff
@@ -14,22 +14,90 @@
...
+            DataType::Uint16 => B::read_i16(slice) as f32,
+            DataType::Int32 => B::read_i32(slice) as f32,
+            DataType::Uint32 => B::read_u32(slice) as f32,
+            DataType::Float32 => B::read_f32(slice) as f32,
+            DataType::Float64 => B::read_f32(slice) as f32,
```
wohe: Not `read_f64`?

SirVer:
ptal.

src/ply.rs
```diff
@@ -56,165 +118,186 @@ struct Property {
...
+    };
+
+    // We align the buffer of this 'BufReader' to points, so that we can index this buffer and know
+    // that it will always contain full points to parse.
+    Ok((BufReader::with_capacity(num_bytes_per_point * 1024, file),
+        header["vertex"].count,
+        point_format))
 }
 
+
```
wohe: Remove extra empty line.

src/ply.rs
```diff
@@ -56,165 +118,186 @@ struct Property {
...
+    }
+
+    let mut num_bytes_per_point = 0;
+    let position_data_type = header["vertex"]["x"].data_type;
+    num_bytes_per_point += position_data_type.size_in_bytes() * 3;
+
+    let color_data_type = {
+        if vertex.has_property("red") && vertex.has_property("green") &&
+           vertex.has_property("blue") {
+            Some(vertex["red"].data_type)
```
wohe: Add a TODO that we might want to verify that the same type is used for red/green/blue, r/g/b, or x/y/z?

PR #24, opened 2017-05-26
Make PLY reading more general.

SirVer:
Switch PLY reading away from assuming that x, y, z are the same datatype
and consecutive in the file and - if there is color - that it comes
directly after the position. It also opens up the possibility of
understanding and using other less common properties in the PLY files,
like intensity or laser_id. But for now this still does not properly
skip unknown properties.

The new method uses a vector of function pointer that is constructed at
header parsing time. Each function in the vector parses one of the
property and advances the buffer pointer. In my benchmarks this is ~10%
faster than needing to branch for each point.

Tested by diffing a created octree against the same created by master.

### DIFF
--- a/src/ply.rs
+++ b/src/ply.rs
@@ -206,15 +206,15 @@ type ReadingFn = fn(nread: &mut usize, buf: &[u8], val: &mut Point);
 // The two macros create a 'ReadingFn' that reads a value of '$data_type' out of a reader, assigns
 // it to '$property' (e.g. 'position.x') of 'point' while casting it to the correct type. I did not
 // find a way of doing this purely using generic programming, so I resorted to this macro.
-macro_rules! create_and_return_reading_function {
+macro_rules! create_and_return_reading_fn {
     ($data_type:expr, $($property:ident).+, $size:ident, $num_bytes:expr, $reading_fn:expr) => (
         {
             $size += $num_bytes;
-            fn _read_func(nread: &mut usize, buf: &[u8], point: &mut Point) {
+            fn _read_fn(nread: &mut usize, buf: &[u8], point: &mut Point) {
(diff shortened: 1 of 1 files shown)


src/ply.rs
```diff
@@ -221,16 +200,83 @@ fn parse_header<R: BufRead>(reader: &mut R) -> Result<(Header, usize)> {
...
-    position_data_type: DataType,
-    // None, if the file does not contain color.
-    color_data_type: Option<DataType>,
+
+type ReadingFunc = fn(nread: &mut usize, buf: &[u8], val: &mut Point);
+
+macro_rules! read_casted_property {
+    ($data_type:expr, point. $($property:ident).+, $size:ident ) => (
+        match $data_type {
+            DataType::Uint8 => {
```
wohe: Pull out a macro for the repetition?

src/ply.rs
```diff
@@ -246,60 +292,75 @@ fn open(ply_file: &Path) -> Result<(BufReader<File>, i64, PointFormat)> {
...
+            "r" | "red" => {
+                readers.push(read_casted_property!(prop.data_type, point.r, num_bytes_per_point));
+            }
+            "g" | "green" => {
+                readers.push(read_casted_property!(prop.data_type, point.g, num_bytes_per_point));
+            }
+            "b" | "blue" => {
+                readers.push(read_casted_property!(prop.data_type, point.b, num_bytes_per_point));
+            }
+            other => println!("Ignoring property '{}' on 'vertex'.", other),
```
wohe: This doesn't seem to actually ignore data, but rather read malformed data. Add a reader that ignores or panic instead?

SirVer:
@wohe ptal?

PR #25, opened 2017-05-26
Performance improvements

SirVer:
- Use an internal iterator instead of a boxed iterator. This is better semantics, since we do not plan to support iterating over the same octree in parallel. It is also faster since it allows for inlining of the callback function at each call site. This buys ~11% performance.
- Use an enum to represent the file input data stream since the number of input formats is finite.
- Activate link time optimization for release builds. This buys another 4%, but increases link time.


SirVer:
@wohe: let's wait till the base branch is merged, I will rebase then.

PR #26, opened 2017-05-30
Adds a skeleton SDL2 based native viewer with superior performance.

SirVer:
This is a first draft: It does not update the points it displays yet and the octree directory is hard coded to `./octree`, but it is a working start.

OpenGL is used rather down-to-the-metal on purpose. My hope is that we can get this to compile to emscripten and maybe replace the typescript based one eventually.

### DIFF
--- a/sdl_viewer/shaders/points.vs
+++ b/sdl_viewer/shaders/points.vs
@@ -1,0 +1,19 @@
+#version 100
+
+precision mediump float;
+
+uniform mat4 world_to_gl;
+uniform float edge_length;
+uniform vec3 min;
+
+attribute vec3 position;
+attribute vec3 color;
+
+varying vec4 v_color;
+
+void main() {
+  v_color = vec4(color / 255., 1.);
+  // TODO(hrapp): Support point size as a uniform.
+  gl_PointSize = 2.;
+  gl_Position = world_to_gl * vec4(position * edge_length + min, 1.0);
+}
--- a/sdl_viewer/shaders/points.vs
+++ b/sdl_viewer/shaders/points.vs
@@ -15,5 +15,8 @@ void main() {
   v_color = vec4(color / 255., 1.);
   // TODO(hrapp): Support point size as a uniform.
   gl_PointSize = 2.;
+  // TODO(hrapp): In the WebGL viewer, we are CPU bound, so doing the decoding
+  // on the GPU made sense. In the native viewer, the CPU is mostly idle, so
+  // pre-decoding this on the CPU might be more performant.
   gl_Position = world_to_gl * vec4(position * edge_length + min, 1.0);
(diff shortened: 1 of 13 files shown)


sdl_viewer/src/main.rs
```diff
@@ -0,0 +1,604 @@
...
+            gl::DeleteShader(vertex_shader_id);
+            gl::DeleteShader(fragment_shader_id);
+        }
+
+        GlProgram { id }
+    }
+}
+
+impl Drop for GlProgram {
+    fn drop(&mut self) {}
```
wohe: Tell GL to delete the program with `id`?

sdl_viewer/src/main.rs
```diff
@@ -0,0 +1,604 @@
...
+    fn stop_rotating(&mut self) {
+        self.mouse_state = MouseState::None;
+    }
+
+    fn mouse_motion(&mut self, x: i32, y: i32) {
+        if self.mouse_state == MouseState::None {
+            return;
+        }
+        let end = Vector2::new(x as f32, y as f32);
+        let delta = end - self.rotate_start;
```
wohe: Rename `rotate_start` to `last_mouse_position` and always update? Makes it slightly easier to follow.

sdl_viewer/src/main.rs
```diff
@@ -0,0 +1,604 @@
...
+                octree::PositionEncoding::Uint8 => (true, gl::UNSIGNED_BYTE),
+                octree::PositionEncoding::Uint16 => (true, gl::UNSIGNED_SHORT),
+                octree::PositionEncoding::Float32 => (false, gl::FLOAT),
+            };
+            gl::BufferData(gl::ARRAY_BUFFER,
+                           node_data.position.len() as GLsizeiptr,
+                           mem::transmute(&node_data.position[0]),
+                           gl::STATIC_DRAW);
+
+            // Specify the layout of the vertex data
```
wohe: nit: Period is missing.

sdl_viewer/shaders/points.vs
```diff
@@ -0,0 +1,19 @@
...
+attribute vec3 position;
+attribute vec3 color;
+
+varying vec4 v_color;
+
+void main() {
+  v_color = vec4(color / 255., 1.);
+  // TODO(hrapp): Support point size as a uniform.
+  gl_PointSize = 2.;
+  gl_Position = world_to_gl * vec4(position * edge_length + min, 1.0);
```
SirVer: added a TODO for now.

sdl_viewer/src/main.rs
SirVer: Found an easier way. SDL actually provides `[x|y]_rel`.

PR #31, opened 2017-05-30
Correctly subsample up to the root

### LINKED ISSUE #22: Some voxels are never displayed
Consistently some voxels are never drawn (a black box is in the data). This is consistent, i.e. nuking and recreating the octree from a .ply file results in the same voxels missing.

Unsure if this is an issue with building or viewing the octree.

SirVer:
A thinko was in this code: if a node was generated by subsampling children, we inserted its parent to be a new leaf node. That was incorrect, in fact the node itself is a new leaf node and its parent must also be generated from subsampling.

Fixes #22.

### DIFF
--- a/src/octree/mod.rs
+++ b/src/octree/mod.rs
@@ -134,7 +134,7 @@ impl Octree {
             }
             let file_name = path.file_name().unwrap();
             let file_name_str = file_name.to_str().unwrap();
-            if !file_name_str.starts_with("r") && file_name_str.ends_with(".xyz") {
+            if !file_name_str.starts_with("r") && !file_name_str.ends_with(".xyz") {
                 continue;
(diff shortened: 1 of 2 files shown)


SirVer:
Based on #30.

src/octree/mod.rs
```diff
@@ -134,7 +134,7 @@ impl Octree {
             }
             let file_name = path.file_name().unwrap();
             let file_name_str = file_name.to_str().unwrap();
-            if !file_name_str.starts_with("r") && file_name_str.ends_with(".xyz") {
+            if !file_name_str.starts_with("r") && !file_name_str.ends_with(".xyz") {
```
wohe: `||`?

PR #34, opened 2017-06-02
Use level_of_detail when drawing nodes.

SirVer:


sdl_viewer/src/main.rs
```diff
@@ -98,17 +104,33 @@ struct NodeView {
...
 impl NodeView {
     fn new(program: &GlProgram, node_data: octree::NodeData) -> Self {
         let vertex_array = GlVertexArray::new();
         vertex_array.bind();
 
-        let buffer_position = GlBuffer::new();
-        let buffer_color = GlBuffer::new();
+        let buffer_position = GlBuffer::new(GlBufferKind::ArrayBuffer);
+        let buffer_color = GlBuffer::new(GlBufferKind::ArrayBuffer);
+        let element_buffer = GlBuffer::new(GlBufferKind::ElementArrayBuffer);
```
wohe: Isn't this usually called an index buffer? And isn't this actually unnecessary here, since we could just reorder `buffer_position` and `buffer_color` according to the random shuffle `indices`, using `glDrawArrays` as before?

sdl_viewer/src/main.rs
```diff
@@ -38,6 +40,17 @@ use std::str;
 const FRAGMENT_SHADER: &'static str = include_str!("../shaders/points.fs");
 const VERTEX_SHADER: &'static str = include_str!("../shaders/points.vs");
 
+fn reshuffle(new_order: &[usize], old_data: Vec<u8>, bytes_per_point: usize) -> Vec<u8> {
```
wohe: `point` -> `vertex`? I think that is more common GL term for the data at this point in time, i.e. before rendering them as points.

PR #35, opened 2017-06-02
Load new nodes when camera view changed.

SirVer:
- Disables LOD when standing still. This allows us to get a full view when standing still and decent FPS when moving. 
- Adds `force load` as a binding to `r`. This forces the viewer to load all currently visible nodes to guarantee that all points are actually displayed. This is of course a hack - if our LOD was good enough, we had more confidence it what we actually need to draw to fully cover the screen.

Loading feels too slow still, but I have no good idea how to make it faster for now.

### DIFF
--- a/sdl_viewer/src/main.rs
+++ b/sdl_viewer/src/main.rs
@@ -31,6 +31,9 @@ use sdl2::video::GLProfile;
 use sdl_viewer::{Camera, gl};
 use sdl_viewer::gl::types::{GLboolean, GLint, GLsizeiptr, GLuint};
 use sdl_viewer::graphic::{GlBuffer, GlProgram, GlVertexArray};
+use std::collections::{HashMap, HashSet};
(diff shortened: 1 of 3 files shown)


sdl_viewer/src/main.rs
```diff
@@ -194,6 +197,62 @@ impl NodeView {
...
+    node_views: HashMap<octree::NodeId, NodeView>,
+    unloaded_nodes: VecDeque<octree::NodeId>,
+    queued: HashSet<octree::NodeId>,
+}
+
+impl NodeViewContainer {
+    fn new() -> Self {
+        NodeViewContainer {
+            node_views: HashMap::new(),
+            unloaded_nodes: VecDeque::new(),
```
wohe: `unloaded_nodes` -> `queue`? Unloaded sounds like they have been removed, but this is the same as `queued` just in the correct ordering, i.e. the queue, right?

PR #36, opened 2017-06-02
Change NodeId to a (level, index) tuple instead of a string.

SirVer:
This allows to make the value cheap to copy, so derive Copy.

### DIFF
--- a/src/octree/node.rs
+++ b/src/octree/node.rs
@@ -47,43 +47,58 @@ impl ChildIndex {
 /// A unique identifier to a node. Currently this is implemented as 'r' being the root and r[0-7]
 /// being the children, r[0-7][0-7] being the grand children and so on. The actual representation
 /// might change though.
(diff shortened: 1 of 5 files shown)


src/octree/node.rs
```diff
@@ -533,21 +542,21 @@ mod tests {
...
-            NodeId::from_string("r123451".into()).child_index()
+            NodeId::from_str("r123451").child_index()
         );
         assert_eq!(
             Some(ChildIndex(7)),
-            NodeId::from_string("r123457".into()).child_index()
+            NodeId::from_str("r123457").child_index()
         );
-        assert_eq!(None, NodeId::from_string("r".into()).child_index());
+        assert_eq!(None, NodeId::from_str("r".into()).child_index());
```
wohe: Isn't that `.into()` unnecessary?

SirVer: yup

PR #39, opened 2017-07-03
Switch to using prost.

### LINKED ISSUE #38: Investigate using prost for protobuf.
Main benefit seems that it can be used to generate the .rs files at build.

https://github.com/danburkert/prost

SirVer:
Fixes #38.


SirVer:
The diff looks huge because of changes to Cargo.lock, but is in fact pretty small.

src/octree/node.rs
```diff
@@ -231,22 +232,30 @@ impl NodeMeta {
...
         Ok(
             NodeMeta {
-                num_points: meta.get_num_points(),
-                position_encoding: PositionEncoding::from_proto(&meta.get_position_encoding()),
+                num_points: meta.num_points.unwrap(),
+                position_encoding: PositionEncoding::from_proto(
+                    proto::node::PositionEncoding::from_i32(
+                        meta.position_encoding.unwrap(),
+                    )
+                            .unwrap()
```
wohe: Is that indentation intentional? Seems strange to me.

SirVer: that's rustformatted... I do not know what it was thinking there.

PR #40, opened 2017-07-03
Add some PLY parsing tests and support skipping over unknown properties.

### LINKED ISSUE #37: build_octree chokes on unknown point properties
A File like this: 

~~~
ply
format binary_little_endian 1.0
element vertex 38400000
property float x
property float y
property float z
property uchar red
property uchar green
property uchar blue
property uchar alpha
element face 0
property list uchar int vertex_indices
end_header
~~~

triggers this error:

~~~
thread 'main' panicked at 'Unknown property 'alpha' on 'vertex'.', src/ply.rs:339
note: Run with `RUST_BACKTRACE=1` for a backtrace.
~~~

solution is to skip over unknown properties while parsing the PLY. The framework for this is in place, it just needs to be done.

SirVer:
Fixes #37.


src/ply.rs
```diff
@@ -264,6 +264,19 @@ macro_rules! read_casted_property {
     )
 }
 
+// Similar, but creates a function that just advances the read pointer.
```
wohe: Similar to what?

src/ply.rs
```diff
@@ -264,6 +264,19 @@ macro_rules! read_casted_property {
...
+macro_rules! create_skip_fn {
+    (&mut $size:ident, $num_bytes:expr) => (
```
wohe: Not related to this macro but GitHub does not allow me to put the comment where it belongs. In `create_and_return_reading_fn` the macro takes a parameter `data_type` that is not mentioned in the macro body. Can it be removed or what am I missing?

SirVer: yes, done.

PR #42, opened 2017-07-04
Do not split a node that is smaller than the desired resolution.

### LINKED ISSUE #9: 100.000 points in the same place breaks building point clouds
One goal was to not throw away any points while building the point cloud. The builder therefore splits nodes by deciding on the number of points inside the node. 

Problem is, if there are more points than we allow in one node in the same spot, we will infinitely recurse trying to split the node in smaller and smaller pieces, but they will always continue more than MAX points.

Some possible solutions come to mind:

1. do not create nodes smaller than `resolution` and throw points away that we could not make fit.
2. allow nodes with more points than MAX in corner situations.
3. Filter by nearest neighbors.

SirVer:
Fixes #9.

### DIFF
--- a/src/bin/build_octree.rs
+++ b/src/bin/build_octree.rs
@@ -110,6 +110,25 @@ fn split<P>(
     rv
 }
 
+fn should_split_node(node: &SplittedNode, resolution: f64) -> bool {
+    if node.num_points <= MAX_POINTS_PER_NODE {
+        return false;
+    }
+    if node.node.bounding_cube.edge_length() as f64 <= resolution {
+        // TODO(hrapp): If the data has billion of points in this small spot, performance will
+        // greatly suffer if we display it. Drop points?
+        println!(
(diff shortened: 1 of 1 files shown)


src/bin/build_octree.rs
```diff
@@ -123,7 +142,7 @@ fn split_node<'a, 'b: 'a, P>(
     let (leaf_nodes, split_nodes): (Vec<_>, Vec<_>) =
         children
             .into_iter()
-            .partition(|n| n.num_points < MAX_POINTS_PER_NODE);
+            .partition(|n| !should_split_node(n, resolution));
```
wohe: Would it make sense to invert this logic for readability, i.e. instead of `!should_split_node()` rather have `is_leaf_node()`?

SirVer: I had this originally, but found that the function is harder to think about then. So I opted for the inversion here.

PR #59, opened 2018-01-08
Switch to using StructGenerator from glbindings

SirVer:
Instead of standalone functions, we now load the GL bindings into a struct. This should make it easier to move all opengl related functionality into one submodule, which again should make it simpler to swap it out against other functionality (e.g. WebGL).


SirVer:
@thomasschiwietz rebased and updated. PTAL.

SirVer:
@thomasschiwietz PTAL.

SirVer:
@wally-the-cartographer merge

wally-the-cartographer:

PR #70, opened 2018-01-15
Bring back a .rustfmt.toml.

SirVer:
I realized that the default style does not reorder imports, which is silly. Also reformat code with new style.


SirVer:
@wally-the-cartographer merge

wally-the-cartographer:

PR #74, opened 2018-01-18
Keep the bounding instead of the cube for an octree.

### LINKED ISSUE #73: Keep actual bouding_box instead of bounding_cube in octree meta data.
Reasoning is this: the bounding cube can easily be derived from the bounding box, but for processing over points it is sometimes useful to have the bounding box, which cannot be derived from the bounding cube.

SirVer:
This is changing the on-disk data structure, i.e. is breaking support for old octrees. They need to be regenerated.

Fixes #73.


SirVer:
@wally-the-cartographer merge

wally-the-cartographer:

PR #82, opened 2018-01-19
Use the workspace feature.

SirVer:
Please rm -rf */target */Cargo.lock.

This might speed up CI as well, since now the correct directory is
cached.


SirVer:
Reminder to myself: Update README (for new target positions), fix `cargo --all test` to `cargo test -all` and change "View billions of points in your browser."

ci/script.sh
```diff
@@ -1,32 +1,15 @@
...
-    cargo build --verbose
-    cargo test
-    popd
-
-    # sdl viewer
-    pushd sdl_viewer
-    cargo build --verbose
-    cargo test
-    popd
+    # root
```
thomasschiwietz: this is not the root directory but web_viewer. add another popd to get to root

SirVer: done!

wally-the-cartographer:

PR #85, opened 2018-01-23
Store meta information of all nodes in root meta.pb

thomasschiwietz:
This pull request is part of fixing #86 

Things to be done in follow up pull requests:
- build_octree: store NodeMeta in HashMap and remove load/store of r*.pb (meta data for each node)
- proto: replace 'repeated' by 'map'
- proto: get rid of bounding_box
- adapt S3Octree

### DIFF
--- a/src/bin/build_octree.rs
+++ b/src/bin/build_octree.rs
@@ -307,27 +307,30 @@ fn main() {
 
     let (bounding_box, num_points) = find_bounding_box(&input);
 
-    let mut meta = proto::Meta::new();
-    meta.mut_bounding_box()
-        .mut_min()
-        .set_x(bounding_box.min().x);
-    meta.mut_bounding_box()
-        .mut_min()
-        .set_y(bounding_box.min().y);
-    meta.mut_bounding_box()
-        .mut_min()
-        .set_z(bounding_box.min().z);
-    meta.mut_bounding_box()
-        .mut_max()
-        .set_x(bounding_box.max().x);
-    meta.mut_bounding_box()
-        .mut_max()
-        .set_y(bounding_box.max().y);
-    meta.mut_bounding_box()
-        .mut_max()
-        .set_z(bounding_box.max().z);
-    meta.set_resolution(resolution);
-    meta.set_version(octree::CURRENT_VERSION);
+    let mut meta = {
+        let mut meta = proto::Meta::new();
+        meta.mut_bounding_box()
+            .mut_min()
+            .set_x(bounding_box.min().x);
+        meta.mut_bounding_box()
+            .mut_min()
+            .set_y(bounding_box.min().y);
+        meta.mut_bounding_box()
+            .mut_min()
+            .set_z(bounding_box.min().z);
+        meta.mut_bounding_box()
+            .mut_max()
+            .set_x(bounding_box.max().x);
+        meta.mut_bounding_box()
+            .mut_max()
+            .set_y(bounding_box.max().y);
+        meta.mut_bounding_box()
+            .mut_max()
+            .set_z(bounding_box.max().z);
+        meta.set_resolution(resolution);
+        meta.set_version(octree::CURRENT_VERSION);
+        meta
+    };
 
     // Ignore errors, maybe directory is already there.
     let _ = fs::create_dir(output_directory);
--- a/src/bin/build_octree.rs
+++ b/src/bin/build_octree.rs
@@ -347,8 +347,9 @@ fn main() {
         nodes_to_subsample.push(leaf_node);
     }
 
-    // We start on the deepest level and work our way up the tree.
+    // sub sampling returns the list of finished nodes including all meta data
     let (nodes_sender, nodes_receiver) = mpsc::channel();    
+    // We start on the deepest level and work our way up the tree.
     for current_level in (0..deepest_level + 1).rev() {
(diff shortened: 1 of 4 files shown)


SirVer:
Please fix first post - it will become the commit message. 

If you want to track longer term progress on a task, I suggest opening an issue and mentioning it in the commit message. The issue will then show the full evolution of the feature.

src/octree/mod.rs
```diff
@@ -92,8 +91,8 @@ fn size_in_pixels(bounding_cube: &Cube, matrix: &Matrix4f, width: i32, height: i
 #[derive(Debug)]
 pub struct OnDiskOctree {
     directory: PathBuf,
-    // Maps from node id to number of points.
-    nodes: HashMap<NodeId, u64>,
+    // Maps from node id to NodeMeta: bounding_box, number of points, PositionEncoding.
```
SirVer: this comment will go out of sync as as NodeMeta changes. It is also now tautologically, as the data type contains the same information.

src/octree/mod.rs
```diff
@@ -130,7 +129,7 @@ impl OnDiskOctree {
             return Err(ErrorKind::InvalidVersion(3).into());
         }
 
-        let meta = {
+        let mut meta = {
```
SirVer: no need for mut if changed below

src/octree/mod.rs
```diff
@@ -152,23 +151,18 @@ impl OnDiskOctree {
...
-            if path.file_name().is_none() {
-                continue;
-            }
-            let file_name = path.file_name().unwrap();
-            let file_name_str = file_name.to_str().unwrap();
-            if !file_name_str.starts_with('r') || !file_name_str.ends_with(".xyz") {
-                continue;
-            }
-            let num_points = fs::metadata(path).unwrap().len() / 12;
+        for meta in meta.mut_nodes().iter() {
```
SirVer: meta.nodes.iter() should be immutable

src/octree/mod.rs
```diff
@@ -142,7 +141,7 @@ impl OnDiskOctree {
         }
 
         let bounding_box = {
-            let bounding_box = meta.bounding_box.unwrap();
+            let bounding_box = meta.bounding_box.clone().unwrap();
```
SirVer: clone unnecessary?

src/octree/mod.rs
```diff
@@ -152,23 +151,18 @@ impl OnDiskOctree {
...
+        for meta in meta.mut_nodes().iter() {
             nodes.insert(
-                NodeId::from_str(path.file_stem().unwrap().to_str().unwrap()),
-                num_points,
+                NodeId::from_str(&meta.id), 
+                NodeMeta {
+                    num_points: meta.num_points,
+                    position_encoding: PositionEncoding::from_proto(meta.position_encoding)?,
+                    bounding_cube: {
+                        let proto = meta.bounding_cube.clone().unwrap();
```
SirVer: as_ref instead of clone

src/octree/mod.rs
```diff
@@ -152,23 +151,18 @@ impl OnDiskOctree {
...
+                        let min = proto.min.unwrap();
```
SirVer: proto.min.as_ref().unwrap()

src/proto.proto
```diff
@@ -51,4 +45,12 @@ message Node {
   AxisAlignedCube bounding_cube = 1;
   PositionEncoding position_encoding = 2;
   int64 num_points = 3;
+  string id = 4;
```
SirVer: Internally we represent a NodeId as (lvl: u8, index: u32). That would be a more efficient schema and closer to our implementation.

src/bin/build_octree.rs
```diff
@@ -159,6 +159,7 @@ fn subsample_children_into(
     output_directory: &Path,
     node: &octree::Node,
     resolution: f64,
+    nodes_sender: &mpsc::Sender<(octree::NodeId, octree::NodeMeta)>,
```
SirVer: finished_node_sender?

src/bin/build_octree.rs
```diff
@@ -372,8 +391,9 @@ fn main() {
 
         pool.scoped(|scope| {
             for node in &subsample_nodes {
+                let nodes_sender_clone = nodes_sender.clone();
                 scope.execute(move || {
```
SirVer: This will queue all nodes metadata in memory, since you do not have started a thread yet that consumes them. I think that is fine for now, since we are assuming that we hold them in memory anyways.

src/bin/build_octree.rs
```diff
@@ -382,4 +402,32 @@ fn main() {
         // their parents.
         nodes_to_subsample.extend(subsample_nodes.into_iter());
     }
+
+    // Add all node meta data to meta.pb
+    // TODO(tschiwietz): How can we make sure not to miss any messages?
+    while let Ok((id, node_meta)) = nodes_receiver.try_recv() {
```
SirVer: This should work:

drop(nodes_sender); 
for node_meta in nodes_receiver {}

The stream terminates as soon as no sender is live anymore. All the clones die with their scope, but the original is still live. You can drop it manually to avoid a deadlock.

SirVer:
@wally-the-cartographer merge

wally-the-cartographer:

PR #87, opened 2018-01-25
Store node meta data in-memory instead of disc

thomasschiwietz:
Fixes part of #86 
This pull request removes the SplittedNode struct and returns NodeId only. Also, num_points in NodeMeta is now optional because it can be unknown from the caller site.

### DIFF
--- a/src/bin/build_octree.rs
+++ b/src/bin/build_octree.rs
@@ -33,26 +33,21 @@ use point_viewer::proto;
 use point_viewer::pts::PtsIterator;
 use protobuf::Message;
 use scoped_pool::{Pool, Scope};
-use std::collections::HashSet;
+use std::collections::{HashMap, HashSet};
 use std::fs::{self, File};
 use std::io::{BufWriter, Stdout};
-use std::path::{Path, PathBuf};
+use std::path::PathBuf;
 use std::sync::mpsc;
 
 const UPDATE_COUNT: i64 = 100000;
 const MAX_POINTS_PER_NODE: i64 = 100000;
 
-struct SplittedNode {
-    node: octree::Node,
-    num_points: i64,
-}
-
+// Return a list a leaf nodes and a list of nodes to be splitted further.
 fn split<P>(
-    output_directory: &Path,
-    resolution: f64,
-    node: &octree::Node,
+    octree_meta: &octree::OctreeMeta,
+    node_id: &octree::NodeId,
     stream: P,
-) -> Vec<SplittedNode>
+) -> (Vec<octree::NodeId>, Vec<octree::NodeId>)
 where
     P: InternalIterator,
 {
@@ -61,24 +56,24 @@ where
     match stream.size_hint() {
         Some(size) => println!(
             "Splitting {} which has {} points ({:.2}x MAX_POINTS_PER_NODE).",
-            node.id,
+            node_id,
             size,
             size as f64 / MAX_POINTS_PER_NODE as f64
         ),
         None => println!(
             "Splitting {} which has an unknown number of points.",
-            node.id
+            node_id
         ),
     };
 
+    let bounding_cube = node_id.find_bounding_cube(&octree_meta.root_bounding_cube);
     stream.for_each(|p| {
-        let child_index = node.get_child_id_containing_point(&p.position);
+        let child_index = octree::ChildIndex::from_bounding_cube(&bounding_cube, &p.position);
         let array_index = child_index.as_u8() as usize;
         if children[array_index].is_none() {
             children[array_index] = Some(octree::NodeWriter::new(
-                output_directory,
-                &node.get_child(child_index),
-                resolution,
+                &octree_meta,
+                &node_id.get_child_id(child_index),
             ));
         }
         children[array_index].as_mut().unwrap().write(p);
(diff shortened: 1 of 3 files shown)


src/bin/build_octree.rs
```diff
@@ -29,26 +29,26 @@ use point_viewer::proto;
...
-struct SplittedNode {
-    node: octree::Node,
-    num_points: i64,
+struct BuildMeta {
+    output_directory: PathBuf,
+    resolution: f64,
+    root_bounding_cube: Cube,
 }
 
 fn split<P>(
```
SirVer: Add a comment about which values are returned here.

src/bin/build_octree.rs
```diff
@@ -57,24 +57,26 @@ where
...
         ),
     };
 
+    let bounding_cube = node_id.bounding_cube(&build_meta.root_bounding_cube);
     stream.for_each(|p| {
-        let child_index = node.get_child_id_containing_point(&p.position);
+        let child_index = octree::ChildIndex::from_bounding_cube(&bounding_cube, &p.position);
         let array_index = child_index.as_u8() as usize;
         if children[array_index].is_none() {
             children[array_index] = Some(octree::NodeWriter::new(
```
SirVer: You pass in all members of BuildMeta. Maybe call it OctreeMeta and put it into the lib?

src/bin/build_octree.rs
```diff
@@ -122,50 +128,56 @@ fn should_split_node(node: &SplittedNode, resolution: f64) -> bool {
...
-        .into_iter()
-        .partition(|n| !should_split_node(n, resolution));
-
-    for child in split_nodes {
+    let (leaf_nodes, split_nodes) = split(build_meta, &node_id, stream);
+    for child_id in split_nodes {
         let leaf_nodes_sender_clone = leaf_nodes_sender.clone();
         scope.recurse(move |scope| {
-            let stream = octree::NodeIterator::from_disk(output_directory, &child.node.id).unwrap();
+            let bounding_cube = child_id.bounding_cube(&build_meta.root_bounding_cube);
```
SirVer: Rename the function into find_bounding_cube? It is not a getter, but does some computation.

src/bin/build_octree.rs
```diff
@@ -122,50 +128,56 @@ fn should_split_node(node: &SplittedNode, resolution: f64) -> bool {
...
+            let child_meta = octree::NodeMeta {
```
SirVer: passing in the OctreeMeta should also allow to not require num_ponits to be optional anymore.

src/octree/node.rs
```diff
@@ -98,7 +107,7 @@ impl NodeId {
     }
 
     /// Returns the NodeId for the corresponding 'child_index'.
-    fn get_child_id(&self, child_index: &ChildIndex) -> Self {
+    pub fn get_child_id(&self, child_index: &ChildIndex) -> Self {
```
SirVer: nit: ChildIndex is copy and only 1 byte, so taking it by value is cheaper than taking a reference.

src/bin/build_octree.rs
```diff
@@ -374,58 +374,70 @@ fn main() {
...
+            for id in &subsample_nodes {
+                let all_nodes_sender_clone = all_nodes_sender.clone();
                 scope.execute(move || {
-                    subsample_children_into(output_directory, node, resolution, &nodes_sender_clone).unwrap();
+                    subsample_children_into(build_meta_ref, id, &all_nodes_sender_clone).unwrap();
                 });
             }
         });
 
+        // TODO(tschiwietz): need to separate add/remove nodes
```
SirVer: I do not undestand this comment

src/bin/build_octree.rs
```diff
@@ -374,58 +374,70 @@ fn main() {
...
         // their parents.
         nodes_to_subsample.extend(subsample_nodes.into_iter());
     }
 
-    // Add all node meta data to meta.pb
-    drop(nodes_sender);
-    for (id, node_meta) in nodes_receiver {
+    // Add all non-zero node meta data to meta.pb
+    for (id, num_points) in finished_nodes {
+        if num_points == 0 {
```
SirVer: Why even sending these?

src/octree/node.rs
```diff
@@ -133,6 +142,37 @@ impl NodeId {
     pub fn index(&self) -> usize {
         self.index as usize
     }
+
+    /// Computes the bounding cube from a NodeID.
+    pub fn bounding_cube(&self, root_bounding_cube: &Cube) -> Cube {
```
SirVer: Add a few tests at the bottom of the file?

src/octree/node.rs
```diff
@@ -133,6 +142,37 @@ impl NodeId {
...
+
+    // Get number of points from the file size of the color data.
+    // Color data is required and always present.
+    fn number_of_points(&self, directory: &Path) -> Result<i64> {
+        let file_meta_data_opt = fs::metadata(self.get_stem(directory).with_extension(COLOR_EXT));
+        if file_meta_data_opt.is_err() {
+            return Err(ErrorKind::NodeNotFound.into())
+        }
+
+        let file_size_bytes = file_meta_data_opt.unwrap().len();
```
SirVer: In which cases can unwrap fail here? if they are reasonbale, we should probably return an error. If they unreasonable, please add a comment why.

thomasschiwietz: From the documentation, errors occurs when:
1. The user lacks permissions to perform metadata call on path.
2. path does not exist.
We return NodeNotFound in any case. Do you think we need more diverse return values?

SirVer: I think that is fine.

SirVer:
@thomasschiwietz I made some nitty changes while reviewing. Could you have a quick look?

SirVer:
@wally-the-cartographer merge

wally-the-cartographer:

PR #88, opened 2018-01-29
No longer reexport basic math types.

SirVer:
Makes usage of cgmath and collision explicit.
Also removes one instance of glob importing proto::*


SirVer:
@wally-the-cartographer merge

wally-the-cartographer:

PR #95, opened 2018-02-01
Remove bounding cube from meta.pb

thomasschiwietz:
This change requires to rebuild octrees.

### DIFF
--- a/src/octree/mod.rs
+++ b/src/octree/mod.rs
@@ -31,7 +31,7 @@ mod node;
 pub use self::node::{ChildIndex, Node, NodeId, NodeIterator, NodeMeta, NodeWriter,
                      PositionEncoding};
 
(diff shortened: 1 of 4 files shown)


src/octree/mod.rs
```diff
@@ -31,7 +31,7 @@ mod node;
 pub use self::node::{ChildIndex, Node, NodeId, NodeIterator, NodeMeta, NodeWriter,
                      PositionEncoding};
 
-pub const CURRENT_VERSION: i32 = 9;
+pub const CURRENT_VERSION: i32 = 10;
```
SirVer: no need to increase the version? The old protos will load fine, they will just contain no longer used data.

SirVer:
@wally-the-cartographer merge

wally-the-cartographer:

PR #99, opened 2018-02-02
Add grpc function to get all points inside a bounding box

thomasschiwietz:


grpc/src/bin/octree_server.rs
```diff
@@ -110,6 +113,34 @@ impl proto_grpc::Octree for OctreeService {
             .map_err(move |e| println!("failed to reply {:?}: {:?}", req, e));
         ctx.spawn(f)
     }
+
+    fn get_points_in_box(
```
SirVer: I think this should be a streaming rpc.

the blog post of the pingcap people (that write the grpc crate we use) might be interesting: https://pingcap.com/blog/2017-09-12-futuresandgrpc/

There is a tiny example: https://github.com/pingcap/grpc-rs/blob/master/examples/route_guide/server.rs#L61

and the documentation:
https://docs.rs/grpcio/0.2.0/grpcio/struct.ServerStreamingSink.html

thomasschiwietz:
Will open a new pull request with a streaming grpc interface

PR #105, opened 2018-02-05
Format TypeScript and JavaScript using Prettier (prettier.io)

pdubroy:
- Add `npm run format` script to auto-format all JS and TS code under client/
- Update existing code to use Prettier output.

### DIFF
--- a/xray/client/package.json
+++ b/xray/client/package.json
@@ -5,7 +5,6 @@
   "main": "index.js",
   "scripts": {
     "build": "rollup -c",
-    "test": "echo \"Error: no test specified\" && exit 1",
     "format": "prettier --write **/*.ts **/*.js"
   },
   "author": "Holger Rapp <hrapp@lyft.com>",
--- a/xray/client/package.json
+++ b/xray/client/package.json
@@ -5,16 +5,25 @@
   "main": "index.js",
   "scripts": {
     "build": "rollup -c",
-    "test": "echo \"Error: no test specified\" && exit 1"
+    "test": "echo \"Error: no test specified\" && exit 1",
+    "format": "prettier --write **/*.ts **/*.js"
   },
   "author": "Holger Rapp <hrapp@lyft.com>",
   "license": "Apache-2.0",
   "devDependencies": {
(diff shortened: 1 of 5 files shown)


xray/client/package.json
```diff
@@ -5,16 +5,25 @@
   "main": "index.js",
   "scripts": {
     "build": "rollup -c",
-    "test": "echo \"Error: no test specified\" && exit 1"
+    "test": "echo \"Error: no test specified\" && exit 1",
```
SirVer: Did prettier put that line here? Remove?

xray/client/package.json
```diff
@@ -5,16 +5,25 @@
...
     "rollup-plugin-node-resolve": "^3.0.2",
     "rollup-plugin-typescript": "^0.8.1",
     "three": "0.89.0"
+  },
+  "prettier": {
+    "arrow-parens": "always",
+    "bracket-spacing": false,
+    "print-width": 80,
+    "single-quote": true,
+    "trailing-comma": "es5"
```
SirVer: in typescript we use es6 for output. Should this be consistent?

pdubroy: For JS code I think we want `"es5"` here — it's a bit weird but the only options are `"none"`, `"es5"`, and `"all"`: https://prettier.io/docs/en/options.html#trailing-commas. If it's set to `"all"`, it will insert trailing commas in function parameter lists. That's not valid ES6 (though it is supported in TypeScript, apparently).

I could enable it for the TypeScript if you want though.

xray/client/package.json
pdubroy: No, it was there before, I just changed the ordering to build/test/format, whereas it was build/format/test before. Probably makes sense to just remove it though.

pdubroy: Done.

xray/client/package.json
SirVer: that's fine.

SirVer:
@wally-the-cartographer merge

wally-the-cartographer:

PR #107, opened 2018-02-06
Add intensity as an optional quantity into the viewer.

SirVer:
For a PLY file containing intensity, this increases resources required to make an octree, and of course also for serving it. For one representative dataset, the data is listed below. Of note is that on disk size increases by ~50%, though the bytes per point only increase from 15 to 19, this is overhead in the many files. The construction time increases to 127% of master, I think this is mainly due to an additional branch per point, but I did not properly benchmark.

~~~
master:
   On Disk size: 11G

   1220.07user 100.08system 5:23.28elapsed 408%CPU (0avgtext+0avgdata 57876maxresident)k
   36274272inputs+218705688outputs (0major+2153637minor)pagefaults 0swaps

this change:
   On disk size: 16G

   1413.66user 146.44system 6:50.26elapsed 380%CPU (0avgtext+0avgdata 49380maxresident)k
   43523920inputs+309758664outputs (5major+5120192minor)pagefaults 0swaps
~~~


SirVer:
@wally-the-cartographer merge

SirVer:
@wally-the-cartographer merge

wally-the-cartographer:

PR #114, opened 2018-02-12
Performance improvements in the SDL viewer.

SirVer:
- Render less points while moving.
- Never use LOD.
- Choose the level to render on the amount of nodes that fit in the
  cache.


SirVer:
@wally-the-cartographer merge

wally-the-cartographer:

PR #119, opened 2018-02-16
Duplicate the logic of not showing points when moving in the web viewer.

SirVer:


web_viewer/client/main.ts
```diff
@@ -57,14 +59,39 @@ class App {
     );
 
     this.lastFrustumUpdateTime = 0;
-    this.viewer = new OctreeViewer(this.scene);
+    this.lastMoveTime = 0;
+    this.needsRender = true;
+    this.viewer = new OctreeViewer(this.scene, () => {
+      console.log('#sirver ');
```
pdubroy: Forgot to remove this?

web_viewer/client/octree_viewer.ts
```diff
@@ -308,6 +309,27 @@ export class OctreeViewer {
       });
   }
 
+  public setMoving(value: boolean) {
+    if (value) {
+      for (const nodeId of Object.keys(this.loadedData)) {
+        if (
```
pdubroy: Seems like confusing logic. Any reason you can't invert the condition here and put the last line inside the `if`?

web_viewer/client/octree_viewer.ts
```diff
@@ -308,6 +309,27 @@ export class OctreeViewer {
...
+          nodeId.length <= this.maxLevelToDisplay ||
+          this.loadedData[nodeId].threePoints === undefined
+        ) {
+          continue;
+        }
+        this.loadedData[nodeId].threePoints.visible = false;
+      }
+    } else {
+      for (const nodeId of Object.keys(this.loadedData)) {
```
pdubroy: Since this looks like it's the exact same loop as in the branch above, I think it would be clearer to move the `if (value)` inside the loop.

web_viewer/client/octree_viewer.ts
SirVer: I cannot wrap my head around the inverted condition.... The current code reads clearer to me. I added a bit of comments to make it better understandable. wdyt?

pdubroy: I just meant this (DeMorgan's law ftw):

```
  for (const nodeId of Object.keys(this.loadedData)) {
    if (nodeId.length > maxLevelToDisplay &&
        this.loadedData[nodeId].threePoints) {
      this.loadedData[nodeId].threePoints.visible = false;
    }
  }
```
As a reader, if I want to understand when the line below *does* get executed, I already have to mentally invert the condition. To me, it seems weird to say, "If such-and-such is true, do nothing. Otherwise, do this thing."

web_viewer/client/octree_viewer.ts
pdubroy: I would just rewrite the whole thing like this:

```
public setMoving(moving: boolean) {
  for (const nodeId of Object.keys(this.loadedData)) {
    const threePoints = this.loadedData[nodeId];
    if (threePoints != null) {
      // If we are moving, only show points above a certain depth. Otherwise, show them all.
      if (moving) {
        threePoints.visible = nodeId.length <= maxLevelToDisplay;
      } else {
        threePoints.visible = true;
      }
    }
  }
}
```

web_viewer/client/octree_viewer.ts
SirVer: okay, done now. I guess my brain finally settled in.

web_viewer/client/octree_viewer.ts
SirVer: that is very beautiful. I am buying this.

pdubroy:
@wally-the-cartographer merge

wally-the-cartographer:

PR #128, opened 2018-04-25
Refactor to enable more than one way of creating tiles from points.

SirVer:
This is for creating colored xrays in a followup PR.


wally-the-cartographer:

PR #130, opened 2018-04-26
Add dispose() into the xRayViewer so that it clears out the THREE scene.

SirVer:
Also ran `yarn format`.

### DIFF
--- a/xray/client/xray_viewer.ts
+++ b/xray/client/xray_viewer.ts
@@ -74,7 +74,7 @@ export class XRayViewer {
   private nextNodesForLevelQueryId: number;
   private displayLevel: number;
   private meta: any; // Will be undefined until we have loaded the metadata.
-  private cleanupCalled = false;
+  private isDisposed = false;
 
   constructor(private scene: THREE.Scene, private prefix: string) {
     this.currentlyLoading = 0;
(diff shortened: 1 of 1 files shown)


pdubroy:
Can we use the same prettier config that we are using in internal code? Your formatting changes here make this different than our internal style. (Which isn't necessarily a problem, because it's third party code, but if the style config here is arbitrary, I'd rather have things be the same.)

SirVer:
@pdubroy I have no strong feelings about the style used here. But for now this is what is checked in.

xray/client/xray_viewer.ts
```diff
@@ -42,51 +42,63 @@ class NodeData {
...
 
 export class XRayViewer {
-  private nodes: {[key: string]: NodeData} = {};
+  private nodes: { [key: string]: NodeData } = {};
   private currentlyLoading: number;
   private nodesToLoad: string[] = [];
   private nextNodesForLevelQueryId: number;
   private displayLevel: number;
   private meta: any; // Will be undefined until we have loaded the metadata.
+  private cleanupCalled: boolean = false;
```
pdubroy: Nit: `boolean` is unnecessary here.

xray/client/xray_viewer.ts
```diff
@@ -183,6 +195,9 @@ export class XRayViewer {
   }
 
   private swapIn(nodeId: string) {
+    if (this.cleanupCalled) {
```
pdubroy: What is this actually guarding against? It looks like the texture loading callback is already guarded by `inScene`, so it seems like it's only `nodesUpdate` that could call into here. If so, I think it makes more sense to move this check to there.

pdubroy:
We should definitely look into the correct way to free the memory of the objects that we are removing from the scene. See [here](https://github.com/mrdoob/three.js/issues/5175) for some info about this.

xray/client/xray_viewer.ts
SirVer: done.

xray/client/xray_viewer.ts
SirVer: This guards that no new node is swapped in, which might happen when an async load returns. Once we cleaned up and removed all nodes from the scene, we defintively do not want to add more in, so I think this is the correct place to guard against - because here is where we swap in the new node.

SirVer:
@wally-the-cartographer merge

wally-the-cartographer:

PR #131, opened 2018-05-02
Adds a binary that generates a single X-Ray tile.

SirVer:


SirVer:
Diff includes and depends on #132 which must be merged before this can be reviewed.

SirVer:
@wally-the-cartographer merge

PR #135, opened 2018-06-07
Remove force-load all feature.

SirVer:
Remove the force-load feature for nodes. It is no longer useful with the large point clouds we deal with on a daily basis and makes a refactoring I want to do unnecessarily harder.


SirVer:
Because it makes a refactoring that I want to do awkward and it is not quite working as I expect it to either: it does not block until everything is loaded. We can bring it back if desired later.

@wally-the-cartographer merge

SirVer:
@wally-the-cartographer merge

wally-the-cartographer:

PR #136, opened 2018-06-07
Give every GL object ownership of the shared GL context.

SirVer:
Before, the `opengl::GL` class was created on the outside and references to it where passed throughout the program. This makes for awkward life-time annotations, so I decided to instead give shared ownership to OpenGL to everything that needs it.

This is done using a std::rc::Rc, a reference counter, that cannot be passed between threads and has the same performance characteristics to a reference at runtime, it is only minimally more expensive on creation (cloning) and deletion.


wally-the-cartographer:

PR #137, opened 2018-06-11
Refactor main draw logic into a separate class with the goal of reusing it from other languages.

SirVer:
- Also no longer unconditionally redraw every second, instead check if new data has been read by the IO thread and redraw if so.


SirVer:
@wally-the-cartographer merge

wally-the-cartographer:

PR #140, opened 2018-06-15
 Speedup `get_visible_nodes`, up to 10x.

SirVer:
This introduces a bunch of ideas that all made the code faster:

- Check for existance of nodes earlier, which prunes some parts of the
  octree traversal a level higher.
- Use a heap to explore the nodes in order of their size which makes
  sorting after impossible and has better pruning behavior in most
  cases.
- Distinguish between Frustum culling cases: if a parent node is fully
  inside the frustum, all of its children are too, no need to check again.
- Return NodeId instead of VisibleNode which is much smaller.

### DIFF
--- a/point_viewer_grpc/src/lib.rs
+++ b/point_viewer_grpc/src/lib.rs
@@ -26,8 +26,7 @@ use futures::{Future, Stream};
 use grpcio::{ChannelBuilder, EnvBuilder};
 use point_viewer::errors::*;
 use point_viewer::math::Cube;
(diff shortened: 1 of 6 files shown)


SirVer:
As disussed offline: Fixed commit message for wrong pruning and clarified that we do not loose the sorting of the nodes by size.

SirVer:
@wally-the-cartographer merge

wally-the-cartographer:

PR #141, opened 2018-06-15
Call get_visible_nodes in a background thread in the sdl viewer to never block forground.

SirVer:


SirVer:
@wally-the-cartographer merge

wally-the-cartographer:

PR #142, opened 2018-06-15
Remove width & height from get_visible_nodes API.

SirVer:


SirVer:
Depends on #141.

src/octree/mod.rs
```diff
@@ -73,10 +71,7 @@ fn size_in_pixels(
     ] {
         rv = rv.grow(project(matrix, p));
     }
-    Vector2::new(
```
thomasschiwietz: Did you think about the case if the aspect ratio (width / height) is not close to 1.0? I think the new approach is only correct if the aspect ratio is close to 1.0. How about replacing the width, height parameters by an aspect ratio parameter and adjust the normalized width or height accordingly?

SirVer: I thought about this, but felt that it was not adding a ton to the heuristic - I mean at worst the aspect ration will be like 2/3 or so in which case this should still be a reasonable approximation. Do you think it makes sense?

thomasschiwietz: Ok. Could put a comment in the code about it?

SirVer: done.

wally-the-cartographer:

PR #146, opened 2018-08-10
Actually stream points in the grpc server.

SirVer:
Instead of reading them all into memory.

### DIFF
--- a/point_viewer_grpc/src/bin/octree_server.rs
+++ b/point_viewer_grpc/src/bin/octree_server.rs
@@ -10,7 +10,7 @@ extern crate protobuf;
 
 use cgmath::Point3;
 use collision::Aabb3;
-use futures::{stream, Stream, Future, Sink};
+use futures::{Stream, Future, Sink};
 use futures::sync::oneshot;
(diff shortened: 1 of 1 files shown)


point_viewer_grpc/src/bin/octree_server.rs
```diff
@@ -69,48 +70,63 @@ impl proto_grpc::Octree for OctreeService {
...
+                }
+            });
         });
-        let f = resp.send_all(stream::iter_ok::<_, grpcio::Error>(replies))
+
+        // TODO(sirver): I did not figure out how to return meaningful errors. At least we return
+        // any error.
+        let rx = rx.map_err(|_| grpcio::Error::RemoteStopped);
+        let f = resp
+            .send_all(rx)
```
thomasschiwietz: Could you explain this line to me, please?
The thread is sending reply messages through to channel back to receiver. Is the GRPC sent after all reply messages have been received or is it streaming?

SirVer: added a couple of comments. ptal.

SirVer:
@wally-the-cartographer merge

wally-the-cartographer:

PR #147, opened 2018-08-24
Add frustum query for octree

ksavinash9:
### Summary
* Add frustum query for octree

### Tests
* Manual

### DIFF
--- a/point_viewer_grpc/src/bin/octree_server.rs
+++ b/point_viewer_grpc/src/bin/octree_server.rs
@@ -8,7 +8,7 @@ extern crate point_viewer;
 extern crate point_viewer_grpc;
 extern crate protobuf;
 
-use cgmath::{Frustum, Matrix4, PerspectiveFov, Point3, Transform}
+use cgmath::{Deg, Matrix4, PerspectiveFov, Point3, Rad, Transform, Vector3};
 use collision::Aabb3;
 use futures::{Stream, Future, Sink};
 use futures::sync::oneshot;
@@ -70,58 +70,88 @@ impl proto_grpc::Octree for OctreeService {
         req: proto::GetPointsInFrustumRequest,
         resp: ServerStreamingSink<proto::GetPointsInFrustumReply>,
     ) {
-        let projection_matrix = Matrix4::from(PerspectiveFov {
-            fov_y: req.fov_y,
-            aspect: req.aspect,
-            z_near:req.z_near,
-            z_far: req.z_far,
-        });
-        let view_position = Point3::new(req.view_position.x, req.view_position.y, req.view_position.z);
-        // view_direction: unit vector defining the direction vector
-        let view_direction = Vector3::new(req.view_direction.x, req.view_direction.y, req.view_direction.z);
-        // view_up: unit vector defining the up vector
-        let view_up = Vector3::new(req.view_up.x, req.view_up.y, req.view_up.z);
-        let view_center = view_position + view_up;
+        use std::thread;
 
-        let view_transform = Transform.look_at(view_position, view_center, view_up);
-        let view_matrix: Matrix4<f32> = view_transform.inverse_transform().unwrap().into();
-        let frustum_matrix = projection_matrix * view_matrix;
-
+        // This creates a async-aware (tx, rx) pair that can wake up the event loop when new data
+        // is piped through it.
+        let (tx, rx) = mpsc::channel(4);
         let octree = self.octree.clone();
-        let now = ::std::time::Instant::now();
-        // TODO(ksavinash9): Extract get_points_from_transformtion() as back end function for
-        // get_points_in_box() and get_points_in_frustum().
-        let visible_nodes = octree.get_visible_nodes(&frustum_matrix);
-        println!(
-            "Currently visible nodes: {}, time to calculate: {:?}",
-            visible_nodes.len(),
-            now.elapsed()
-        );
+        thread::spawn(move || {
+            // This is the secret sauce connecting an OS thread to a event-based receiver. Calling
+            // wait() on this turns the event aware, i.e. async 'tx' into a blocking 'tx' that will
+            // make this thread block when the event loop is not quick enough with piping out data.
+            let mut tx = tx.wait();
 
-        let mut reply = proto::GetPointsInFrustumReply::new();
-        let frustum = Frustum::from_matrix4(projection_matrix);
-        visible_nodes.for_each(|node| {
-            // TODO(ksavinash9): multi-thread this loop.
-            let node_iterator = match octree::NodeIterator::from_disk(self.meta, &node) {
-                Ok(node_iterator) => node_iterator,
-                Err(Error(ErrorKind::NodeNotFound, _)) => continue,
-                Err(err) => return Err(err),
+            let projection_matrix = Matrix4::from(PerspectiveFov {
+                fovy: Rad::from(Deg((req.fov_y as f32).into())),
+                aspect: (req.aspect as f32).into(),
+                near: (req.z_near as f32).into(),
+                far: (req.z_far as f32).into(),
+            });
+            let req_view_position = req.view_position.clone().unwrap();
+            let req_view_direction = req.view_position.clone().unwrap();
+            let req_view_up = req.view_position.clone().unwrap();
+
+            let view_position = Point3::new(req_view_position.x, req_view_position.y, req_view_position.z);
+            // view_direction: unit vector defining the direction vector
+            let view_direction = Vector3::new(req_view_direction.x, req_view_direction.y, req_view_direction.z);
+            // view_up: unit vector defining the up vector
+            let view_up = Vector3::new(req_view_up.x, req_view_up.y, req_view_up.z);
+            let view_center = view_position + view_up;
+
+            let view_transform: Matrix4<f32> = Transform::look_at(view_position, view_center, view_up);
+            let view_matrix: Matrix4<f32> = view_transform.inverse_transform().unwrap().into();
+            let frustum_matrix = projection_matrix * view_matrix;
+
+            let now = ::std::time::Instant::now();
+            // TODO(ksavinash9): Extract get_points_from_transformtion() as back end function for
+            // get_points_in_box() and get_points_in_frustum().
+            let visible_nodes = octree.get_visible_nodes(&frustum_matrix);
+            println!(
+                "Currently visible nodes: {}, time to calculate: {:?}",
+                visible_nodes.len(),
+                now.elapsed()
+            );
+
+            let mut reply = proto::GetPointsInFrustumReply::new();
+
+            let bytes_per_point = {
+                let mut reply = proto::GetPointsInFrustumReply::new();
+                let initial_proto_size = reply.compute_size();
+                let mut v = point_viewer::proto::Vector3f::new();
+                v.set_x(1.);
+                v.set_y(1.);
+                v.set_z(1.);
+                reply.mut_points().push(v);
+                let final_proto_size = reply.compute_size();
+                final_proto_size - initial_proto_size
             };
 
-            let mut points = Vec::with_capacity(node_iterator.size_hint().unwrap());
-            node_iterator.for_each(|p| points.push((*p).clone()));
-            points.for_each(|p| {
-                let child_relation = frustum.contains(&p);
-                if child_relation == Relation::In {
-                    let mut v = point_viewer::proto::Vector3f::new();
-                    v.set_x(p.position.x);
-                    v.set_y(p.position.y);
-                    v.set_z(p.position.z);
-                    reply.mut_points().push(v);
+            // Proto message must be below 4 MB.
+            let max_message_size = 4 * 1024 * 1024;
+            let mut reply_size = 0;
+            octree.points_in_frustum(&frustum_matrix).for_each(|p| {
+                let mut v = point_viewer::proto::Vector3f::new();
+                v.set_x(p.position.x);
+                v.set_y(p.position.y);
+                v.set_z(p.position.z);
+                reply.mut_points().push(v);
+
+                reply_size += bytes_per_point;
+                if reply_size > max_message_size - bytes_per_point {
+                    tx.send((reply.clone(), WriteFlags::default())).unwrap();
+                    reply.mut_points().clear();
+                    reply_size = 0;
                 }
             });
-        }
-        reply
+        });
+
+        let rx = rx.map_err(|_| grpcio::Error::RemoteStopped);
+        let f = resp
+            .send_all(rx)
+            .map(|_| {})
+            .map_err(|e| println!("failed to reply: {:?}", e));
+        ctx.spawn(f)
     }
 
     fn get_points_in_box(
--- a/point_viewer_grpc/src/bin/octree_server.rs
+++ b/point_viewer_grpc/src/bin/octree_server.rs
@@ -30,6 +30,38 @@ struct OctreeService {
     meta: point_viewer::proto::Meta,
 }
 
+pub trait GetBytes {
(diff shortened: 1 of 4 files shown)


ksavinash9:
I signed it!

ksavinash9:
@SirVer @thomasschiwietz Please, review and provide feedbacks.

thomasschiwietz:
We paired on this PR. Does it make sense for me to review?

SirVer:
@thomasschiwietz Yes, it makes sense. In fact I would say that is sufficient for checking it in if you guys are happy.

I also suggest showing the API to the customers and ask for their opinion.

src/octree/mod.rs
```diff
@@ -214,6 +214,11 @@ impl OnDiskOctree {
     }
 }
 
+// TODO (ksavinash9) trait implementation
```
thomasschiwietz: Remove this code.

point_viewer_grpc/src/bin/octree_server.rs
```diff
@@ -64,6 +64,61 @@ impl proto_grpc::Octree for OctreeService {
...
+        let view_position = Point3::new(req.view_position.x, req.view_position.y, req.view_position.z);
+        let view_direction = Vector3::new(req.view_direction.x, req.view_direction.y, req.view_direction.z);
+        let view_up = Vector3::new(req.view_up.x, req.view_up.y, req.view_up.z);
+        let view_center = view_position + view_up;
+
+        let view_transform = Transform.look_at(view_position, view_center, view_up);
+        let view_matrix: Matrix4<f32> = view_transform.inverse_transform().unwrap().into();
+        let frustum_matrix = projection_matrix * view_matrix;
+
+        let octree = self.octree.clone();
```
thomasschiwietz: I think this is fine for the first PR. Later, we should extract everything from here into a function that accepts the matrix and returns the points. This way we can unify the code path for Frustum and AABB.

ksavinash9: Thanks for the help and review @thomasschiwietz

point_viewer_grpc_proto_rust/src/proto.proto
```diff
@@ -51,3 +53,17 @@ message GetPointsInBoxRequest {
 message GetPointsInBoxReply {
   repeated point_viewer.proto.Vector3f points = 1;
 }
+
+message GetPointsInFrustumRequest {
+  point_viewer.proto.Vector3f view_position = 1;
+  point_viewer.proto.Vector3f view_direction = 2;
+  point_viewer.proto.Vector3f view_up = 3;
```
aliksey: What does this parameter mean?
And the next one, fov_y, in radians?

point_viewer_grpc/src/bin/octree_server.rs
thomasschiwietz: Please add a TODO() at this line: Extract get_points_from_transformtion() as back end for get_points_in_box() and get_points_in_frustum().

point_viewer_grpc/src/bin/octree_server.rs
```diff
@@ -64,6 +64,61 @@ impl proto_grpc::Octree for OctreeService {
...
+        println!(
+            "Currently visible nodes: {}, time to calculate: {:?}",
+            visible_nodes.len(),
+            now.elapsed()
+        );
+
+        let mut reply = proto::GetPointsInFrustumReply::new();
+        let frustum = Frustum::from_matrix4(projection_matrix);
+        visible_nodes.for_each(|node| {
+            let node_iterator = match octree::NodeIterator::from_disk(self.meta, &node) {
```
thomasschiwietz: Please add a comment here:
// TODO(ksavinash9): multi-thread this loop.

point_viewer_grpc_proto_rust/src/proto.proto
ksavinash9: @aliksey 
view_up is the direction of up vector
fov_y is in degrees so it would be degrees*(pi/180) radians

aliksey: Yes, thank you

thomasschiwietz: The unit of fov_y is radians. Please rename the variable to fov_y_radians.

thomasschiwietz: Please add a comment that view_up is a unit vector.

point_viewer_grpc_proto_rust/src/proto.proto
```diff
@@ -51,3 +53,17 @@ message GetPointsInBoxRequest {
 message GetPointsInBoxReply {
   repeated point_viewer.proto.Vector3f points = 1;
 }
+
+message GetPointsInFrustumRequest {
+  point_viewer.proto.Vector3f view_position = 1;
+  point_viewer.proto.Vector3f view_direction = 2;
```
thomasschiwietz: Please add a comment that view_direction is a unit vector.

SirVer:
This code currently does not compile:

~~~
 close delimiter: `}`
   --> point_viewer_grpc/src/bin/octree_server.rs:123:5
    |
123 |     }
    |     ^
    |
note: unclosed delimiter
   --> point_viewer_grpc/src/bin/octree_server.rs:101:31
    |
101 |         visible_nodes.for_each(|node| {
    |                               ^
error: expected `;`, found `use`
  --> point_viewer_grpc/src/bin/octree_server.rs:12:1
   |
12 | use collision::Aabb3;
   | ^^^ expected `;`
error: aborting due to 2 previous errors
error: Could not compile `point_viewer_grpc
~~~

Also did you have plans to test this?

ksavinash9:
@SirVer @thomasschiwietz Please review and provide feedbacks.
collision and cgmath crates don't provide a way to detect points in frustum. Also, I couldn't find a way to compare a point with 6 planes of frustum. So, I have used clipping matrix approach.

Is that the correct way of verifying points in frustum?
Should I implement trait for Matrix4 or Frustum?

ksavinash9:
Nikolai is an exceptional Rust developer. Hence, kindly review @nnmm

point_viewer_grpc/src/bin/octree_server.rs
```diff
@@ -64,6 +64,87 @@ impl proto_grpc::Octree for OctreeService {
...
+        &self,
+        ctx: RpcContext,
+        req: proto::GetPointsInFrustumRequest,
+        resp: ServerStreamingSink<proto::GetPointsInFrustumReply>,
+    ) {
+        use std::thread;
+
+        // This creates a async-aware (tx, rx) pair that can wake up the event loop when new data
+        // is piped through it.
+        let (tx, rx) = mpsc::channel(4);
```
nnmm: What does the 4 do?

point_viewer_grpc/src/bin/octree_server.rs
```diff
@@ -64,6 +64,87 @@ impl proto_grpc::Octree for OctreeService {
...
+                far: (req.z_far as f32).into(),
+            });
+            let req_view_position = req.view_position.clone().unwrap();
+            let req_view_direction = req.view_position.clone().unwrap();
+            let req_view_up = req.view_position.clone().unwrap();
+
+            let view_position = Point3::new(req_view_position.x, req_view_position.y, req_view_position.z);
+            // view_direction: unit vector defining the direction vector
+            let _view_direction = Vector3::new(req_view_direction.x, req_view_direction.y, req_view_direction.z);
+            // view_up: unit vector defining the up vector
```
nnmm: Why does this one have an underscore?

point_viewer_grpc/src/bin/octree_server.rs
```diff
@@ -64,6 +64,87 @@ impl proto_grpc::Octree for OctreeService {
...
+            let view_up = Vector3::new(req_view_up.x, req_view_up.y, req_view_up.z);
+            let view_center = view_position + view_up;
```
nnmm: How is the view_center defined (how is it different from view_position)?

point_viewer_grpc/src/bin/octree_server.rs
```diff
@@ -64,6 +64,87 @@ impl proto_grpc::Octree for OctreeService {
...
+
+            let view_transform: Matrix4<f32> = Transform::look_at(view_position, view_center, view_up);
+            let view_matrix: Matrix4<f32> = view_transform.inverse_transform().unwrap().into();
+            let frustum_matrix = projection_matrix * view_matrix;
+
+            let mut reply = proto::GetPointsInFrustumReply::new();
+
+            let bytes_per_point = {
```
nnmm: I feel like this part contains lots of duplicated code from points_in_box below. You could probably extract the common elements into a separate function.

point_viewer_grpc/src/bin/octree_server.rs
```diff
@@ -64,6 +64,87 @@ impl proto_grpc::Octree for OctreeService {
...
+            let req_view_up = req.view_position.clone().unwrap();
+
+            let view_position = Point3::new(req_view_position.x, req_view_position.y, req_view_position.z);
+            // view_direction: unit vector defining the direction vector
+            let _view_direction = Vector3::new(req_view_direction.x, req_view_direction.y, req_view_direction.z);
+            // view_up: unit vector defining the up vector
+            let view_up = Vector3::new(req_view_up.x, req_view_up.y, req_view_up.z);
+            let view_center = view_position + view_up;
+
+            let view_transform: Matrix4<f32> = Transform::look_at(view_position, view_center, view_up);
```
nnmm: A comment on what that does would be appreciated. According to the cgmath docs, "Create a transformation that rotates a vector to look at center from eye, using up for orientation." – I'm not sure I understand what that means. What does this do with a vector?

point_viewer_grpc_proto_rust/src/proto.proto
```diff
@@ -51,3 +53,17 @@ message GetPointsInBoxRequest {
 message GetPointsInBoxReply {
   repeated point_viewer.proto.Vector3f points = 1;
 }
+
+message GetPointsInFrustumRequest {
```
nnmm: Is this the most natural representation of a frustum? Personally, I'd find a Matrix4d for the view frame (view_position + view_direction + view_up) more natural, and separate horizontal_fov and vertical_fov expressed as tan(φ) preferable:
```
\    |____/
 \   | φ /
  \  |  /
   \ | /
    \|/
```
Hope the drawing helps.

src/octree/mod.rs
```diff
@@ -115,6 +115,47 @@ impl<'a> InternalIterator for PointsInBoxIterator<'a> {
...
+// https://stackoverflow.com/questions/6301085/how-to-check-if-an-object-lies-outside-the-clipping-volume-in-opengl
+//
+// TODO(ksavinash9) check with @Sirver if contains should be a trait for Matrix4
+fn contains(
+    projection_matrix: &Matrix4<f32>,
+    point: &Point3<f32>,
+) -> bool {
+    let v = Vector4::new(point.x, point.y, point.z, 1.);
+    let clip_v = projection_matrix * v;
+    return f32::abs(clip_v.x) < clip_v.w &&
```
nnmm: You should also be able to do `clip_v.x.abs()`

point_viewer_grpc_proto_rust/src/proto.proto
SirVer: The matrix is not constraint though: it might represent another kind of frustum then a view frustum. The API as presented is harder to misuse I think.

We will add a function that takes a generic Matrix eventually.

src/octree/mod.rs
```diff
@@ -80,8 +80,8 @@ fn relative_size_on_screen(bounding_cube: &Cube, matrix: &Matrix4<f32>) -> f32 {
 
 #[derive(Debug)]
 pub struct OnDiskOctree {
-    meta: OctreeMeta,
-    nodes: FnvHashMap<NodeId, NodeMeta>,
+    pub meta: OctreeMeta,
```
SirVer: is making these pub this necessary?

src/octree/mod.rs
```diff
@@ -115,6 +115,47 @@ impl<'a> InternalIterator for PointsInBoxIterator<'a> {
...
+                f(p);
+            });
+        }
+    }
+}
+
+// TODO(ksavinash9) confirm with @Sirver and @thomasschiwietz about clipping vector approach
+// https://stackoverflow.com/questions/6301085/how-to-check-if-an-object-lies-outside-the-clipping-volume-in-opengl
+//
+// TODO(ksavinash9) check with @Sirver if contains should be a trait for Matrix4
```
SirVer: Can you file a bug with `collision` about the missing API for checking if a point is inside a frustum?

SirVer: I'd say this is fine, no need for a trait.

src/octree/mod.rs
```diff
@@ -209,11 +250,25 @@ impl OnDiskOctree {
...
+            intersecting_nodes,
+        }
+    }
+
     pub fn bounding_box(&self) -> &Aabb3<f32> {
         &self.meta.bounding_box
     }
 }
 
+// TODO (ksavinash9) trait implementation for AABB3 and frustum maxtrix
```
SirVer: not really needed. remove.

point_viewer_grpc_proto_rust/src/proto.proto
nnmm: Ah, I'm not suggesting to pass a projection matrix, only a coordinate transform that determines which way the frustum is looking (including which way is up), and from where – you're right that passing any Matrix4 is overly general. How about translation + rotation quaternion? That's very close to what people are already working with.

Thoughts on the suggestion for the FOV units/representation?

point_viewer_grpc/src/bin/octree_server.rs
ksavinash9: Defines size of the buffer for channel.

point_viewer_grpc/src/bin/octree_server.rs
ksavinash9: It's not being used.

point_viewer_grpc/src/bin/octree_server.rs
ksavinash9: view_center: a vector to look at `center` from `eye` using `up` for orientation.

point_viewer_grpc/src/bin/octree_server.rs
ksavinash9: Thanks

src/octree/mod.rs
ksavinash9: Thanks

ksavinash9: Will do.

src/octree/mod.rs
ksavinash9: Thanks

src/octree/mod.rs
ksavinash9: Absolutely Not

point_viewer_grpc_proto_rust/src/proto.proto
SirVer: I have no preference here. I defer to @thomasschiwietz who knows the most about geometrical representation.

src/octree/mod.rs
SirVer: Please link in the code.

point_viewer_grpc/src/bin/octree_server.rs
SirVer: Add this comment.

point_viewer_grpc_proto_rust/src/proto.proto
thomasschiwietz: translation + quaternion sounds like a good idea for people working with poses. if there are no customers for position + dir + up then let's remove it. fov in radians.

PR #159, opened 2018-10-01
Update grpcio crate to 0.4.0

kerrick-lyft:
I've confirmed that I can build this locally with `cargo build --release`.

The Travis build is running but I'm guessing it will fail because it's failing on master currently: https://travis-ci.org/googlecartographer/point_cloud_viewer/builds/435799509

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -153,7 +153,7 @@ dependencies = [
  "cgmath 0.16.1 (registry+https://github.com/rust-lang/crates.io-index)",
  "num 0.1.42 (registry+https://github.com/rust-lang/crates.io-index)",
  "rand 0.4.2 (registry+https://github.com/rust-lang/crates.io-index)",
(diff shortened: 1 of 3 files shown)


ksavinash9:
We need to wait for @SirVer to confirm.

SirVer:
In general I am fine with this change. What triggered it though? I.e. why do you need it?

The main thing that is worrying is the update of avsoftware internal to Lyft around this. That might be more work. @kerrick-lyft Have you tried avsoftware with this change?

SirVer:
Master is failing because we expect protobuf 1.4.3, but grpcio mandates (and uses) protobuf 2. To unbreak master (and get this change in) we need to update protobuf usage from 1.4.3 to >= 2.0.0.

I filed #160 for this. @kerrick-lyft do you want to give this a shot?

SirVer:
Never mind, I looked into this today in #161. Closing this PR, as this is going into #161. Thanks for the fixes, Kerrick!

PR #163, opened 2018-10-09
Adds a new coloring made for X-Rays: color by z stddev.

SirVer:
2018-10-09-125220_2188x1581_scrot

### DIFF
--- a/xray/src/bin/build_xray_quadtree.rs
+++ b/xray/src/bin/build_xray_quadtree.rs
@@ -63,7 +63,7 @@ fn parse_arguments() -> clap::ArgMatches<'static> {
                 .required_if("coloring_strategy", "colored_with_intensity"),
             clap::Arg::with_name("max_stddev")
                 .help(
-                    "Maximum stddev for colored_with_height_stddev. Every stddev above this \
+                    "Maximum standard deviation for colored_with_height_stddev. Every stddev above this \
                      will be clamped to this value and appear saturated in the X-Rays. \
                      Only used for 'colored_with_height_stddev'.",
                 )
--- a/xray/src/bin/build_xray_quadtree.rs
+++ b/xray/src/bin/build_xray_quadtree.rs
@@ -61,6 +61,15 @@ fn parse_arguments() -> clap::ArgMatches<'static> {
                 .long("min_intensity")
                 .takes_value(true)
                 .required_if("coloring_strategy", "colored_with_intensity"),
+            clap::Arg::with_name("max_stddev")
+                .help(
+                    "Maximum stddev for colored_with_height_stddev. Every stddev above this \
+                     will be clamped to this value and appear saturated in the X-Rays. \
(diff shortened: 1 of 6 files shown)


xray/src/bin/build_xray_quadtree.rs
```diff
@@ -61,6 +61,15 @@ fn parse_arguments() -> clap::ArgMatches<'static> {
                 .long("min_intensity")
                 .takes_value(true)
                 .required_if("coloring_strategy", "colored_with_intensity"),
+            clap::Arg::with_name("max_stddev")
+                .help(
+                    "Maximum stddev for colored_with_height_stddev. Every stddev above this \
```
ksavinash9: Nit: Maximum standard deviation for colored_with_height_stddev?

standard deviation as stddev isn't very obvious to me. But, I could be wrong.

xray/src/generation.rs
```diff
@@ -29,6 +79,9 @@ pub enum ColoringStrategyKind {
 
     // Min and max intensities.
     ColoredWithIntensity(f32, f32),
+
+    // Colored in heat-map colors by stddev. Takes the max stddev to clamp on.
+    ColoredWithHeightStddev(f32),
```
ksavinash9: Nit: ColoredWithHeightSD?

xray/src/generation.rs
```diff
@@ -203,6 +258,38 @@ impl ColoringStrategy for PointColorColoringStrategy {
...
+    fn new(max_stddev: f32) -> Self {
+        HeightStddevColoringStrategy {
+            max_stddev, 
+            per_column_data: FnvHashMap::default(),
+        }
+    }
+}
+
+impl ColoringStrategy for HeightStddevColoringStrategy {
+    fn process_discretized_point(&mut self, p: &Point, x: u32, y: u32, _: u32) {
```
ksavinash9: Except XRayColoringStrategy, none of the strategy impl are using `z` argument.
How about args: ..., or tuples or Options<>?

xray/src/generation.rs
SirVer: I kept it since stddev seems more common to me.

xray/src/generation.rs
SirVer: I am not sure what you are asking for here.

ksavinash9: Am suggesting either we use variadic functions using a tuple
https://github.com/rust-lang/rfcs/issues/376

Or use Options, whichever is more Rustful.

SirVer: - variadic functions are not implemented in Rust so far. It seems it is also still quite controversial.
- Option seems wrong, because it is always passed by the caller, but not always used by the callee. Option is used if a value is sometimes passed, sometimes not.
- We cannot get rid of `z` because it is used by XRayColoringStrategy

SirVer:
@wally-the-cartographer merge

PR #169, opened 2018-11-05
Flip Y-Axis of X-Rays to get a right handed coordinate system.

SirVer:
This means that the images origin is now at the bottom left of the images, i.e. the images also use a right handed coordinate system.

Also a grab bag of unrelated other small refactorings:
- Unifying the way the two javascript directories are build.
- Add a few missing functions that are in the octree, but not the quadtree crate.
- Make the lowest level of the X-Rays use nearest interpolation, so that the actual resolution of the X-Rays is visible.
- Bump the X-Ray quadtree version, since the tiles are flipped.


octree_web_viewer/client/octree_viewer.ts
```diff
@@ -247,6 +247,10 @@ class NodeData {
   }
 }
 
+function now(): number {
+  return +new Date();
```
pdubroy: Same as above, you can use `performance.now()`.

octree_web_viewer/client/package.json
```diff
@@ -4,18 +4,23 @@
...
     "prettier": "^1.10.2",
+    "rollup": "^0.55.3",
+    "rollup-plugin-commonjs": "^8.3.0",
+    "rollup-plugin-node-resolve": "^3.0.2",
+    "typescript": "=2.7.2"
+  },
+  "dependencies": {
     "@types/dat-gui": "0.6.1",
-    "@types/three": "0.0.27",
+    "@types/three": "0.92.17",
```
pdubroy: Shouldn't the minor version match the version of three.js you're using?

xray/src/generation.rs
```diff
@@ -272,6 +272,50 @@ impl HeightStddevColoringStrategy {
     }
 }
 
+/// Build a parent image created of the 4 children tiles. All tiles are optionally, in which case
```
pdubroy: Nit: why three slashes? Is it a Rust thing? I've never seen that before.

octree_web_viewer/client/main.ts
```diff
@@ -14,10 +14,11 @@
 
 'use strict';
 
+import * as THREE from 'three';
 import { FirstPersonController } from './control';
 import { OctreeViewer } from './octree_viewer';
 
-export function now(): number {
+function now(): number {
   return +new Date();
```
pdubroy: `new Date().getTime()` would be clearer here. But you can also just use `performance.now()`.

octree_web_viewer/client/package.json
SirVer: Dunno. I copied that from our internal repositories. 0.95 does not exist for the types, so it seems I need to do this somehow.

xray/src/generation.rs
SirVer: https://doc.rust-lang.org/beta/rustdoc/what-is-rustdoc.html

Rust has something like doxygen build in.

SirVer:
> lgtm with some suggestions. Also, shouldn't the PR title say "y-axis" and not "x-axis"?

No, because we align X with X. To get a right handed coordinate system this means we align Y with minus Y.

SirVer:
@wally-the-cartographer merge

PR #170, opened 2018-11-12
Refactors to make external reuse easier and fix out of files error.

### LINKED ISSUE #121: Octree generation can complain about too many open files
This seems rare, only happened once so far. This fixed it: `ulimit -S -n 2048`.

Let's either figure out how to work around this or add this command and its implications to the README.

SirVer:
- Refactor to reuse xray generation code & octree generation as a library with providing our own scoped pool instance. 
- Reformat frontend code to reduce diff to internal version.
- Attempt to fix #121.


SirVer:
@wally-the-cartographer merge

PR #172, opened 2018-11-23
xray: extract out functionality for reading meta.pb from disk

pdubroy:
Extract out a helper function for deserializing meta.pb from the file system

### DIFF
--- a/xray/src/lib.rs
+++ b/xray/src/lib.rs
@@ -23,12 +23,27 @@ use cgmath::Point2;
 use fnv::FnvHashSet;
 use quadtree::{NodeId, Rect};
 use std::fs::File;
-use std::io::{Cursor, Read};
+use std::io::BufReader;
 use std::path::Path;
+
+pub use xray_proto_rust::proto;
 
 pub const CURRENT_VERSION: i32 = 2;
 
-#[derive(Debug)]
+pub fn read_meta_proto_from_file<P: AsRef<Path>>(filename: P) -> proto::Meta {
+    let f = File::open(&filename).expect("Could not open meta.pb");
+    let meta = protobuf::parse_from_reader::<proto::Meta>(&mut BufReader::new(f))
+        .expect("Could not parse meta.pb");
+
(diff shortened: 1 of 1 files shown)


xray/src/lib.rs
```diff
@@ -23,12 +23,27 @@ use cgmath::Point2;
...
+
 pub const CURRENT_VERSION: i32 = 2;
 
-#[derive(Debug)]
+pub fn read_meta_proto_from_file<P: AsRef<Path>>(filename: P) -> proto::Meta {
+    let f = File::open(&filename).expect("Could not open meta.pb");
+    let meta = protobuf::parse_from_reader::<proto::Meta>(&mut BufReader::new(f))
+        .expect("Could not parse meta.pb");
+
+    assert!(meta.version == CURRENT_VERSION,
```
SirVer: Can you add a TODO that this should return an error instead of crashing the program? This is no longer just a binary, but morphs into a library.

xray/src/lib.rs
```diff
@@ -23,12 +23,27 @@ use cgmath::Point2;
...
-use std::io::{Cursor, Read};
+use std::io::BufReader;
 use std::path::Path;
 
+pub use xray_proto_rust::proto;
+
 pub const CURRENT_VERSION: i32 = 2;
 
-#[derive(Debug)]
+pub fn read_meta_proto_from_file<P: AsRef<Path>>(filename: P) -> proto::Meta {
```
SirVer: nit: I prefer the newer way of writing this: `pub fn read_meta_proto_from_file(filename: impl AsRef<Path>) -> proto::Meta`. Both are equivalent in this case, but I find the newer to be easier to understand.

xray/src/lib.rs
```diff
@@ -76,5 +76,3 @@ impl Meta {
 
 pub mod backend;
```
SirVer: move these up as well so that all pubs are at the top?

SirVer:
seems this is no longer of interest. closing.

PR #174, opened 2018-11-28
Allow camera poses to be saved if the octree is served locally.

SirVer:


sdl_viewer/src/lib.rs
```diff
@@ -271,6 +277,41 @@ impl PointCloudRenderer {
...
+    states: Vec<camera::State>,
+}
+
+fn save_camera(index: usize, pose_path: &Option<PathBuf>, camera: &Camera) {
+    if pose_path.is_none() {
+        println!("Not serving from a local directory. Cannot save camera.");
+        return;
+    }
+    assert!(index < 10);
+    let mut states = ::std::fs::read_to_string(pose_path.as_ref().unwrap()).and_then(|data| {
```
pdubroy: Nit: `use std::fs;`, as you do for cmp, io, thread, etc. and then just `fs::read_to_string` here.

SirVer: Will do in the followup.

SirVer:
@wally-the-cartographer merge

wally-the-cartographer:

PR #175, opened 2018-11-29
Support the 3D Connexion space mice.

SirVer:


SirVer:
meow @tblaette

sdl_viewer/src/lib.rs
```diff
@@ -522,6 +533,17 @@ impl SdlViewer {
                 }
             }
 
+            if let Some(j) = joystick.as_ref() {
```
tblaette: what is the `.as_ref()` for?

SirVer: Joystick is a Some(Joystick). If I try to pattern match, the compiler tries to move it out of the Option. I do not want that, I just want a reference to it and keep it inside the Option. hence the as_ref().

SirVer:
@wally-the-cartographer merge

wally-the-cartographer:

PR #178, opened 2018-12-13
Extract get_nodes_for_level out into a separate, public function.

pdubroy:


xray/src/backend.rs
```diff
@@ -11,20 +10,31 @@ use std::io;
...
 
+impl NodeMeta {
+    pub fn id(&self) -> String {
+        self.id.clone()
+    }
+
+    pub fn bounding_rect(&self) -> BoundingRect {
+        self.bounding_rect.clone()
+    }
+}
```
tblaette: do you need these two getter functions? Looks like they are never called.

xray/src/backend.rs
```diff
@@ -109,82 +122,88 @@ struct HandleNodesForLevel {
     meta: Arc<Meta>,
 }
 
+pub fn get_nodes_for_level(
```
tblaette: what about putting this as a member function of `Meta` in `point_cloud_viewer/xray/src/lib.rs`?
Also, this could be named `get_nodes_in_frustum_for_level()` or something similar.

xray/src/backend.rs
pdubroy: I do need them, but I will instead just make the fields public.

xray/src/backend.rs
pdubroy: Ok, I will make it a method. As discussed I will just leave the name as is for now, as I think we want to change the implementation to take a bounding box rather than a frustum.

pdubroy:
@wally-the-cartographer merge

wally-the-cartographer:

PR #179, opened 2018-12-14
Add all_points rpc query and benchmark for streaming points

nnmm:
Useful for performance optimization. The GetAllPoints query is helpful because it's a common use case and just using a huge bounding box is hacky. Furthermore, in the benchmark we don't know which pointcloud, and thus bounding box, will be used.

### DIFF
--- a/point_viewer_grpc/src/bin/octree_benchmark.rs
+++ b/point_viewer_grpc/src/bin/octree_benchmark.rs
@@ -60,7 +60,7 @@ fn main() {
             .expect("octree_directory not given"),
     );
     let num_points = u64::from_str(matches.value_of("num-points").unwrap_or("50000000"))
-        .expect("Could not num-points");
+        .expect("num-points needs to be a number");
     if matches.is_present("no-client") {
         server_benchmark(octree_directory, num_points)
     } else {
@@ -70,7 +70,10 @@ fn main() {
 }
 
 fn server_benchmark(octree_directory: PathBuf, num_points: u64) {
-    let octree = OnDiskOctree::new(&octree_directory).unwrap();
+    let octree = OnDiskOctree::new(&octree_directory).expect(&format!(
+        "Could not create octree from '{}'",
+        octree_directory.display()
+    ));
     let mut counter: u64 = 0;
     octree.all_points().for_each(|_p: &Point| {
         if counter % 1000000 == 0 {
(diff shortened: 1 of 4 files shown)


point_viewer_grpc/src/bin/octree_benchmark.rs
```diff
@@ -0,0 +1,112 @@
...
+        ])
+        .get_matches();
+
+    let octree_directory = PathBuf::from(
+        matches
+            .value_of("octree_directory")
+            .expect("octree_directory not given"),
+    );
+    let num_points = u64::from_str(matches.value_of("num-points").unwrap_or("50000000"))
+        .expect("Could not num-points");
```
tblaette: nit: give a proper error message

point_viewer_grpc/src/bin/octree_benchmark.rs
```diff
@@ -0,0 +1,112 @@
...
+    if matches.is_present("no-client") {
+        server_benchmark(octree_directory, num_points)
+    } else {
+        let port = value_t!(matches, "port", u16).unwrap_or(50051);
+        full_benchmark(octree_directory, num_points, port)
+    }
+}
+
+fn server_benchmark(octree_directory: PathBuf, num_points: u64) {
+    let octree = OnDiskOctree::new(&octree_directory).unwrap();
```
tblaette: do an `.expect("proper error message")` instead of `.unwrap()`

nnmm: Done

point_viewer_grpc/src/bin/octree_benchmark.rs
nnmm: Done

point_viewer_grpc/src/bin/octree_benchmark.rs
```diff
@@ -0,0 +1,112 @@
...
+    let env = Arc::new(Environment::new(1));
+    let ch = ChannelBuilder::new(env).connect(&format!("localhost:{}", port));
+    let client = OctreeClient::new(ch);
+
+    let req = proto::GetAllPointsRequest::new();
+    let receiver = client.get_all_points(&req).unwrap();
+
+    let mut counter: u64 = 0;
+
+    for rep in receiver.wait() {
```
nnmm: Do you know whether wait() is the right thing to do here?

tblaette:
@SirVer can you merge? Or can we do that, too?

SirVer:
@wally-the-cartographer merge

@tblaette you should be able to talk to wally as well.

SirVer:
@wally-the-cartographer merge

PR #180, opened 2018-12-18
Make Semantic Map Viewer request handlers public

tblaette:
## Description:

To allow for a more individualized handling of serving x-rays make the `iron` handlers public.


tblaette:
meow @SirVer

PR #181, opened 2018-12-19
Changed the GRPC client interface to a callback based one to support streaming

SirVer:
A more ergonomic interface would require async, which is more work that I am willing to put in currently.

Before, all points returned from gRPC where streamed into a Vec, which then was passed on to the consumers. Now the consumers need to provide a callback function that takes a bunch of Points (~100000 per round) and the function can return a bool to indicate if it wants to continue (true) or stop (false) being called with new points.

### DIFF
--- a/point_viewer_grpc/src/lib.rs
+++ b/point_viewer_grpc/src/lib.rs
@@ -43,21 +43,21 @@ pub struct GrpcOctree {
 }
 
 impl GrpcOctree {
-    pub fn new(addr: &str) -> Self {
+    pub fn new(addr: &str) -> Result<Self> {
         let env = Arc::new(EnvBuilder::new().build());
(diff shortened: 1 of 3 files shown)


point_viewer_grpc/src/lib.rs
```diff
@@ -88,11 +91,20 @@ impl GrpcOctree {
...
+                    return Err(grpcio::Error::QueueShutdown);
+                }
+                points.clear();
                 Ok(())
             })
             .wait()
-            .unwrap();
-        points
+            .map_err(|_| point_viewer::errors::ErrorKind::Grpc);
+        if result.is_err() && !interrupted {
```
ksavinash9: Not sure if I completely understand this

PR #182, opened 2019-01-02
[Web Viewer] Add listeners for mouse events only once

akovalev:
## Scope
Add listeners for mouse events only once to prevent ever-growing number of JS event listeners attached to `canvas` element.

## Implementation
Modify the constructor of `Control` so that listeners for both `mouseup` and  `mousemove` events are added as a part of initialisation logic. The current implementation implies dynamic adding and removal of those event listeners in the runtime right after `mousedown` and `mouseup` events are fired respectively, but the problem with that approach is that `addEventListener`/`removeEventListener` functions  are passed references to completely different handlers every time they are invoked which results in incorrect removal of event listeners, hence growing number of JS event listeners

### Metrics 
These are the metrics captured by means of Performance Monitor in Chrome during ~1 minute long session

_Before_
screenshot 2019-01-02 15 06 05

_After_
screenshot 2019-01-02 15 07 32


akovalev:
I signed it!

akovalev:
Having briefly looked at the implementation of [`Control`][1] in xray client, I believe a similar fix might be applicable to it as well...

[1]: https://github.com/googlecartographer/point_cloud_viewer/blob/master/xray/client/control.ts#L150-L159

SirVer:
@wally-the-cartographer merge

wally-the-cartographer:

PR #183, opened 2019-01-03
[Web Viewer] Refactor NodeLoader#load method

akovalev:
## Scope
Having read the code of `OctreeViewer` component, I realised that it was pretty hard to grasp what's going on in its `nodesUpdate` method. That's why I decided to abstract away the logic responsible for both splitting nodes into multiple batches and scheduling fetch requests so that there's no more than concurrent requests at a given moment of time into `NodeLoader#load` method.

## Implementation
In order to achieve above-mentioned goal, new class called `AsyncTaskQueue` was [introduced][1] and [integrated][2] into `NodeLoader`. I believe it's better to review this PR commit by commit.

Also I have slightly updated README as it took me some time to figure out how to start the web viewer's client locally.

[1]: 635a7f6
[2]: 853e1e3


octree_web_viewer/client/lib/async-tasks-queue.ts
```diff
@@ -0,0 +1,42 @@
...
+  private pendingCount: number;
+
+  constructor(
+    public maxConcurrentTasks: number
+  ) {
+    this.queue = [];
+    this.pendingCount = 0;
+  }
+
+  public addTask(fn: () => Promise<T>): Promise<T> {
```
pdubroy: I'd write this as:

```
public async addTask(fn: () => Promise<T>): T {
  ...
}
```

octree_web_viewer/client/octree_viewer.ts
```diff
@@ -79,7 +81,30 @@ class NodeRenderData {
...
+  private batchSize = 50;
+  private maxConcurrentRequests = 2;
+  private queue = new AsyncTasksQueue<void>(this.maxConcurrentRequests);
+
   public load(
+    scene: THREE.Scene,
+    material: THREE.ShaderMaterial,
+    nodes: NodeData[],
+    onNewNodeData: () => void
+  ): Promise<void[]> {
```
pdubroy: Is there a reason to return `void[]`? Assuming I understand this correctly, something like the following might be cleaner:

```
public async load(/* ... */) {
  await Promise.all(/* ... */);
}
```

octree_web_viewer/client/octree_viewer.ts
```diff
@@ -325,39 +347,11 @@ export class OctreeViewer {
 
   private nodesUpdate(nodeIds: string[]) {
```
pdubroy: Prefer to write this as:

```
private async nodesUpdate(nodeIds: string[]) {
  ...
}
```

octree_web_viewer/client/octree_viewer.ts
```diff
@@ -325,39 +347,11 @@ export class OctreeViewer {
...
-    this.handleNextBatch();
-    console.log(`nodeUpdate took ${performance.now() - start}ms.`);
-  }
-
-  private handleNextBatch() {
-    if (this.batches.length == 0 || this.currentlyLoading > 2) {
-      return;
-    }
-    this.currentlyLoading += 1;
     this.nodeLoader
```
pdubroy: ...then you should be able to write this as:

```
await this.nodeLoader.load(/* ... *);
onsole.log(`nodeUpdate took ${performance.now() - start}ms.`);
```

akovalev:
@pdubroy thanks for your feedback, I tried to incorporate it in 10d8417

PR #184, opened 2019-01-07
WIP: Sketches out a possible implementation for a layer approach for the octree.

SirVer:
It models the current octree as the required position layer + optional intensity (f32) & color (vec3u8).

Octree generation gets way slower with this, and the client is changed to a callback to a bunch of points (1e6) instead of a callback to every single point.

This misses some implementations (see NOCOM) and it does not add a builder API (to add new layers at generation time). It also does not contain mutability (i.e. changing some layers after generation time with new data), but these fits obviously into the design space now.


SirVer:
@feuerste Does this still have a purpose or can I close this from your point?

feuerste:
We are currently implementing layers (attributes) anyways, so we can close this one...

PR #185, opened 2019-01-07
Fix screen space ordering of nodes

thomasschiwietz:
- Replace the point projection code by the standard way to do it
- Clip the projected points to the (visible) hemicube
- Initialize the Aabb3 with a projected point of the bounding box. Previously, the (0,0,0) was included in every Aabb3
- Unrelated code formatings due to rustfmt


SirVer:
@wally-the-cartographer merge

wally-the-cartographer:

PR #187, opened 2019-01-07
New heuristic to keep frame rate constant while moving

thomasschiwietz:
The number of nodes that are drawn wile moving is increase/decreased dependent on a frame rate threshold.
Also, screen space node bounding box computation is simplified a bit.

### DIFF
--- a/sdl_viewer/src/lib.rs
+++ b/sdl_viewer/src/lib.rs
@@ -84,7 +84,7 @@ struct PointCloudRenderer {
     needs_drawing: bool,
     max_nodes_in_memory: usize,
     world_to_gl: Matrix4<f32>,
-    max_level_moving: usize,
+    max_nodes_moving: usize,
     show_octree_nodes: bool,
     node_views: NodeViewContainer,
     box_drawer: BoxDrawer,
@@ -137,7 +137,7 @@ impl PointCloudRenderer {
             gamma: 1.,
             get_visible_nodes_params_tx,
             get_visible_nodes_result_rx,
-            max_level_moving: 4,
+            max_nodes_moving: max_nodes_in_memory,
             needs_drawing: true,
             show_octree_nodes: false,
             max_nodes_in_memory,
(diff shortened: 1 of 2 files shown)


SirVer:
This also changes all other clients (not only visualization), i.e. octree client and so on which is not desirable.

sdl_viewer/src/lib.rs
```diff
@@ -196,31 +196,10 @@ impl PointCloudRenderer {
...
-            } else {
-                min_level_to_display = current;
-            }
-        }
-        filtered_visible_nodes = self.visible_nodes
-            .iter()
-            .filter(|id| id.level() <= min_level_to_display)
-            .collect();
-        assert!(filtered_visible_nodes.len() < self.max_nodes_in_memory);
+        let max_nodes_to_display =
```
SirVer: Add a comment that we use a heuristic to find the correct number of nodes to keep the framerate decent.

sdl_viewer/src/lib.rs
```diff
@@ -252,12 +231,13 @@ impl PointCloudRenderer {
...
-                    self.max_level_moving -= 1;
+                if fps < 20. {
+                    self.max_nodes_moving = (self.max_nodes_moving as f32 * 0.9) as usize;
                 }
-                if fps > 25. && self.max_level_moving < max_level_seen {
-                    self.max_level_moving += 1;
+                if fps > 25. && self.max_nodes_moving < self.max_nodes_in_memory {
+                    self.max_nodes_moving = (self.max_nodes_moving as f32 * 1.1) as usize;
                 }
+                println!("#sirver self.max_nodes_moving: {:#?}", self.max_nodes_moving);
```
SirVer: please remove this before submitting.

thomasschiwietz:
@wally-the-cartographer merge

wally-the-cartographer:
Merge requested by authorized user thomasschiwietz. Merge queue now has a length of 1.

PR #191, opened 2019-02-04
Have the elapsed time affecting the amount of movement and rotation.

negin-z:
- Capture the elapsed time between two consecutive rendered frames.
- Incorporate the elapsed time in calculating the movement distance and rotation.

### DIFF
--- a/sdl_viewer/src/camera.rs
+++ b/sdl_viewer/src/camera.rs
@@ -16,6 +16,7 @@ use cgmath::{Decomposed, Deg, InnerSpace, Matrix4, One, PerspectiveFov, Quaterni
              Rotation3, Transform, Vector3, Zero};
 use opengl;
 use std::f32;
+use time;
 
 #[derive(Debug)]
 pub struct Camera {
@@ -55,7 +56,7 @@ pub struct State {
 impl Camera {
     pub fn new(gl: &opengl::Gl, width: i32, height: i32) -> Self {
         let mut camera = Camera {
-            movement_speed: 1.5,
+            movement_speed: 10.,
             moving_backward: false,
             moving_forward: false,
             moving_left: false,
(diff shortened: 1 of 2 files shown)


sdl_viewer/src/camera.rs
```diff
@@ -152,7 +153,9 @@ impl Camera {
             self.pan += pan.normalize();
         }
 
-        const TURNING_SPEED : Rad<f32> = Rad(0.015);
+        let elapsed_seconds = elapsed.num_milliseconds() as f32 / 1000.;
```
thomasschiwietz: Replace by num_seconds()

sdl_viewer/src/camera.rs
```diff
@@ -207,13 +210,13 @@ impl Camera {
     }
 
     pub fn pan(&mut self, x: f32, y: f32, z: f32) {
```
thomasschiwietz: There are duplicated functions for rotation and panning:
-  mouse_drag_pan() and pan()
-  mouse_drag_rotate() and rotate(). 

The mouse_* functions use constants (100) inside of the function while the non-mouse functions use constants outside of the function. Could we unify this to two function such as rotate(up, around) and pan(x,y,z) and set all  constants from the call site? Specifically, I find it inconsistent that Joystick computes the constants at the call site while mouse does not.

sdl_viewer/src/camera.rs
negin-z: done!

sdl_viewer/src/camera.rs
negin-z: Created a ticket to follow-up on it as a clean-up task

negin-z:
@thomasschiwietz We need to merge this final commit. It reverts a code clean-up which breaks the navigation in sdl-viewer.

negin-z:
I signed it!

SirVer:
@wally-the-cartographer merge

wally-the-cartographer:

PR #192, opened 2019-02-05
Restrict type roots to avoid conflicts with types in parent node_modules

feuerste:
When the point cloud viewer is included in another repo as git subtree, the typescript compiler by default picks up types from all parent node_module directories, which can lead to conflicts and errors. This PR fixes these problems.


pdubroy:
@wally-the-cartographer merge

wally-the-cartographer:

PR #193, opened 2019-02-07
Octree refactoring to allow for more kinds of octree data sources.

feuerste:
This PR changes how an octree is constructed. The octree itself is not a trait anymore, but it contains an octree data provider trait. This is necessary in order to add other data sources such as cloud data storage.

### DIFF
--- a/octree_web_viewer/src/bin/points_web_viewer.rs
+++ b/octree_web_viewer/src/bin/points_web_viewer.rs
@@ -66,12 +66,12 @@ fn main() {
 
     // The actix-web framework handles requests asynchronously using actors. If we need multi-threaded
     // write access to the Octree, instead of using an RwLock we should use the actor system.
-    let octree: Arc<dyn octree::Octree> = {
-        let web_octree = match octree::OnDiskOctree::new(octree_directory) {
-            Ok(web_octree) => web_octree,
+    let octree: Arc<octree::Octree> = {
+        let octree = match octree::octree_from_directory(octree_directory) {
+            Ok(octree) => octree,
             Err(err) => panic!("Could not load octree: {}", err),
         };
-        Arc::new(web_octree)
+        Arc::new(octree)
     };
 
     let sys = actix::System::new("octree-server");
--- a/sdl_viewer/src/bin/sdl_viewer.rs
+++ b/sdl_viewer/src/bin/sdl_viewer.rs
@@ -18,8 +18,7 @@ fn main() {
     SdlViewer::new()
         .register_octree_factory("grpc:".into(), |p| {
             Ok(Box::new(
-                point_viewer_grpc::GrpcOctree::new(p.as_str())
-                    .expect("Could not connect to GRPC server"),
+                point_viewer_grpc::octree_from_address(p.as_str()).expect("Could not create octree."),
             ))
         })
         .run();
--- a/sdl_viewer/src/bin/sdl_viewer.rs
+++ b/sdl_viewer/src/bin/sdl_viewer.rs
@@ -18,7 +18,8 @@ fn main() {
     SdlViewer::new()
         .register_octree_factory("grpc:".into(), |p| {
             Ok(Box::new(
-                point_viewer_grpc::octree_from_address(p.as_str()).expect("Could not create octree."),
+                point_viewer_grpc::octree_from_address(p.as_str())
+                    .expect("Could not create octree."),
             ))
         })
         .run();
--- a/src/generation.rs
+++ b/src/generation.rs
@@ -22,6 +22,8 @@ use crate::{InternalIterator, Point};
 use cgmath::{EuclideanSpace, Point3};
 use collision::{Aabb, Aabb3};
 use fnv::{FnvHashMap, FnvHashSet};
+use math::Cube;
+use octree::{self, OnDiskOctreeDataProvider};
 use pbr::ProgressBar;
 use protobuf::Message;
 use scoped_pool::{Pool, Scope};
@@ -36,6 +38,7 @@ const MAX_POINTS_PER_NODE: i64 = 100000;
 
 // Return a list a leaf nodes and a list of nodes to be splitted further.
 fn split<P>(
+    octree_data_provider: &OnDiskOctreeDataProvider,
     octree_meta: &octree::OctreeMeta,
     node_id: &octree::NodeId,
     stream: P,
@@ -64,6 +67,7 @@ where
         let array_index = child_index.as_u8() as usize;
         if children[array_index].is_none() {
             children[array_index] = Some(octree::NodeWriter::new(
+                octree_data_provider,
                 octree_meta,
                 &node_id.get_child_id(child_index),
             ));
@@ -75,7 +79,7 @@ where
     // writing a point. This only saves some disk space during processing - all nodes will be
     // rewritten by subsampling the children in the second step anyways. We also ignore file
     // removing error. For example, we never write out the root, so it cannot be removed.
-    octree::NodeWriter::new(octree_meta, node_id);
+    octree::NodeWriter::new(octree_data_provider, octree_meta, node_id);
 
     let mut leaf_nodes = Vec::new();
     let mut split_nodes = Vec::new();
@@ -121,6 +125,7 @@ fn should_split_node(
 
 fn split_node<'a, P>(
     scope: &Scope<'a>,
+    octree_data_provider: &'a OnDiskOctreeDataProvider,
     octree_meta: &'a octree::OctreeMeta,
     node_id: &octree::NodeId,
     stream: P,
@@ -128,13 +133,19 @@ fn split_node<'a, P>(
 ) where
     P: InternalIterator,
 {
-    let (leaf_nodes, split_nodes) = split(octree_meta, node_id, stream);
+    let (leaf_nodes, split_nodes) = split(octree_data_provider, octree_meta, node_id, stream);
     for child_id in split_nodes {
         let leaf_nodes_sender_clone = leaf_nodes_sender.clone();
         scope.recurse(move |scope| {
-            let stream = octree::NodeIterator::from_disk(octree_meta, &child_id).unwrap();
+            let stream = octree::NodeIterator::from_data_provider(
+                octree_data_provider,
+                octree_meta,
+                &child_id,
+            )
+            .unwrap();
             split_node(
                 scope,
+                octree_data_provider,
                 octree_meta,
                 &child_id,
                 stream,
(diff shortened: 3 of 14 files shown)


feuerste:
@nnmm @SirVer FYI

SirVer:
I notice quite a lot of formatting changes in this. I suggest making a prelim PR that just runs `cargo fmt --all` over the code base so we have consistency going forward.

@feuerste ^^

feuerste:
> I notice quite a lot of formatting changes in this. I suggest making a prelim PR that just runs `cargo fmt --all` over the code base so we have consistency going forward.
> 
> @feuerste ^^

Here you go: #194

src/octree/mod.rs
```diff
@@ -80,18 +80,28 @@ fn relative_size_on_screen(bounding_cube: &Cube, matrix: &Matrix4<f32>) -> f32 {
...
+    fn meta_proto(&self) -> Result<proto::Meta>;
+    fn data(
+        &self,
+        node_id: &NodeId,
+        node_extensions: Vec<&str>,
+    ) -> Result<HashMap<String, Box<dyn Read>>>;
+    fn number_of_points(&self, node_id: &NodeId) -> Result<i64>;
+}
+
+pub struct Octree {
```
catevita: line 93 mod.rs: Is it enough that the trait OctreeDataProvider is Send + Sync? 
See line 75 in pointswebviewer.rs (client of web viewer)

catevita: question already answered

src/octree/node.rs
```diff
@@ -469,8 +469,12 @@ impl Drop for NodeWriter {
 }
 
 impl NodeWriter {
-    pub fn new(octree_meta: &OctreeMeta, node_id: &NodeId) -> Self {
-        let stem = node_id.get_stem(&octree_meta.directory);
+    pub fn new(
+        octree_data_provider: &OnDiskOctreeDataProvider,
```
catevita: i'd add a comment about NodeWriter being implemented only for OnDiskDataProvider

octree_web_viewer/src/bin/points_web_viewer.rs
```diff
@@ -73,7 +73,10 @@ fn main() {
     let octree_directory = PathBuf::from(matches.value_of("octree_directory").unwrap());
 
     let octree = {
-        let octree = match octree::OnDiskOctree::new(octree_directory) {
+        let data_provider = octree::OnDiskOctreeDataProvider {
```
nnmm: How about a more user-friendly interface: Octree::from_directory() that hides the data provider?

feuerste: Yes, good idea, I just added two free functions `octree_from_directory` and `octree_from_address`.

src/octree/mod.rs
```diff
@@ -238,13 +262,98 @@ impl OnDiskOctree {
...
+    }
+
+    pub fn get_node_data(&self, node_id: &NodeId) -> Result<NodeData> {
+        // TODO(hrapp): If we'd randomize the points while writing, we could just read the
+        // first N points instead of reading everything and skipping over a few.
+        let mut position_color_reads = self
+            .data_provider
+            .data(node_id, vec![node::POSITION_EXT, node::COLOR_EXT])?;
+
+        let get_data = |all_reads: &mut HashMap<String, Box<dyn Read>>,
```
tblaette: you actually don't have to pass in `all_reads` because the closure can access `position_color_reads` anyway. If you still want to pass it in because of readability, you can convert the closure into a normal function `fn get_data(...)` which can still be defined inside of `fn get_node_data()`. If a closure doesn't capture any variable from its environment, it's more of a matter of taste whether to use a closure or a function.

src/octree/mod.rs
```diff
@@ -267,6 +376,7 @@ impl OnDiskOctree {
             }
         }
         PointsInBoxIterator {
+            octree_data_provider: &*self.data_provider,
```
tblaette: Is the `&*` needed because you go from a `Box` to a normal reference? It looks kind of odd to me...

sdl_viewer/src/bin/sdl_viewer.rs
```diff
@@ -20,9 +20,11 @@ use sdl_viewer::SdlViewer;
 fn main() {
     SdlViewer::new()
         .register_octree_factory("grpc:".into(), |p| {
+            let octree_data_provider = point_viewer_grpc::GrpcOctreeDataProvider::new(p.as_str())
+                .expect("Could not connect to GRPC server.");
             Ok(Box::new(
-                point_viewer_grpc::GrpcOctree::new(p.as_str())
-                    .expect("Could not connect to GRPC server"),
+                point_viewer::octree::Octree::new(Box::new(octree_data_provider))
+                    .expect("Could not create octree."),
```
tblaette: you could call `octree_from_address()` instead

src/generation.rs
```diff
@@ -293,8 +308,11 @@ pub fn build_octree(
     let octree_meta = &octree::OctreeMeta {
         bounding_box,
         resolution,
+    };
+    let octree_data_provider = OnDiskOctreeDataProvider {
         directory: output_directory.as_ref().to_path_buf(),
     };
+    let octree_data_provider = &octree_data_provider;
```
tblaette: I would rather remove this line and write `&octree_data_provider` wherever this variable is used below.

nnmm:
Sorry for the late review, it's a huge PR so I'm having a bit of a hard time reading it all. What's the reason for passing the data provider everywhere instead of making it a member of OctreeMeta, where the directory was stored before?

I think creating a directory for the backends could be nice.

Also, could you comment on what structs/traits map to which structs? E.g. why is the Octree get_visible_nodes gone, but OctreeDataProvider gained a meta_proto method?

src/octree/node.rs
```diff
@@ -267,32 +254,45 @@ impl NodeMeta {
...
+        octree_meta: &OctreeMeta,
+        id: &NodeId,
+    ) -> Result<Self> {
         let bounding_cube = id.find_bounding_cube(&Cube::bounding(&octree_meta.bounding_box));
         let position_encoding = PositionEncoding::new(&bounding_cube, octree_meta.resolution);
-        let intensity_reader = File::open(&stem.with_extension(INTENSITY_EXT))
-            .map(|f| Some(BufReader::new(f)))
+        let intensity_reader = octree_data_provider
+            .data(id, vec![INTENSITY_EXT])
+            .map(|mut f| Some(BufReader::new(f.remove(INTENSITY_EXT).unwrap())))
```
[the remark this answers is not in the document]
feuerste: Changed it to catch this case.

src/octree/node.rs
```diff
@@ -469,8 +469,12 @@ impl Drop for NodeWriter {
...
 
 impl NodeWriter {
-    pub fn new(octree_meta: &OctreeMeta, node_id: &NodeId) -> Self {
-        let stem = node_id.get_stem(&octree_meta.directory);
+    pub fn new(
+        octree_data_provider: &OnDiskOctreeDataProvider,
+        octree_meta: &OctreeMeta,
+        node_id: &NodeId,
+    ) -> Self {
+        let stem = node_id.get_stem(&octree_data_provider.directory);
```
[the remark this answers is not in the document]
feuerste: Done.

src/octree/mod.rs
```diff
@@ -80,18 +80,28 @@ fn relative_size_on_screen(bounding_cube: &Cube, matrix: &Matrix4<f32>) -> f32 {
...
 }
 
-#[derive(Debug)]
-pub struct OnDiskOctree {
+pub trait OctreeDataProvider: Send + Sync {
+    fn meta_proto(&self) -> Result<proto::Meta>;
+    fn data(
+        &self,
+        node_id: &NodeId,
+        node_extensions: Vec<&str>,
```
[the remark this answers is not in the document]
feuerste: Done, also adjusted to use an enum now.

src/octree/mod.rs
```diff
@@ -349,90 +461,45 @@ fn maybe_push_node(
...
-        })
+        let file_size_bytes = file_meta_data_opt.unwrap().len();
+        // color has 3 bytes per point
+        Ok((file_size_bytes / 3) as i64)
     }
 }
+
+pub fn octree_from_directory(directory: impl AsRef<Path>) -> Result<Octree> {
+    let data_provider = OnDiskOctreeDataProvider {
+        directory: directory.as_ref().to_owned(),
```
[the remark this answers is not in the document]
feuerste: Done.

src/octree/mod.rs
```diff
@@ -80,18 +80,28 @@ fn relative_size_on_screen(bounding_cube: &Cube, matrix: &Matrix4<f32>) -> f32 {
...
+pub struct Octree {
+    data_provider: Box<dyn OctreeDataProvider>,
     meta: OctreeMeta,
     nodes: FnvHashMap<NodeId, NodeMeta>,
 }
 
-pub trait Octree: Send + Sync {
-    fn get_visible_nodes(&self, projection_matrix: &Matrix4<f32>) -> Vec<NodeId>;
-    fn get_node_data(&self, node_id: &NodeId) -> Result<NodeData>;
+pub struct OnDiskOctreeDataProvider {
```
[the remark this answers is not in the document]
feuerste: Done.

src/octree/mod.rs
```diff
@@ -218,7 +243,6 @@ impl OnDiskOctree {
         };
 
         let meta = OctreeMeta {
-            directory: directory.into(),
             resolution: meta_proto.resolution,
             bounding_box: bounding_box,
```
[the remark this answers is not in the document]
feuerste: Done.

src/octree/mod.rs
feuerste: Done, removed it from the parameters list.

src/octree/mod.rs
feuerste: Yes, it's needed here, otherwise the compiler complains...

sdl_viewer/src/bin/sdl_viewer.rs
feuerste: Done.

src/generation.rs
feuerste: It would only be possible if we cloned the data provider, since below it is moved into another scope, so I will rather go with the reference approach.

feuerste:
> Sorry for the late review, it's a huge PR so I'm having a bit of a hard time reading it all. What's the reason for passing the data provider everywhere instead of making it a member of OctreeMeta, where the directory was stored before?

This is actually a good idea to clean up interfaces. I was just reluctant because meta data for me is static data. It seems odd to have a dynamic data provider inside static data. If we renamed `OctreeMeta` to something more meaningful, I would be all for it. Do you have a suggestion for a good name?

> I think creating a directory for the backends could be nice.

For sure, I think this could go into a separate PR though.

> Also, could you comment on what structs/traits map to which structs? E.g. why is the Octree get_visible_nodes gone, but OctreeDataProvider gained a meta_proto method?

`get_visible_nodes` of `Octree` is still there, so no change here. The `meta_proto` function was added to the `OctreeDataProvider` to have a flexible means of creating the `Octree` with a meta proto coming from all kinds of sources, e.g. reading it from file, from S3, or via a GRPC call.

feuerste:
@nnmm I just looked into the code again regarding the changes we talked about yesterday (offline), namely

1. moving `nodes` from `Octree` into `OctreeMeta` and
2. passing `&Octree` instead of `&OctreeDataProvider` and `&OctreeMeta` to the `NodeIterator`

Unfortunately neither of them at the current state works nicely, as `build_octree` in `generation.rs` internally uses `NodeIterator` and `OctreeMeta`, too, but at that point of time (during octree generation) neither knows about a full `Octree` nor about its `nodes`. So I would suggest for now to leave the PR as it is and possibly address above changes in a later PR. What do you think?

feuerste:
@wally-the-cartographer merge

wally-the-cartographer:

wally-the-cartographer:
Refusing to merge, since the combined checks state is failure

feuerste:
@wally-the-cartographer merge

wally-the-cartographer:

PR #196, opened 2019-02-12
Make client-side cancelling a request not panic

nnmm:
# Summary
The `tx.send` will fail if the client cancels a request. This is a normal condition and panicking could confuse the user.
With this PR, the server ignores failing sends in all but the last iteration, where it will print a message.

### DIFF
--- a/point_viewer_grpc/src/service.rs
+++ b/point_viewer_grpc/src/service.rs
@@ -124,7 +124,7 @@ fn stream_points_back_to_sink(
 
                 reply_size += bytes_per_point;
                 if reply_size > max_message_size - bytes_per_point {
-                    tx.send((reply.clone(), WriteFlags::default())).unwrap();
+                    tx.send((reply.clone(), WriteFlags::default())).unwrap_or(());
                     reply.mut_positions().clear();
                     reply.mut_colors().clear();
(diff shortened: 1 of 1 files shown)


nnmm:
ptal @feuerste @SirVer

point_viewer_grpc/src/service.rs
```diff
@@ -124,7 +124,7 @@ fn stream_points_back_to_sink(
 
                 reply_size += bytes_per_point;
                 if reply_size > max_message_size - bytes_per_point {
-                    tx.send((reply.clone(), WriteFlags::default())).unwrap();
+                    tx.send((reply.clone(), WriteFlags::default())).unwrap_or(());
```
SirVer: common way to ignore errors is `let _ = tx.send()...`

But I am worried that this continues the loop - once send fails, this thread should stop its work.

feuerste:
Can you please add comments in the source code as well to explain the behavior?

point_viewer_grpc/src/service.rs
nnmm: I hadn't thought about that. Should I make the callback return a Result?

SirVer: sure.

PR #198, opened 2019-02-13
Make x-ray background transparent

linouk23:
### Desc
As a UI developer, I would like the background color of x-rays tiles to be transparent, so that I can easily overlay them on other map data.

### DIFF
--- a/xray/src/generation.rs
+++ b/xray/src/generation.rs
@@ -501,9 +501,8 @@ pub fn build_xray_quadtree(
                         tile_size_px,
                         image::FilterType::Lanczos3,
                     );
+                    let image = create_image_with_transparent_background(&image);
                     image
-                        .as_rgb8()
-                        .unwrap()
                         .save(&get_image_path(output_directory, node_id))
                         .unwrap();
                     node_id.parent_id().map(|id| tx_clone.send(id).unwrap());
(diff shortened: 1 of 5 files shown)


pdubroy:
If you only turn white pixels to transparent, I'm pretty sure we will see aliasing artifacts around the edges. Is it possible to instead make all pixels transparent? E.g., starting with setting `alpha` to 0 in `get_pixel_color`.

xray/src/generation.rs
```diff
@@ -315,11 +315,16 @@ pub fn build_parent(children: &[Option<image::RgbImage>]) -> image::RgbImage {
...
     let child_size_px = child_size_px.expect("No children passed to 'build_parent'.");
-    let mut large_image = image::RgbImage::from_pixel(
+    let mut large_image = image::RgbaImage::from_pixel(
         child_size_px * 2,
         child_size_px * 2,
-        image::Rgb {
-            data: [255, 255, 255],
+        image::Rgba {
+            data: [
+                TRANSPARENT.to_u8().red,
```
pdubroy: Can you make it a comment line option whether to make the background white or transparent?

pdubroy:
@wally-the-cartographer merge

wally-the-cartographer:

PR #203, opened 2019-02-15
Fix octree generation.

SirVer:
On octree generation, parent nodes are generated by subsampling from their children. The code is checking for all 8 potential children, expecting that the NodeIterator correctly returns `NodeNotFound` if a node does not exist.

However the OnDiskDataProvider returned `std::io::ErrorKind::NotFound` instead, which was taken as a hard error, ending octree generation.


SirVer:
@feuerste Please review, take over and shepherd into master.

feuerste:
@wally-the-cartographer merge

wally-the-cartographer:

PR #204, opened 2019-02-15
Change octree::NodeId to be a u128

SirVer:
This removes the limitation of octrees having a deepest level of `resolution*2**(64/3)`. This is ~2500m cube length for the default resolution of 1mm. 

This change modifies NodeId to be a single 128 bit unsigned integer. The highest 8 bit are the level, while the rest is used used for the index. This fits a max level of 120/3 = 40, which is ~1099512km (~86x the diameter of the earth) for the default resolution. In my tests this comes surprisingly without performance penalties on my machine (see end of PR description). 

The change is backwards compatible (i.e. existing octrees can be read without change), but the octree version is still bumped, since in the NodeId proto we no longer have (level, index) but instead (high_64bit, low_64bit).

This also adds a new binary `upgrade_octree` to update existing octrees to the new version.

Timing numbers, commands used:
~~~
target/release/build_octree /Users/hrapp/Downloads/pointcloud.ply --output_directory <dir>
target/release/octree_benchmark --no-client <dir> --num-points 100000000
~~~

Best runs out of 2 or 3, real time:

|               | Octree Generation | Octree benchmark | 
| ------------- | ------------- | ------------------ |
| master  |  11:57.19 | 0:06.42 |
| this work  |  12:10.94 | 0:05.94 | 

Detailed runbook in this gist: https://gist.github.com/SirVer/9ba8bd6ca02a1051d0c98ad3cef15383


SirVer:
@feuerste Could you please review and see if you agree with the approach taken. This change is needed to fix AVMAPPING-2010, but it is also crucial to deal with larger areas, i.e. it will bite us eventually if we do not do it.

feuerste:
I just sat together with @thomasschiwietz discussing these changes. We are reluctant to change the data format due to the following reasons:
- The current default resolution is 1 mm. What's the point of having cubes with a side length of an order of magnitude smaller than the accuracy of lidar? If we changed it to 5 cm, it would cover about 125 km, making it ready also for sharding
- Quadtree would need the same adjustments in order to be consistent
- All current octrees would need to be updated

SirVer:
> * The current default resolution is 1 mm. What's the point of having cubes with a side length of an order of magnitude smaller than the accuracy of lidar? If we changed it to 5 cm, it would cover about 125 km, making it ready also for sharding

Note that this is (not only) the resolution of the smallest cube, but also dictates the bits of the fixed point encoding in the leaf nodes. I have a bit of trouble thinking through in my head right now what upping to 5cm actually means for these.

> * Quadtree would need the same adjustments in order to be consistent

I disagree. The leaf nodes of the X-Rays are 10m, and only 2 (not 3) bits are needed for indexing the children, making 64 bit ample and sufficient for the whole world. No change needed.

> * All current octrees would need to be updated

No, this change is backwards compatible, all old octree's are still loading fine.

thomasschiwietz:
> > * The current default resolution is 1 mm. What's the point of having cubes with a side length of an order of magnitude smaller than the accuracy of lidar? If we changed it to 5 cm, it would cover about 125 km, making it ready also for sharding
> 
> Note that this is (not only) the resolution of the smallest cube, but also dictates the bits of the fixed point encoding in the leaf nodes. I have a bit of trouble thinking through in my head right now what upping to 5cm actually means for these.

Point quantization distance is 0.2 mm for 5 cm edge length using 8-bits (leaf nodes). I think this is more than sufficient.

> 
> > * Quadtree would need the same adjustments in order to be consistent
> 
> I disagree. The leaf nodes of the X-Rays are 10m, and only 2 (not 3) bits are needed for indexing the children, making 64 bit ample and sufficient for the whole world. No change needed.

Leaf node resolution in dense X-Rays is 5 cm and moving to 1 cm soon. This is about data format consistency.

> 
> > * All current octrees would need to be updated
> 
> No, this change is backwards compatible, all old octree's are still loading fine.

Good.

I think this patch is overshooting AVMAPPING-2010. The data set was broken. 
In my opinion we should set the default resolution to 5 cm and merge this patch if we require octrees bigger than 125 km^^3.

SirVer:
> Leaf node resolution in dense X-Rays is 5 cm and moving to 1 cm soon.

This is pixels, not leaf nodes. The leaf nodes are still 10m, I think?

Let's discuss this in person once I am back. I understand the concerns and we should do the right thing. But I feel this is not as clear cut, so I'd like to spend some time discussing together.

thomasschiwietz:
> > Leaf node resolution in dense X-Rays is 5 cm and moving to 1 cm soon.
> 
> This is pixels, not leaf nodes. The leaf nodes are still 10m, I think?

Right, 256 pixels * resolution

> Let's discuss this in person once I am back. I understand the concerns and we should do the right thing. But I feel this is not as clear cut, so I'd like to spend some time discussing together.

Good idea. I also have thoughts about the octree structure type in general.

catevita:
> 
> Good idea. I also have thoughts about the octree structure type in general.
Please let me know when you are discussing this topic!

SirVer:
Merged master and repushed. However after doing that, timing went up:

~~~
Command being timed: "target/release/build_octree /Users/hrapp/Downloads/pointcloud.ply --output_directory /Users/hrapp/Downloads/v2/octree_just_colors"
	User time (seconds): 1699.17
	System time (seconds): 3106.57
	Percent of CPU this job got: 406%
	Elapsed (wall clock) time (h:mm:ss or m:ss): 19:42.75
	Average shared text size (kbytes): 0
	Average unshared data size (kbytes): 0
	Average stack size (kbytes): 0
	Average total size (kbytes): 0
	Maximum resident set size (kbytes): 45208
	Average resident set size (kbytes): 0
	Major (requiring I/O) page faults: 67
	Minor (reclaiming a frame) page faults: 1477831
	Voluntary context switches: 457494
	Involuntary context switches: 30700709
	Swaps: 0
	File system inputs: 0
	File system outputs: 0
	Socket messages sent: 0
	Socket messages received: 0
	Signals delivered: 0
	Page size (bytes): 4096
	Exit status: 0
~~~

This needs another round of benchmarks against current master. because either this did regress in speed, or master did since I last benchmarked.

SirVer:
I did rerun the benchmarks on this sha and this time I got was 14mins:

~~~rust
	User time (seconds): 1705.38
	System time (seconds): 2040.02
	Percent of CPU this job got: 440%
	Elapsed (wall clock) time (h:mm:ss or m:ss): 14:10.53
	Average shared text size (kbytes): 0
	Average unshared data size (kbytes): 0
	Average stack size (kbytes): 0
	Average total size (kbytes): 0
	Maximum resident set size (kbytes): 45944
	Average resident set size (kbytes): 0
	Major (requiring I/O) page faults: 32
	Minor (reclaiming a frame) page faults: 1348331
	Voluntary context switches: 451845
	Involuntary context switches: 21911766
	Swaps: 0
	File system inputs: 0
	File system outputs: 0
	Socket messages sent: 0
	Socket messages received: 0
	Signals delivered: 0
	Page size (bytes): 4096
	Exit status: 0
~~~

This is in striking distance of the original performance, so this is good to go. 

It seems that performance is very flaky.

SirVer:
I have signed it.

feuerste:
@wally-the-cartographer merge

wally-the-cartographer:
Merge requested by authorized user feuerste. Merge queue now has a length of 2.

PR #206, opened 2019-02-18
fixing typeroots to include repo as a git subtree

catevita:
small change to allow a correct use of point_cloud_viewer as a submodule in git projects.
the same syntax is also in xrays/client

ptal! @feuerste


feuerste:
> ok!

You need to navigate to "Files changed", "Review changes", and approve there.

PR #208, opened 2019-02-18
Fix rotating with the mouse.

SirVer:
It was very slow due to the latest changes for constant rotation speed.

The mouse has a different quality though: you want to follow the mouse
movement as much as possible, i.e. the current rotation speed is
entirely ignored. This renames the delta_angle to rotation speed and
adds a new absolute rotation change request that is only used for mouse
drags.


SirVer:
@pdubroy Could you have a look and shepherd this into master?

Negin, fyi.

sdl_viewer/src/camera.rs
```diff
@@ -182,18 +204,24 @@ impl Camera {
...
-        if !self.delta_theta.is_zero() || !self.delta_phi.is_zero() {
+        if !self.rotation_speed.theta.is_zero()
+            || !self.rotation_speed.phi.is_zero()
+            || !self.delta_rotation.theta.is_zero()
+            || !self.delta_rotation.phi.is_zero()
+        {
             moved = true;
-            self.theta += self.delta_theta * elapsed_seconds;
-            self.phi += self.delta_phi * elapsed_seconds;
+            self.theta += self.rotation_speed.theta * elapsed_seconds + self.delta_rotation.theta;
```
pdubroy: Do you really want to combine `rotation_speed` and `delta_rotation`? Could both be non-zero if there is both joystick and mouse input? If so, consider giving `delta_rotation`priority, since mouse dragging implies direct manipulation and it seems like it would be weird if the joystick input contributed additional rotation.

SirVer: While I agree with your reasoning, that use case will be so rare that it seems unnecessary to hedge against. I'd prefer to check it in as is.

pdubroy: Does it require more than an `if` statement? You're right that it's probably not necessary to handle it, but IMO it would make the purpose of these fields easier to understand.

SirVer: No, it does not. Done now. Thanks for keeping me honest.

SirVer:
@pdubroy PTAL

SirVer:
@wally-the-cartographer merge

wally-the-cartographer:

Issue #209, opened 2019-02-18
NodeNotFound Octree Crashes


tpwrules:
I am trying to build an octree from a point cloud with the following header:
```
ply
format binary_little_endian 1.0
element vertex 186546528
property float x
property float y
property float z
property uchar red
property uchar green
property uchar blue
end_header
```

Using the latest build from master as of the time of this ticket submission, the program crunches for a minute or so then dies because it tries to unwrap a NodeNotFound error. The full traceback is here: https://pastebin.com/mWXkRpWh

SirVer:
@tpwrules Can you make the data available?

tpwrules:
Unfortunately, the data is a) several gigabytes and b) not to be made available according to my superiors. But I have done some research which might aid in reproducing the issue with another dataset.

This block seems to be the issue: https://github.com/googlecartographer/point_cloud_viewer/blob/a88f232bbbee12bb343432e61afe64e91b0073fb/src/generation.rs#L170-L179

If the node `child_id` does not exist, the code will notice that the return from `octree::NodeIterator::from_data_provider` is a `NodeNotFound` error and skip processing that node. But the call to `octree_data_provider.number_of_points(&child_id)` will also return `NodeNotFound` in that case and unwrapping that with `?` will crash the program.

I stopped the crashing by separately matching the number_of_points result:
```
let number_of_points = match octree_data_provider.number_of_points(&child_id) {
    Ok(number_of_points) => number_of_points,
    Err(Error(ErrorKind::NodeNotFound, _)) => continue,
    Err(err) => return Err(err),       
};
let node_iterator = match octree::NodeIterator::from_data_provider(
    octree_data_provider,
    octree_meta,
    &child_id,
    number_of_points,
) {
    Ok(node_iterator) => node_iterator,
    Err(Error(ErrorKind::NodeNotFound, _)) => continue,
    Err(err) => return Err(err),
};
```

I'm not sure whether this fix is complete, but it seems to make everything work as I would expect and I get a nice view in my web browser when I spin up the server. There is another function that calls `number_of_points` which may have the same problem. I wonder if it's because the system deletes nodes with 0 points:
https://github.com/googlecartographer/point_cloud_viewer/blob/a88f232bbbee12bb343432e61afe64e91b0073fb/src/octree/node.rs#L477-L487

If necessary, I can try to create a dataset that's problematic, but I hope this is enough information to solve the problem.

SirVer:
[...]

PR #215, opened 2019-02-20
Add simple octree generation regression test.

feuerste:
This is a simple regression test, which would have caught #209. Hopefully this catches more refactoring errors, too.

### DIFF
--- a/src/generation.rs
+++ b/src/generation.rs
@@ -453,3 +453,48 @@ pub fn build_octree(
         BufWriter::new(File::create(&output_directory.as_ref().join("meta.pb")).unwrap());
     meta.write_to_writer(&mut buf_writer).unwrap();
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use crate::color::Color;
+    use cgmath::Vector3;
+    use tempdir::TempDir;
+
+    struct Points {
+        points: Vec<Point>,
+    }
+
+    impl InternalIterator for Points {
+        fn for_each<F: FnMut(&Point)>(self, mut func: F) {
+            for p in &self.points {
+                func(p);
+            }
+        }
+        fn size_hint(&self) -> Option<usize> {
+            Some(self.points.len())
+        }
+    }
+    #[test]
+    fn test_generation() {
+        let default_point = Point {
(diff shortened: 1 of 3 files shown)


feuerste:
@SirVer FYI

src/generation.rs
```diff
@@ -453,3 +453,48 @@ pub fn build_octree(
...
+        };
+        let mut points = vec![default_point; 100_001];
+        points[100_000].position = Vector3::new(2.0, 0.0, 0.0);
+        let mut bounding_box = Aabb3::zero();
+        for point in &points {
+            bounding_box = bounding_box.grow(Point3::from_vec(point.position));
+        }
+        let pool = scoped_pool::Pool::new(10);
+        let tmp_dir = TempDir::new("octree").unwrap();
+        build_octree(&pool, tmp_dir, 1.0, bounding_box, Points { points });
```
tblaette: would be good to delete the temp dir when the test has finished

src/generation.rs
```diff
@@ -453,3 +453,48 @@ pub fn build_octree(
...
+            for p in &self.points {
+                func(p);
+            }
+        }
+        fn size_hint(&self) -> Option<usize> {
+            Some(self.points.len())
+        }
+    }
+    #[test]
+    fn test_generation() {
```
tblaette: What does this test actually test? It seems like it fails only if `build_octree()` panics. Is that the intended test?

src/generation.rs
feuerste: That's exactly what `TempDir` from the [`tempdir`](https://docs.rs/tempdir/0.3.7/tempdir/) crate is about :)

src/generation.rs
feuerste: Yes, I've overseen this twice already during refactoring, so it was definitely time for a test.

feuerste:
Thanks for your reviews, Tom and Holger!

feuerste:
@wally-the-cartographer merge

wally-the-cartographer:

src/generation.rs
tblaette: Ha! Cool, didn't know this! :-)

PR #216, opened 2019-02-21
Parallelize the octree server

nnmm:
This is an effort to parallelize the octree server.
* I printed the thread id in stream_points_back_to_sink() while making several parallel requests. Interestingly, it was always the same (grpc_cq_poll_0 or something similar), but when I set cq_count to > 1 and requests_slot_per_cq to 1, more threads appeared and performance improved.
* For the case of the full point cloud query, I implemented a helper method to only fetch a certain fraction of the octree. It can then be called for all of these fractions in parallel.
* I did a grid search for 16 parallel threads and found the optimum speedup in the full benchmark to be at values of 8-10 for cq_count and 2-3 for the requests_slot_per_cq parameter.
* The optimum speedup for the full benchmark is still only around 4x for 16 threads, but much higher for the grpc-less benchmark (server_benchmark).
* I appreciate discussion on how to best use grpc for best parallel performance. For instance, we're using the RpcContext's spawn method (https://docs.rs/grpcio/0.4.1/grpcio/struct.RpcContext.html) – is there heavy work in this future? Maybe someone experienced could offer comments on how this part of the code and gRPC work exactly.

### DIFF
--- a/point_viewer_grpc/src/bin/octree_benchmark.rs
+++ b/point_viewer_grpc/src/bin/octree_benchmark.rs
@@ -11,14 +11,7 @@
 // WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 // See the License for the specific language governing permissions and
 // limitations under the License.
-extern crate clap;
-extern crate futures;
-extern crate grpcio;
-extern crate point_viewer;
-extern crate point_viewer_grpc;
-extern crate point_viewer_grpc_proto_rust;
-
-use std::path::PathBuf;
+use std::path::{Path, PathBuf};
 use std::str::FromStr;
 use std::sync::Arc;
 use std::thread;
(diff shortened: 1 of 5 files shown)


nnmm:
ptal @SirVer @feuerste @catevita

SirVer:
I'd like to understand the motivation behind this work, because I have a hunch that this is the wrong approach.

If you are trying to improve the use case of having the gRPC server locally give you more points more quickly, the answer is to remove the gRPC server :) (see @feuerste 
 work). If you try to have the gRPC server running remote give you more points more quickly, the bottleneck will likely be connection throughput and this will not buy you much.

nnmm:
Removing the gRPC server would be great for my main use case, but wouldn't we also want to support e.g. multiple sdl_viewers making requests to a centralized instance of octree_server in the cloud?

There is also a new GetAllPointsParallelRequest in there that is useful even with a use case without server, just for reading an entire points from different threads. If you prefer, I can close this PR and extract this part into a new one.

SirVer:
> e.g. multiple sdl_viewers making requests to a centralized instance of octree_server in the cloud?

yes, but how much of your measurements would translate well into this use case? (not a rhetorical question).

> There is also a new GetAllPointsParallelRequest in there that is useful even with a use case without server, just for reading an entire points from different threads. If you prefer, I can close this PR and extract this part into a new one.

defering to @feuerste

PR #224, opened 2019-03-01
GRPC server with octree ids and octree factory.

feuerste:


feuerste:
@nnmm @krzesi-mir 
Can you please review/have a look?

point_viewer_grpc/src/service.rs
```diff
@@ -245,19 +255,52 @@ impl proto_grpc::Octree for OctreeService {
...
+pub fn start_grpc_server(
+    host: &str,
+    port: u16,
+    location: impl Into<PathBuf>,
+    factory: OctreeFactory,
+) -> Server {
     let env = Arc::new(Environment::new(1));
-    let meta = read_meta_proto(octree_directory).unwrap();
-    let octree = Arc::from(octree_from_directory(octree_directory).unwrap());
+    let meta_cache = Arc::new(Mutex::new(LruCache::new(10)));
```
catevita: could it be useful to set the cache size with an optional command line argument (taking a default parameter)? see for example sdl_viewer.

feuerste: I changed the lru cache to a hash map, since we don't have too many octrees at the moment, so setting the size currently is not relevant any more...

feuerste:
ptal @nnmm @krzesi-mir

point_viewer_grpc/src/service.rs
```diff
@@ -213,7 +223,7 @@ impl proto_grpc::Octree for OctreeService {
         let frustum_matrix = projection_matrix.concat(&view_transform.into());
         stream_points_back_to_sink(
             OctreeQuery::Frustum(frustum_matrix),
-            self.octree.clone(),
+            Arc::clone(&self.get_service_data(&req.octree_id)),
```
krzesi-mir: Do we need to `Arc::clone` here?

point_viewer_grpc/src/service.rs
```diff
@@ -236,7 +246,7 @@ impl proto_grpc::Octree for OctreeService {
         };
         stream_points_back_to_sink(
             OctreeQuery::Box(bounding_box),
-            self.octree.clone(),
+            Arc::clone(&self.get_service_data(&req.octree_id)),
```
krzesi-mir: See: https://github.com/googlecartographer/point_cloud_viewer/pull/224/files#r262019098

point_viewer_grpc/src/service.rs
```diff
@@ -245,19 +255,53 @@ impl proto_grpc::Octree for OctreeService {
...
         &self,
         ctx: RpcContext,
-        _req: proto::GetAllPointsRequest,
+        req: proto::GetAllPointsRequest,
         resp: ServerStreamingSink<proto::PointsReply>,
     ) {
-        stream_points_back_to_sink(OctreeQuery::FullPointcloud, self.octree.clone(), &ctx, resp)
+        stream_points_back_to_sink(
+            OctreeQuery::FullPointcloud,
+            Arc::clone(&self.get_service_data(&req.octree_id)),
```
krzesi-mir: See: https://github.com/googlecartographer/point_cloud_viewer/pull/224/files#r262019098

point_viewer_grpc/src/service.rs
```diff
@@ -245,19 +255,53 @@ impl proto_grpc::Octree for OctreeService {
...
 
-pub fn start_grpc_server(octree_directory: &Path, host: &str, port: u16) -> Server {
+impl OctreeService {
+    pub fn get_service_data(&self, octree_id: &str) -> Arc<OctreeServiceData> {
+        match self.data_cache.read().unwrap().get(octree_id) {
+            None => {
+                let octree = self
+                    .factory
+                    .generate_octree(self.location.join(&octree_id).to_string_lossy())
+                    .unwrap();
```
krzesi-mir: I'm a little averse to `unwraps` ;) Do you think we could at least replace these unwraps by `expect` and provide some meaningful message?

point_viewer_grpc/src/service.rs
feuerste: Good catch, this was some leftover...
Done!

point_viewer_grpc/src/service.rs
feuerste: Done

point_viewer_grpc/src/service.rs
feuerste: Done

krzesi-mir:
LGTM

point_viewer_grpc/src/service.rs
feuerste: I implemented proper error handling... see new code.

point_viewer_grpc/src/service.rs
```diff
@@ -48,6 +50,26 @@ enum OctreeQuery {
...
+    let f = sink
+        .fail(RpcStatus::new(
+            RpcStatusCode::Internal,
+            Some(error.to_string()),
+        ))
+        .map_err(move |err| eprintln!("Failed to reply: {:?}", err));
+    ctx.spawn(f);
+}
+
+fn send_fail<T>(ctx: &RpcContext, sink: UnarySink<T>, error: &Error) {
```
krzesi-mir: These two functions have exactly the same bodies - is it possible to have just one, a little more generic, function here?

feuerste: I tried and thought about it, but the two sinks [ServerStreamingSink](https://docs.rs/grpcio/0.4.4/grpcio/struct.ServerStreamingSink.html) and [UnarySink](https://docs.rs/grpcio/0.4.4/grpcio/struct.UnarySink.html) have no common trait. They have different return types for the `fail` method, so I couldn't find a way to simplify it... Do you know one?

krzesi-mir: I can't think of any way to simplify it...

point_viewer_grpc/src/service.rs
```diff
@@ -167,16 +95,23 @@ impl proto_grpc::Octree for OctreeService {
...
         sink: UnarySink<proto::GetNodeDataReply>,
     ) {
-        let data = self
+        let service_data = match self.get_service_data(&req.octree_id) {
+            Ok(service_data) => service_data,
+            Err(e) => return send_fail(&ctx, sink, &e),
+        };
+        let node_data = match service_data
             .octree
             .get_node_data(&NodeId::from_str(&req.id).unwrap())
```
krzesi-mir: How about this `unwrap`? :D

krzesi-mir:
LGTM
This is probably a suggestion for another PR but I would get rid of some more `unwrap`'s here and there. Especially when parsing arguments (see `get_points_in_frustum` and `get_points_in_box` functions in `service.rs`). I don't think the grpc server should panic upon receiving a corrupted message.

point_viewer_grpc/src/service.rs
feuerste: Ok, just handled this one, too. However, I would like to leave the remaining ones, since they are all pretty safe and replacing them would only clutter the source.

feuerste:
> This is probably a suggestion for another PR but I would get rid of some more `unwrap`'s here and there. Especially when parsing arguments (see `get_points_in_frustum` and `get_points_in_box` functions in `service.rs`). I don't think the grpc server should panic upon receiving a corrupted message.

If handling at all, this should definitely go into a separate PR, since they were all here before and this PR is actually not about making the code safer ;)

feuerste:
@wally-the-cartographer merge

feuerste:
@wally-the-cartographer merge

wally-the-cartographer:

PR #228, opened 2019-03-08
Adds a trait to extend the SDL viewer with custom draw elements

SirVer:
This can be used to implement things like the vertex renderer that we use in house.


feuerste:
@wally-the-cartographer merge

wally-the-cartographer:

PR #232, opened 2019-03-14
Increase octree precision to double to support octrees covering the world.

feuerste:
In order to handle octrees covering the whole earth, another 64 bit encoding needs to be added and bounding boxes need to be in double precision in order to have enough precision when setting the resolution to e.g. 1 mm.


feuerste:
@nnmm @SirVer 
FYI

xray_proto_rust/src/proto.proto
```diff
@@ -23,9 +23,14 @@ message Vector2f {
...
 
+message Vector2d {
+  double x = 1;
+  double y = 2;
+}
+
 message Rect {
-  Vector2f min = 1;
-  float edge_length = 2;
+  Vector2d min = 3;
```
catevita: Hi, I am not so familiar with protos yet, why does this proto declaration begin with 3?

feuerste: Whenever a new field is added, the number it carries will be the last used one plus one.

catevita: thanks for the explanation!

sdl_viewer/src/node_drawer.rs
```diff
@@ -110,13 +110,22 @@ impl NodeDrawer {
 
             self.program.gl.Uniform1f(
```
SirVer: Could you add a TODO about the bug we just discussed?

src/octree/node.rs
```diff
@@ -291,15 +291,17 @@ pub enum PositionEncoding {
...
 
 impl PositionEncoding {
     pub fn new(bounding_cube: &Cube, resolution: f64) -> PositionEncoding {
-        let min_bits = (f64::from(bounding_cube.edge_length()) / resolution).log2() as u32 + 1;
+        let min_bits = (bounding_cube.edge_length() / resolution).log2() as u32 + 1;
         match min_bits {
             0...8 => PositionEncoding::Uint8,
             9...16 => PositionEncoding::Uint16,
-            _ => PositionEncoding::Float32,
+            17...24 => PositionEncoding::Float32, // 24 because of the float mantissa
```
SirVer: Please add the diameter of the node for the 1. mm resolution here as a comment. This would have given me more context.

src/bin/upgrade_octree.rs
```diff
@@ -42,6 +42,30 @@ fn upgrade_version9(directory: &Path, mut meta: proto::Meta) {
     meta.write_to_writer(&mut buf_writer).unwrap();
```
tblaette: Should you call `upgrade_version10()` here to go directly from 9 -> 11 instead of upgrading to the outdated version 10?

feuerste: The loop below is taking care of this.

src/octree/node.rs
feuerste: Done.

sdl_viewer/src/node_drawer.rs
feuerste: Done.

src/bin/upgrade_octree.rs
tblaette: ah, I missed that 👍

feuerste:
Thanks for all your feedback!

feuerste:
@wally-the-cartographer merge

feuerste:
@wally-the-cartographer merge

PR #233, opened 2019-03-14
Web Browser: new API implementation

catevita:
This implementation of backend and client allows to switch and serve different pointclouds stored in octrees.
The program syntax is also backward compatible to the previous browser implementation


catevita:
fyi @SirVer 
ptal! @feuerste  @negin-z

feuerste:
How about replacing all occurrences of `uuid` with `octree_id`, so it is more generic and consistent with the grpc version?

octree_web_viewer/src/bin/points_web_viewer.rs
```diff
@@ -12,44 +12,99 @@
...
 
 /// HTTP web viewer for 3d points stored in OnDiskOctrees
 #[derive(StructOpt, Debug)]
 #[structopt(name = "points_web_viewer", about = "Visualizing points")]
-struct CommandlineArguments {
+pub struct CommandLineArguments {
     /// The octree directory to serve, including a trailing slash.
-    octree_path: String,
+    /// this overrides the path_* options
+    #[structopt(name = "DIR")]
```
feuerste: I find the current way to specify either a full path or prefix/id/suffix pretty confusing. How about using mutually exclusive arguments (see e.g. https://github.com/TeXitoi/structopt/issues/23#issuecomment-326974584)?

octree_web_viewer/src/bin/points_web_viewer.rs
```diff
@@ -12,44 +12,99 @@
...
+    octree_path: Option<String>,
     /// Port to listen on.
     #[structopt(default_value = "5433", long = "--port")]
     port: u16,
     /// IP string.
     #[structopt(default_value = "127.0.0.1", long = "--ip")]
     ip: String,
+    /// instead of DIR: specify path prefix for octree dir
+    #[structopt(long = "--prefix")]
```
feuerste: Nit: You can skip the `--` everywhere in `long`

octree_web_viewer/src/bin/points_web_viewer.rs
```diff
@@ -12,44 +12,99 @@
...
+    path_prefix: Option<String>,
+    /// Optional: suffix for subfolder of octree dir
+    #[structopt(default_value = "/", long = "--suffix")]
```
feuerste: Should the default be an empty string?

octree_web_viewer/src/bin/points_web_viewer.rs
```diff
@@ -12,44 +12,99 @@
...
+    path_suffix: String,
+    /// instead of DIR: specify path folder for octree dir
+    #[structopt(long = "--uuid")]
+    octree_id: Option<String>,
+    /// Cache items
+    #[structopt(default_value = "3", long = "--cache_items")]
```
feuerste: Nit: I would increase the default value to something like 20.

octree_web_viewer/src/state.rs
```diff
@@ -0,0 +1,101 @@
...
+
+impl OctreeKeyParams {
+    pub fn get_octree_address(&self, octree_key: &String) -> Result<String, PointsViewerError> {
+        let mut join_prefix = "";
+        let mut join_suffix = "";
+
+        if !self.prefix.ends_with("/") {
+            join_prefix = "/";
+        }
+        if !self.suffix.starts_with("/") {
```
feuerste: What if the suffix is empty?

octree_web_viewer/src/state.rs
```diff
@@ -0,0 +1,101 @@
...
+            join_suffix = "/";
+        }
+
+        Ok(format!(
+            "{}{}{}{}{}",
+            self.prefix, join_prefix, octree_key, join_suffix, self.suffix
```
feuerste: Would it make sense to use `PathBuf`/`Path` for the components and `join` them?

octree_web_viewer/src/state.rs
```diff
@@ -0,0 +1,101 @@
...
+    }
+
+    pub fn load_octree(
+        &self,
+        uuid: impl AsRef<str>,
+    ) -> Result<Arc<octree::Octree>, PointsViewerError> {
+        // exists
+        let octree_id = uuid.as_ref();
+        // taking care of initial octree
+        if octree_id.len() == 9 && octree_id.starts_with("init_uuid") {
```
feuerste: As mentioned above, can we pass the real id instead and get rid of this code here?

octree_web_viewer/src/state.rs
```diff
@@ -0,0 +1,101 @@
...
+        let octree_key: String = octree_id.to_string();
+        self.insert_octree(octree_key)
+    }
+
+    fn insert_octree(
+        &self,
+        uuid: impl Into<String>,
+    ) -> Result<Arc<octree::Octree>, PointsViewerError> {
+        let octree_key = uuid.into();
+        let addr = &self.key_params.get_octree_address(&octree_key)?.clone();
```
feuerste: I think the clone can be removed.

octree_web_viewer/src/backend.rs
```diff
@@ -75,106 +74,103 @@ fn pad(input: &mut Vec<u8>) {
     }
 }
 
-pub struct NodesData {
-    octree: Arc<Octree>,
+fn get_octree_from_state(
+    octree_id: String,
```
feuerste: Can we pass a `&str` or `AsRef<>` instead?

octree_web_viewer/src/bin/points_web_viewer.rs
catevita: Done

octree_web_viewer/client/main.ts
```diff
@@ -48,20 +58,19 @@ class App {
...
 
+    this.reset_scene_viewer();
+
     this.lastFrustumUpdateTime = 0;
     this.lastMoveTime = 0;
     this.needsRender = true;
-    this.viewer = new OctreeViewer(this.scene, () => {
-      this.needsRender = true;
-    });
+    this.uuid = "init_uuid";
```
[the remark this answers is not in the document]
catevita: it makes totally sense, I'd like to make this change a separate pr

octree_web_viewer/src/backend.rs
```diff
@@ -75,106 +74,103 @@ fn pad(input: &mut Vec<u8>) {
     }
 }
 
-pub struct NodesData {
-    octree: Arc<Octree>,
+fn get_octree_from_state(
+    uuid: String,
```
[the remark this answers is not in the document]
catevita: done!

octree_web_viewer/src/bin/points_web_viewer.rs
catevita: done

octree_web_viewer/src/bin/points_web_viewer.rs
catevita: done

octree_web_viewer/src/state.rs
catevita: outdated, using PathBuf s. comment on the next lines

octree_web_viewer/src/state.rs
catevita: indeed, done!

octree_web_viewer/src/state.rs
catevita: I'd like to make this change in a new pr when possible

octree_web_viewer/src/backend.rs
catevita: done

octree_web_viewer/client/main.ts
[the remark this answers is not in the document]
catevita: edit: I'm changing it in this pr because it is rather confusing for the target user @negin-z thank you for your offline contributions!

octree_web_viewer/client/main.ts
```diff
@@ -29,6 +29,16 @@ class App {
   private lastMoveTime: number;
   private needsRender: boolean;
 
+  public octree_id: string;  // octree identifier
+
+  private reset_scene_viewer() {
```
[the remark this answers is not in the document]
negin-z: Also you can add `this.lastFrustumUpdateTime = 0; ` to trigger animating the scene.

octree_web_viewer/client/octree_viewer.ts
```diff
@@ -274,13 +277,19 @@ export class OctreeViewer {
 
     this.nodeLoader = new NodeLoader();
     this.currentlyLoading = 0;
+    this.octree_id = 'init_id';
+  }
+
+  public load_new_tree(octree_id: string) {
```
negin-z: I think we shall leave setting the octree_id only at constructor part and not exposing a public function for setting it.

octree_web_viewer/client/main.ts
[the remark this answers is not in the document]
negin-z: > I have two suggestions to resetting the scene:
> 1 - Dispose the obsolete scene and belonging elements which would get stale.
> 2 - Inform the obsolete viewer to stop querying for NodeData on its list. (@SirVer can we postpone this to another PR?)

Maybe it is good for now to leave //TODOs for the clean-up plans?

octree_web_viewer/src/bin/points_web_viewer.rs
```diff
@@ -12,44 +12,100 @@
...
+    octree_id: Option<String>,
+    /// Cache items
+    #[structopt(default_value = "20", long = "cache_items")]
+    cache_max: usize,
+}
+
+/// init app state with command arguments
+/// backward compatibilty is ensured
+pub fn state_from(args: CommandLineArguments) -> Result<AppState, PointsViewerError> {
+    let mut has_suffix = false;
```
feuerste: How about removing `has_suffix` and later just checking for `!suffix.to_string_lossy().is_empty()`?

octree_web_viewer/src/bin/points_web_viewer.rs
```diff
@@ -12,44 +12,100 @@
...
+    let suffix = match args.path_suffix {
```
feuerste: What do you think about replacing this by `let suffix = args.path_suffix.unwrap_or_default();`?

octree_web_viewer/src/bin/points_web_viewer.rs
```diff
@@ -12,44 +12,100 @@
...
     /// Port to listen on.
-    #[structopt(default_value = "5433", long = "--port")]
+    #[structopt(default_value = "5433", long = "port")]
     port: u16,
     /// IP string.
-    #[structopt(default_value = "127.0.0.1", long = "--ip")]
+    #[structopt(default_value = "127.0.0.1", long = "ip")]
     ip: String,
+    /// instead of DIR: specify path prefix for octree dir
+    #[structopt(long = "prefix", parse(from_os_str), conflicts_with = "octree_path")]
```
feuerste: Would it help to group `prefix` and `octree_id` together, see https://github.com/TeXitoi/structopt/blob/master/examples/group.rs for an example?

octree_web_viewer/src/bin/points_web_viewer.rs
```diff
@@ -12,44 +12,100 @@
...
+        Some(suffix) => {
+            has_suffix = true;
+            suffix
+        }
+        None => PathBuf::from(""),
+    };
+
+    let app_state = match args.octree_path {
+        Some(octree_directory) => {
+            let mut prefix = octree_directory.parent().unwrap();
```
feuerste: It will be rare, but if the octree is located at `/`, this will fail...

octree_web_viewer/src/bin/points_web_viewer.rs
```diff
@@ -12,44 +12,100 @@
...
+            if has_suffix && octree_directory.ends_with(&suffix) {
+                prefix = prefix.parent().unwrap();
```
feuerste: This will fail, if the path only contains one or no subfolder

octree_web_viewer/src/bin/points_web_viewer.rs
```diff
@@ -12,44 +12,100 @@
...
+            }
+            let octree_id = octree_directory.strip_prefix(&prefix)?;
+            AppState::new(args.cache_max, prefix, suffix, octree_id.to_str().unwrap())
+        }
+        None => {
+            let prefix = args
+                .path_prefix
+                .ok_or_else(|| {
+                    PointsViewerError::NotFound(
+                        "Input argoment Syntax is incorrect: check prefix".to_string(),
```
feuerste: Nit: `argoment` -> `argument`

octree_web_viewer/src/bin/points_web_viewer.rs
```diff
@@ -12,44 +12,100 @@
...
+                    )
+                })
+                .unwrap();
+            let octree_id = args
+                .octree_id
+                .ok_or_else(|| {
+                    PointsViewerError::NotFound(
+                        "Input argoment Syntax is incorrect: check octree_id".to_string(),
```
feuerste: Nit: `argoment` -> `argument`

octree_web_viewer/src/state.rs
```diff
@@ -0,0 +1,95 @@
+use crate::backend_error::PointsViewerError;
+use point_viewer::octree;
+use std::collections::HashMap;
+use std::path::{Path, PathBuf};
+use std::sync::{Arc, RwLock};
+
+/// path information for the octrees
+#[derive(Clone)]
+pub struct OctreeKeyParams {
+    /// Location prefix, including
```
feuerste: Nit: remove `, including`, or clarify comment.

octree_web_viewer/src/state.rs
```diff
@@ -0,0 +1,95 @@
...
+    prefix: PathBuf,
+    /// Tree ID
```
feuerste: This is not the `Tree ID`, right?

octree_web_viewer/src/state.rs
```diff
@@ -0,0 +1,95 @@
...
+    suffix: PathBuf,
+}
+
+impl OctreeKeyParams {
+    pub fn get_octree_address(
+        &self,
+        octree_key: impl AsRef<Path>,
+    ) -> Result<PathBuf, PointsViewerError> {
```
feuerste: Do we really need a `Result` with an error here? I guess returning `PathBuf` would be enough, as the method internally doesn't give any error...

octree_web_viewer/client/main.ts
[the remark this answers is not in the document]
catevita: Hi Negin, thanks for the review. as told offline, let's pair tomorrow together on the scene disposal, a draft is already in the code. The obsolete viewer mechanism could be done in the next pr.

octree_web_viewer/client/octree_viewer.ts
catevita: done!

octree_web_viewer/src/state.rs
catevita: done!

octree_web_viewer/src/state.rs
catevita: done!

octree_web_viewer/src/state.rs
catevita: ..something went wrong here... fixed!

octree_web_viewer/src/state.rs
catevita: ..something went wrong here... fixed!

octree_web_viewer/src/bin/points_web_viewer.rs
catevita: done, thanks!

octree_web_viewer/src/bin/points_web_viewer.rs
catevita: done, thanks!

octree_web_viewer/src/bin/points_web_viewer.rs
catevita: Thank you to point this out! i am addressing it in the next commit/push

octree_web_viewer/src/bin/points_web_viewer.rs
catevita: done

octree_web_viewer/src/bin/points_web_viewer.rs
catevita: done

octree_web_viewer/src/bin/points_web_viewer.rs
catevita: i ended doing something similar

catevita:
update: a lot of work on the client was done offline with/from negin. the server part should have all the requested changes. frontend has some todos, that can be formulated in tickets and handled from some more experienced frontend programmer.

feuerste:
`.DS_Store` shouldn't be checked in...

octree_web_viewer/src/state.rs
```diff
@@ -0,0 +1,92 @@
...
+        addr.push(self.prefix.clone());
+        addr.push(octree_key.as_ref());
+        addr.push(self.suffix.clone());
+        addr
+    }
+}
+
+#[derive(Clone)]
+pub struct AppState {
+    /// LRU Cache for Octrees
```
feuerste: Nit: Not `LRU` any more...

octree_web_viewer/Cargo.toml
```diff
@@ -28,7 +28,10 @@ crossbeam-utils = "0.6.5"
 failure = "0.1.5"
 futures = "0.1.25"
 json = "0.11.3"
-structopt = "0.2.14"
+lru = "0.1.11"
```
feuerste: I think we don't use `lru` any more...

octree_web_viewer/client/main.ts
```diff
@@ -20,124 +20,203 @@ import { FirstPersonController } from './control';
...
+            `/init_tree`,
+            {
+                method: 'GET',
+                credentials: 'same-origin',
+            }
+        );
+
+        const result = window
+            .fetch(request)
+            .then((response) => response.text()) // todo error handling?
```
feuerste: Can you please reformat to our default TODO style: `TODO(cvitadello): Error handling`?

octree_web_viewer/client/main.ts
```diff
@@ -20,124 +20,203 @@ import { FirstPersonController } from './control';
...
-    if (this.controller.update()) {
-      this.lastMoveTime = time;
-      this.viewer.setMoving(true);
-      this.needsRender = true;
+        this.lastFrustumUpdateTime = 0;
+        this.lastMoveTime = 0;
+    }
+
+    private cleanup() {
+        // TODO block requests from the viewer that is going to be replaced
```
feuerste: Please reformat to `TODO(cvitadello): ...`

octree_web_viewer/client/main.ts
```diff
@@ -20,124 +20,203 @@ import { FirstPersonController } from './control';
...
+        this.octreeId = "loading...";
+        this.gui = new GUI();
+
+        this.octreeIdControl =
+            this.gui
+                .add(this, 'octreeId')
+                .name('Point Cloud ID')
+                .onFinishChange(this.run);
+
+        //TODO error handling
```
feuerste: Please reformat to `TODO(cvitadello): ...`

octree_web_viewer/src/backend.rs
```diff
@@ -75,106 +74,103 @@ fn pad(input: &mut Vec<u8>) {
...
+            // string which cannot be used as a buffer. So we would need to manually convert this into
+            // an Array with is very slow.
+            // The alternative is to binary encode the whole request and parse it on the client side,
+            // which requires careful constructing on the server and parsing on the client.
+            let mut reply_blob = Vec::<u8>::new();
+
+            let mut num_nodes_fetched = 0;
+            let mut num_points = 0;
+            let octree: Arc<octree::Octree> =
+                get_octree_from_state(&octree_id.into_inner(), &state).unwrap();
```
feuerste: Can we remove `into_inner()`?

octree_web_viewer/src/bin/points_web_viewer.rs
```diff
@@ -12,44 +12,121 @@
...
     port: u16,
     /// IP string.
-    #[structopt(default_value = "127.0.0.1", long = "--ip")]
+    #[structopt(default_value = "127.0.0.1", long = "ip")]
     ip: String,
+    /// instead of DIR: specify path prefix for octree dir
+    #[structopt(long = "prefix", parse(from_os_str), group = "path_stubs")]
+    path_prefix: Option<PathBuf>,
+    /// Optional: suffix for subfolder of octree dir
+    #[structopt(long = "suffix", parse(from_os_str))]
```
feuerste: Do we need to add `group = "path_stubs"` here as well?

octree_web_viewer/src/bin/points_web_viewer.rs
```diff
@@ -12,44 +12,121 @@
...
+    octree_id: Option<String>,
+    /// Cache items
+    #[structopt(default_value = "20", long = "cache_items")]
+    cache_max: usize,
+}
+
+/// init app state with command arguments
+/// backward compatibilty is ensured
+pub fn state_from(args: CommandLineArguments) -> Result<AppState, PointsViewerError> {
+    let suffix = args.path_suffix.unwrap_or_else(|| PathBuf::from(""));
```
feuerste: Would `unwrap_or_default()` do the same thing?

octree_web_viewer/src/bin/points_web_viewer.rs
```diff
@@ -12,44 +12,121 @@
...
+
+    let app_state = match args.octree_path {
+        Some(octree_directory) => {
+            let prefix_opt = octree_directory.parent();
+            if suffix.to_string_lossy().is_empty() {
+                if let Some(prefix) = prefix_opt {
+                    let octree_id = octree_directory.strip_prefix(&prefix)?;
+                    AppState::new(args.cache_max, prefix, suffix, octree_id.to_str().unwrap())
+                } else {
+                    AppState::new(args.cache_max, PathBuf::new(), suffix, "/")
```
feuerste: This gives a weird state. Should it be `AppState::new(args.cache_max, PathBuf::new(), suffix, octree_directory)`?

octree_web_viewer/src/bin/points_web_viewer.rs
```diff
@@ -12,44 +12,121 @@
...
+    #[structopt(long = "octree_id", group = "path_stubs")]
+    octree_id: Option<String>,
+    /// Cache items
+    #[structopt(default_value = "20", long = "cache_items")]
+    cache_max: usize,
+}
+
+/// init app state with command arguments
+/// backward compatibilty is ensured
+pub fn state_from(args: CommandLineArguments) -> Result<AppState, PointsViewerError> {
```
feuerste: The whole logic of having either one directory or possibly three optional parts is pretty complex... would it make sense to treat octree id and suffix as one entity, i.e. octree id (similar to the grpc octree data provider)?

octree_web_viewer/src/bin/points_web_viewer.rs
```diff
@@ -12,44 +12,121 @@
...
+            if suffix.to_string_lossy().is_empty() {
+                if let Some(prefix) = prefix_opt {
+                    let octree_id = octree_directory.strip_prefix(&prefix)?;
+                    AppState::new(args.cache_max, prefix, suffix, octree_id.to_str().unwrap())
+                } else {
+                    AppState::new(args.cache_max, PathBuf::new(), suffix, "/")
+                }
+            } else {
+                let mut components = octree_directory.components();
+                let mut prefix: PathBuf = PathBuf::new();
```
feuerste: Could the `prefix` also come from a command line argument?

octree_web_viewer/src/bin/points_web_viewer.rs
```diff
@@ -12,44 +12,121 @@
...
+                while !suffix.as_path().eq(components.as_path()) {
+                    prefix.push(tmp_octree_id.unwrap());
+                    tmp_octree_id = components.next();
+                }
+
+                AppState::new(
+                    args.cache_max,
+                    prefix,
+                    suffix,
+                    (tmp_octree_id.unwrap().as_ref() as &Path).to_string_lossy(),
```
feuerste: Can we replace this with `tmp_octree_id.unwrap().to_string_lossy()`?

octree_web_viewer/src/bin/points_web_viewer.rs
```diff
@@ -12,44 +12,121 @@
...
-        let octree = match octree::octree_from_directory(octree_directory) {
-            Ok(octree) => octree,
-            Err(err) => panic!("Could not load octree: {}", err),
-        };
-        Arc::from(octree)
-    };
     let sys = actix::System::new("octree-server");
 
-    let _ = utils::start_octree_server(octree, &ip_port);
+    //let _ = start_octree_server(app_state, &ip_port, octree_id);
```
feuerste: Can we remove this comment?

octree_web_viewer/src/bin/points_web_viewer.rs
```diff
@@ -12,44 +12,121 @@
...
+    let args = CommandLineArguments::from_args();
 
     let ip_port = format!("{}:{}", args.ip, args.port);
-    let octree_directory = PathBuf::from(&args.octree_path);
+
+    // initialize app state
+    let app_state: Arc<AppState> = Arc::new(state_from(args).unwrap());
     // The actix-web framework handles requests asynchronously using actors. If we need multi-threaded
     // write access to the Octree, instead of using an RwLock we should use the actor system.
+    //put octree arc in cache
```
feuerste: Nit: Insert space between `//` and text

octree_web_viewer/src/state.rs
```diff
@@ -0,0 +1,92 @@
...
+    pub octree_map: Arc<RwLock<HashMap<String, Arc<octree::Octree>>>>,
+    /// information for retieving octree path
+    pub key_params: OctreeKeyParams,
+    /// backward compatibility to input arguments
+    pub init_octree_id: String,
```
feuerste: How about getting rid of this variable and initially creating an octree inside `octree_map` with this id? Then `return_init_id` could look it up inside the hash map (for 0 or more than 1 entries it could give an error)

octree_web_viewer/client/octree_viewer.ts
```diff
@@ -47,329 +47,332 @@ void main() {
...
+
+    private loadedData: { [key: string]: NodeData } = {};
+    private nodeLoader: NodeLoader;
+    private batches: NodeData[][] = [];
+    private currentlyLoading: number;
+    private useTransparency: boolean;
+
+
+    constructor(private scene: THREE.Scene, private onNewNodeData: () => void, octreeId: string) {
+        this.material = new THREE.ShaderMaterial({
```
negin-z: Nit: you could write `private octreeId: string` in the constructor and get rid of the line `private octreeId: string;` above and the line `this.octreeId = octreeId;` below.

catevita: done

octree_web_viewer/client/main.ts
negin-z: Does this imply the she would be the future assignee?

feuerste: If you already know, who is supposed to do the work, we should put her name in here...

negin-z: I have been working on a forked version of this component. It can be assigned to me, though I don't have the capacity for getting to it at the moment.

octree_web_viewer/src/state.rs
catevita: This is easy with an lru but the currently used hash map requires a key to access the item. The only alternative it is returning an iterator that randomly gets me all the values of the hashmap.  I'd like to preserve the right of the client to restart the browser and get deterministically in the same condition the program launched at any time, no matter how many keys were inserted. Can this be subject of a next pr?

octree_web_viewer/src/bin/points_web_viewer.rs
catevita: done!

octree_web_viewer/src/bin/points_web_viewer.rs
catevita: done!

octree_web_viewer/src/bin/points_web_viewer.rs
catevita: this is possible only if <DIR> is not specified. This assures that the full path which was the only input before continues to work. the group secification should clear this conflict

octree_web_viewer/src/state.rs
catevita: done

octree_web_viewer/Cargo.toml
catevita: done

octree_web_viewer/client/main.ts
catevita: If you agree negin, i'd put your name on that because you have a better understanding of what it has to be done in the future.

octree_web_viewer/src/bin/points_web_viewer.rs
catevita: unfortunately it is not possible
`no method named `to_string_lossy` found for type `std::path::Component<'_>` in the current scope
  --> octree_web_viewer/src/bin/points_web_viewer.rs:88:45
   |
88 |                     (tmp_octree_id.unwrap().to_string_lossy()),`

octree_web_viewer/src/backend.rs
catevita: this won't compile
`error[E0277]: the trait bound `actix_web::extractor::Path<std::string::String>: std::convert::AsRef<str>` is not satisfied
   --> octree_web_viewer/src/backend.rs:115:17
    |
115 |                 get_octree_from_state(&octree_id, &state).unwrap();
    |                 ^^^^^^^^^^^^^^^^^^^^^ the trait `std::convert::AsRef<str>` is not implemented for `actix_web::extractor::Path<std::string::String>``

octree_web_viewer/src/bin/points_web_viewer.rs
catevita: I thought that suffix could be accepted with both combinations (octree_directory or prefix+octree_id), but maybe it is confusing.

octree_web_viewer/src/bin/points_web_viewer.rs
catevita: it won't compile

octree_web_viewer/src/bin/points_web_viewer.rs
catevita: indeed, thanks. i'd  add a comment for this else clause, which is meant for the case that octree_directory is root.

octree_web_viewer/src/bin/points_web_viewer.rs
catevita: This is a good idea..can I implement it in the next pr or do you mean to edit here again? I'd prefer a quick pr cause here i am losing focus of all the different parts

catevita:
PTAL! @dudv

octree_web_viewer/client/main.ts
```diff
@@ -20,124 +20,203 @@ import { FirstPersonController } from './control';
...
+                credentials: 'same-origin',
+            }
+        );
+
+        const result = window
+            .fetch(request)
+            .then((response) => response.text()) // todo error handling?
+            .then((body) => {
+                return body;
+            });
```
negin-z: The code `.then((body) => {
    return body;
})` is excessive. would you remove it or put a //TODO to do so?

catevita:
@feuerste as discussed offline, the input args implementation is stripped off of this pr

octree_web_viewer/src/backend.rs
feuerste: Ah, I see, this is because `octree_id` is not a regular Rust `Path`, but an actix `Path`.

octree_web_viewer/src/bin/points_web_viewer.rs
feuerste: As discussed, the version with two or three parts will go into another PR.

octree_web_viewer/src/bin/points_web_viewer.rs
```diff
@@ -12,44 +12,69 @@
...
     ip: String,
+    #[structopt(default_value = "20", long = "cache_items")]
+    cache_max: usize,
+}
+
+/// init app state with command arguments
+/// backward compatibilty is ensured
+pub fn state_from(args: CommandLineArguments) -> Result<AppState, PointsViewerError> {
+    // initial implementation: suffix from args not yet supported
+    let suffix = PathBuf::from("");
```
feuerste: Nit: `PathBuf::from("");` -> `PathBuf::new();`

octree_web_viewer/src/bin/points_web_viewer.rs
```diff
@@ -12,44 +12,69 @@
...
+    let prefix_opt = args.octree_path.parent();
```
feuerste: How about using `let prefix = args.octree_path.parent().unwrap_or_default()` instead and then not branch at all, but only use the current first branch?

octree_web_viewer/src/state.rs
feuerste: Ok, let's leave it as it is for now.

octree_web_viewer/src/state.rs
```diff
@@ -0,0 +1,92 @@
...
+        let octree: Arc<octree::Octree> = Arc::from(octree::octree_from_directory(&addr)?);
+        {
+            // write access to state
+            let mut wmap = self.octree_map.write().unwrap();
+            wmap.insert(octree_key.clone(), Arc::clone(&octree));
+        }
+        Ok(octree)
+    }
+
+    pub fn return_init_id(&self) -> Result<String, PointsViewerError> {
```
feuerste: This function is not needed, as `init_octree_id` is `pub` anyways... I would either remove this function or make the members of `AppState` private.
If you keep this function, it doesn't need to return a `Result`, since it never gives an error. How about renaming it to `get_init_id`?

octree_web_viewer/client/main.ts
```diff
@@ -20,124 +20,203 @@ import { FirstPersonController } from './control';
...
+        this.camera.updateMatrixWorld(false);
+    }
+
+    private initScene() {
+        this.scene = new THREE.Scene();
+        this.scene.add(this.camera);
+    }
+
+    private initRenderer() {
+        const renderArea = document.getElementById('renderArea');
```
dudv: Please create a method `getViewportSize` returning `{width, height}` and reuse it when creating renderer and camera. Also you can make DOM element you are rendering to a class property. This way you'll get rid of those repeated `document.getElementById('renderArea')` calls.

octree_web_viewer/client/main.ts
```diff
@@ -20,124 +20,203 @@ import { FirstPersonController } from './control';
...
-        this.camera.matrixWorldInverse
-      );
-      this.viewer.frustumChanged(
-        matrix,
-        this.renderer.context.canvas.width,
-        this.renderer.context.canvas.height
-      );
+
+    private onWindowResize() {
+        this.camera.aspect = window.innerWidth / window.innerHeight;
```
dudv: Why are you using here the window dimensions and not dimensions of `document.getElementById('renderArea')`?

octree_web_viewer/src/state.rs
catevita: I made the members of app state private and renamed the function

octree_web_viewer/client/main.ts
catevita: I do not know why this was used here.

catevita: shall I put a todo here?

octree_web_viewer/src/bin/points_web_viewer.rs
```diff
@@ -12,44 +12,58 @@
...
+    #[structopt(default_value = "20", long = "cache_items")]
+    cache_max: usize,
+}
+
+/// init app state with command arguments
+/// backward compatibilty is ensured
+pub fn state_from(args: CommandLineArguments) -> Result<AppState, PointsViewerError> {
+    // initial implementation: suffix from args not yet supported
+    let suffix = PathBuf::new();
+    let prefix = args.octree_path.parent().unwrap_or_else(|| Path::new(""));
```
feuerste: Would `unwrap_or_default()` instead also work?

octree_web_viewer/src/bin/points_web_viewer.rs
```diff
@@ -12,44 +12,58 @@
...
+    let octree_id = args.octree_path.strip_prefix(&prefix)?;
```
feuerste: Did you test it with an empty prefix? I had pretty bad experiences with the `strip_prefix` method previously, just want to make sure that we don't run into problems again.

octree_web_viewer/src/bin/points_web_viewer.rs
catevita: it does not compile.

```
 note: the method `unwrap_or_default` exists but the following trait bounds were not satisfied:
           `&std::path::Path : std::default::Default`
```

octree_web_viewer/src/bin/points_web_viewer.rs
catevita: good point. I am double checking it.

catevita: I had it running fine with empty prefix.

catevita:
@wally-the-cartographer merge

PR #236, opened 2019-03-20
Web Browser Api with octree factory

catevita:
This implementation generalizes the way octrees can be created for the web browser.


catevita:
PTAL @feuerste

PR #237, opened 2019-03-21
ECEF octree

nnmm:
* The PLY reader can now parse an offset included in the header and apply it to each point

### DIFF
--- a/src/ply.rs
+++ b/src/ply.rs
@@ -27,6 +27,7 @@ use std::str;
 struct Header {
     format: Format,
     elements: Vec<Element>,
+    offset: Vector3<f64>,
 }
 
 #[derive(Debug, Copy, Clone, PartialEq)]
@@ -120,6 +121,7 @@ fn parse_header<R: BufRead>(reader: &mut R) -> Result<(Header, usize)> {
 
     let mut format = None;
     let mut current_element = None;
+    let mut offset = Vector3::new(0., 0., 0.);
     let mut elements = Vec::new();
     loop {
         line.clear();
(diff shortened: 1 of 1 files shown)


nnmm:
ptal @feuerste

src/bin/build_octree.rs
```diff
@@ -29,10 +29,19 @@ struct CommandlineArguments {
     /// This decides on the number of bits used to encode each node.
     #[structopt(long = "resolution", default_value = "0.001")]
     resolution: f64,
+    // Flag to skip bbox calculation and use a fixed bbox.
+    // The fixed bbox is extends 1048576 length units in each direction.
+    #[structopt(long = "fixed_bbox")]
```
feuerste: How about using
```rust
use structopt::clap::{_clap_count_exprs, arg_enum};

arg_enum! {
    #[derive(Debug)]
    pub enum RootBbox {
        FromData,
        EarthECEF,
    }
}
```
in `src/generation.rs` and here
```rust
#[structopt(
    long = "bbox_type",
    raw(possible_values = "&RootBbox::variants()", case_insensitive = "true"),
    default_value = "FromData"
)]
bbox_type: RootBbox,
...
build_octree_from_file(
    &pool,
    args.output_directory,
    args.resolution,
    args.input,
    args.bbox_type,
);
```
so we get the enum instance directly?

src/ply.rs
```diff
@@ -27,6 +27,7 @@ use std::str;
 struct Header {
     format: Format,
     elements: Vec<Element>,
+    offset: Vector3<f64>,
```
feuerste: I think in general it may be nicer to have the offset change in a separate PR, since the two are unrelated. What do you think?

src/ply.rs
```diff
@@ -172,7 +174,14 @@ fn parse_header<R: BufRead>(reader: &mut R) -> Result<(Header, usize)> {
                 current_element.as_mut().unwrap().properties.push(property);
             }
             "end_header" => break,
-            "comment" => (),
+            "comment" if entries.len() == 5 => {
```
feuerste: The two `if`s can go into one line:
```rust
"comment" if entries.len() == 5 && entries[1] == "offset:" => {
```

src/ply.rs
nnmm: This is now the offset PR.

src/ply.rs
```diff
@@ -351,28 +365,25 @@ fn open(ply_file: &Path) -> Result<(BufReader<File>, i64, Vec<ReadingFn>)> {
...
 
 impl PlyIterator {
     pub fn from_file<P: AsRef<Path>>(ply_file: P) -> Result<Self> {
-        let (reader, num_total_points, readers) = open(ply_file.as_ref())?;
-        Ok(PlyIterator {
-            reader,
-            readers,
-            num_total_points,
-        })
+        open(ply_file.as_ref())
```
feuerste: Would it make sense to remove the `open` method and move it here, since `from_file` now is a one liner with (almost the) same signature?

nnmm: I agree.

nnmm: Done.

nnmm:
@wally-the-cartographer merge

krzesi-mir:
@wally-the-cartographer merge

tblaette:
@wally-the-cartographer merge

PR #238, opened 2019-03-22
Fixes grpc server starving.

feuerste:
#224 introduced an RwLock. The grpc server potentially waited forever for write access, since the read lock wasn't released. This PR fixes this bug.


feuerste:
@wally-the-cartographer merge

wally-the-cartographer:

PR #239, opened 2019-03-22
Input arguments for points_viewer

catevita:
Here I tried to streamline how will a user launch the program.
Backward compatibility: no options: the path will be taken as is 

Serving multiple trees with common structure (ancestor and subfolder) is done by specifying the depth  of the subfolder structure. By default is such depth 0.
For example run
./target/points_viewer <common_path>/octree_id/<subfolder_structure> --suffix_depth =2
if depth =2 i expect <subfolder_structure> = subfolder1/subfolder2 
In subfolder 2 reside the files of the octree. by changing the identifier in the GUI another octree with
<common_path>/another_octree_id/subfolder1/subfolder2 can be displayed in the browser

Advantage: reducing the number and length of the inputs

### DIFF
--- a/octree_web_viewer/src/bin/points_web_viewer.rs
+++ b/octree_web_viewer/src/bin/points_web_viewer.rs
@@ -29,7 +29,6 @@ use structopt::StructOpt;
 #[structopt(name = "points_web_viewer", about = "Visualizing points")]
 pub struct CommandLineArguments {
     /// The octree directory to serve, including a trailing slash.
-    /// this overrides <--prefix> and <--octree_id> options
     #[structopt(name = "DIR", parse(from_os_str))]
     octree_path: PathBuf,
     /// Port to listen on.
--- a/octree_web_viewer/src/bin/points_web_viewer.rs
+++ b/octree_web_viewer/src/bin/points_web_viewer.rs
@@ -58,7 +58,7 @@ pub fn state_from(args: CommandLineArguments) -> Result<AppState, PointsViewerEr
         let mut octree_directory = args.octree_path.parent().unwrap_or_else(|| Path::new(""));
         while suffix_depth > 1 {
             octree_directory = octree_directory.parent().unwrap_or_else(|| Path::new(""));
-            suffix_depth = suffix_depth - 1;
+            suffix_depth -= 1;
         }
         suffix = args.octree_path.strip_prefix(&octree_directory)?;
         octree_path = PathBuf::from(octree_directory);
--- a/octree_web_viewer/src/bin/points_web_viewer.rs
+++ b/octree_web_viewer/src/bin/points_web_viewer.rs
@@ -18,13 +18,19 @@ use octree_web_viewer::utils::start_octree_server;
 use point_viewer::octree::OctreeFactory;
 use std::path::{Path, PathBuf};
 use std::sync::Arc;
+use structopt::clap::ArgGroup;
 use structopt::StructOpt;
 
 /// HTTP web viewer for 3d points stored in OnDiskOctrees
+///
+/// Octrees that share a common path and (optional) identical subfolder structure can be served
+/// just by giving a full directory path, <common_octree_path>/<octree_identifier>/<optional_subfolder_suffix> as <DIR>.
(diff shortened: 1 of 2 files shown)


catevita:
PTAL @feuerste

octree_web_viewer/src/bin/points_web_viewer.rs
```diff
@@ -21,10 +21,15 @@ use std::sync::Arc;
...
 /// HTTP web viewer for 3d points stored in OnDiskOctrees
+///
+/// Octrees that share a common path and (optional) identical subfolder structure can be served
+/// just by giving a full directory path, <common_octree_path>/<octree_identifier>/<optional_subfolder_suffix> as <DIR>.
+/// The option "--suffix_depth" allows to specify where the actual octree files reside in relation to the octree identifier.
 #[derive(StructOpt, Debug)]
 #[structopt(name = "points_web_viewer", about = "Visualizing points")]
 pub struct CommandLineArguments {
     /// The octree directory to serve, including a trailing slash.
+    /// this overrides <--prefix> and <--octree_id> options
```
feuerste: There is no more `--prefix` or `--octree_id`, right?

octree_web_viewer/src/bin/points_web_viewer.rs
```diff
@@ -33,18 +38,35 @@ pub struct CommandLineArguments {
...
 
 /// init app state with command arguments
 /// backward compatibilty is ensured
 pub fn state_from(args: CommandLineArguments) -> Result<AppState, PointsViewerError> {
-    // initial implementation: suffix from args not yet supported
-    let suffix = PathBuf::new();
-    let prefix = args.octree_path.parent().unwrap_or_else(|| Path::new(""));
     let octree_factory = OctreeFactory::new();
-    let octree_id = args.octree_path.strip_prefix(&prefix)?;
+    let mut suffix = Path::new("");
```
feuerste: How about
```rust
let mut octree_prefix_id = args.octree_path;
for _ in 0..args.suffix_depth {
    octree_prefix_id = octree_prefix_id.parent().unwrap_or_default();
}
let suffix = args.octree_path.strip_prefix(&octree_prefix_id)?;
let prefix = octree_prefix_id.parent().unwrap_or_default();
let octree_id = octree_prefix_id.strip_prefix(&prefix)?;
```
?

octree_web_viewer/src/bin/points_web_viewer.rs
catevita: exactly, this comment line has to go away

catevita:
> I am a little confused now, do you propose not to have `prefix` and `octree_id` any more?
I thought this implementation is easy and also quite user friendly for launching the program.
prefix and octree_id are obtained from the full path and there is less complexity when parsing the string or checking whether input arguments clash when used together.
I also have the other implementation with prefix and octree id + suffix as well, but I found this idea quite simple and appealing.

feuerste:
> I am a little confused now, do you propose not to have `prefix` and `octree_id` any more?
> > I thought this implementation is easy and also quite user friendly for launching the program.
> > prefix and octree_id are obtained from the full path and there is less complexity when parsing the string or checking whether input arguments clash when used together.
> > I also have the other implementation with prefix and octree id + suffix as well, but I found this idea quite simple and appealing.

I am not 100% convinced. Here we are in open source and we should try to keep things as simple as possible. In my opinion the server and client should work in a similar way, currently we would on the server side pass the full path plus a suffix depth, on the client side then suddenly change the octree id. It is not clear what part of the path (minus the suffix depth) the octree id is. To me it would just have been clearer in a similar way as the GRPC client/server works:
* If the server is started with a full path (including id and suffix), the client just sees one octree and can't change it.
* If the server is started with a path and an id (including suffix), the client first sees the octree specified by the id, but it can change it by specifying another id (including suffix again).

catevita:
This is a very good point. 

The client can visualize the octree ID in the GUI. 

octree with octree id

This could also be confusing. I think I have then to insert more logic in ts to have the client working as GRPC works, as well as a slight more complicated input argument extraction.

PR #240, opened 2019-03-22
Fixed bounding box

nnmm:
* Allow choosing a fixed bbox that encompasses the earth instead of determining the box from data
* The bbox dimensions were set to a number of meters such that the earth can fit in the bounding box

### DIFF
--- a/src/bin/build_octree.rs
+++ b/src/bin/build_octree.rs
@@ -30,7 +30,7 @@ struct CommandlineArguments {
     #[structopt(long = "resolution", default_value = "0.001")]
     resolution: f64,
     // Flag to skip bbox calculation and use a fixed bbox.
-    // The fixed bbox is extends 1048576 length units in each direction.
+    // The fixed bbox is extends 6400000 length units in each direction.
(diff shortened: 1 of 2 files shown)


nnmm:
ptal @feuerste

src/bin/build_octree.rs
```diff
@@ -29,10 +29,24 @@ struct CommandlineArguments {
     /// This decides on the number of bits used to encode each node.
     #[structopt(long = "resolution", default_value = "0.001")]
     resolution: f64,
+    // Flag to skip bbox calculation and use a fixed bbox.
+    // The fixed bbox is extends 6400000 length units in each direction.
```
feuerste: ... -> `The fixed bbox extents have 6_400_000 length units in each direction, so they cover the whole earth when using ECEF coordinates.`?

nnmm:
@wally-the-cartographer merge

tblaette:
@wally-the-cartographer merge

wally-the-cartographer:
Merge requested by authorized user tblaette. Merge queue now has a length of 1.

PR #245, opened 2019-03-27
Support for XBox 360 controller (USB only).

SirVer:
Only tested on Mac.

You need to install a driver from here before this will work: 
https://github.com/360Controller/360Controller/releases

This is not very flexible and it could be that under Linux the scaling works differently so this does not feel nice. However, since this is only used for demos I did not feel like adding dynamic configuration for the scaling values of the different controllers into the code at this point in time.

### DIFF
--- a/sdl_viewer/src/lib.rs
+++ b/sdl_viewer/src/lib.rs
@@ -314,10 +314,10 @@ struct XBoxJoystick {
 
 impl Joystick for XBoxJoystick {
     fn act(&self, camera: &mut Camera) {
-        let right = f32::from(self.joystick.axis(0).unwrap()) / 1000.;
-        let forward = f32::from(self.joystick.axis(1).unwrap()) / 1000.;
-        let turning_right = -f32::from(self.joystick.axis(3).unwrap()) / 32000.;
-        let turning_up = -f32::from(self.joystick.axis(4).unwrap()) / 32000.;
+        let right = f64::from(self.joystick.axis(0).unwrap()) / 1000.;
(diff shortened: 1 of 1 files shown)


catevita:
the driver did get me goosebumps, i took the latest dmg and i had to reboot the mac several times before I could start it again.. do you suggest some version of the driver?

SirVer:
I just used the latest one and it just worked for me.

There is no need to test code that you review, it is the author's responsibility to make sure the code is working.

SirVer:
thanks for the review!

SirVer:
@wally-the-cartographer merge

SirVer:
@cschuet Wally down again?

gaschler:
Wally may be down in addition, but you have a build error

```
    Checking point_viewer_grpc v0.1.0 (/home/travis/build/googlecartographer/point_cloud_viewer/point_viewer_grpc)
error[E0308]: mismatched types
   --> sdl_viewer/src/lib.rs:321:20
    |
321 |         camera.pan(right, 0., forward);
    |                    ^^^^^ expected f64, found f32
help: you can cast an `f32` to `f64` in a lossless way
    |
321 |         camera.pan(right.into(), 0., forward);
    |                    ^^^^^^^^^^^^
```

SirVer:
@gaschler Thanks! But it is down, otherwise it would have posted here that not all checks pass - at least my initial implementation of it did that.

SirVer:
@wally-the-cartographer merge

PR #246, opened 2019-03-27
Fix for subsampling resulting in empty nodes

nnmm:
Subsampling would sometimes result in empty nodes, when there is a long chain of nodes with only one child, as you'd get with big bounding boxes. However, octree queries work by recursing into the octree, starting from the root node, and take nonexistence as an indication that no nodes exist below that points.

### DIFF
--- a/src/generation.rs
+++ b/src/generation.rs
@@ -394,9 +394,7 @@ pub fn build_octree(
         pool.scoped(|scope| {
             scope.execute(|| {
                 for (id, num_points) in finished_nodes_receiver {
-                    if num_points > 0 {
-                        finished_nodes.insert(id, num_points);
-                    }
+                    finished_nodes.insert(id, num_points);
                 }
             });
 
--- a/src/octree/mod.rs
+++ b/src/octree/mod.rs
@@ -479,7 +479,7 @@ fn maybe_push_node(
     node: Node,
     projection_matrix: &Matrix4<f32>,
 ) {
-    if !nodes.contains_key(&node.id) {
+    if nodes.get(&node.id).map_or(0, |meta| meta.num_points) == 0 {
         return;
     }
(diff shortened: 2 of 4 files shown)


nnmm:
ptal @SirVer

nnmm:
fyi @feuerste @catevita

SirVer:
I do not understand how there can be empty nodes. And if they are, I would rather fix this bug.

nnmm:
@SirVer  The problem would arise when there was a chain  rABCDEFG, rABCDEFGH, rABCDEFGHI, rABCDEFGHI, rABCDEFGHIJ, rABCDEFGHIJK of nodes with only one child. One-eight of points would always get pushed up into the parent node, until at some point there was only one point being pushed into the parent node. That point would get pushed further up, because points to push up during subsampling were selected by taking the points index modulo 8, so the first point would always get subsampled – but even if it wasn't pushed up, the nodes above would be the empty ones instead. So emptiness is hard to avoid. This solution is not the only one, and probably not the fastest, but it works.

SirVer:
We had an offline discussion about this and considered various alternatives. We came to the following conclusions:

While we had plenty of ideas how we can work around empty nodes in our current scheme, it feels that we need to eventually deal with them anyhow, because it is easy to construct cases where we will encounter them (e.g. merging octrees or ill distributed data).

We decided to go ahead and deal with empty nodes (i.e. going ahead with nikolai's approach), but put in some effort to prune empty nodes at the leaf of the octree at creation time.

catevita:
LGTM! thanks for the fix, @feuerste the modification in node drawer is not needed anymore

nnmm:
@wally-the-cartographer merge

wally-the-cartographer:

PR #248, opened 2019-04-16
Replace iterators with default ones for more flexibility.

feuerste:
- The current node, point, and file iterators are based on a custom one only providing a `for_each` method.
- This is pretty inflexible, in particular when we want to cancel iteration, which is e.g. possible with the default iterators by using [`try_for_each`](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.try_for_each).
- Furthermore, by moving to default iterators we get all the functionality of the [iterator trait](https://doc.rust-lang.org/std/iter/trait.Iterator.html).
- A drawback is that the new iterators are slightly slower, since more function calls are involved:
  - `target/release/build_octree /Users/mfeuerstein/dev/slam_assets/e318a37b-a481-4b93-8d07-f0358b71bf22/pointcloud_just_colors.ply --output_directory /Users/mfeuerstein/dev/slam_assets/e318a37b-a481-4b93-8d07-f0358b71bf22/octree_just_colors/`:
    - old:
       User time (seconds): 979.59
       System time (seconds): 251.63
       Percent of CPU this job got: 506%
       Elapsed (wall clock) time (h:mm:ss or m:ss): 4:02.86
    - new:
	User time (seconds): 1178.92
	System time (seconds): 242.69
	Percent of CPU this job got: 517%
	Elapsed (wall clock) time (h:mm:ss or m:ss): 4:34.66
  - `target/release/octree_benchmark /Users/mfeuerstein/dev/slam_assets/fd93587d-2c3d-48f9-8e52-1f9c0f541d86/octree_just_colors/ --no-client`
    - old:
	User time (seconds): 2.16
	System time (seconds): 0.32
	Percent of CPU this job got: 99%
	Elapsed (wall clock) time (h:mm:ss or m:ss): 0:02.49
    - new:
	User time (seconds): 3.95
	System time (seconds): 0.33
	Percent of CPU this job got: 99%
	Elapsed (wall clock) time (h:mm:ss or m:ss): 0:04.29
  - `target/release/octree_benchmark /Users/mfeuerstein/dev/slam_assets/fd93587d-2c3d-48f9-8e52-1f9c0f541d86/octree_just_colors/`
    - old:
	User time (seconds): 10.75
	System time (seconds): 2.81
	Percent of CPU this job got: 113%
	Elapsed (wall clock) time (h:mm:ss or m:ss): 0:11.99
    - new:
	User time (seconds): 10.28
	System time (seconds): 2.70
	Percent of CPU this job got: 112%
	Elapsed (wall clock) time (h:mm:ss or m:ss): 0:11.56
- If we later need an iterator pointing to references of points, we can consider switching to the https://docs.rs/streaming-iterator/0.1.4/streaming_iterator/

### DIFF
--- a/src/octree/mod.rs
+++ b/src/octree/mod.rs
@@ -148,7 +148,7 @@ pub struct FilteredPointsIterator<'a> {
     octree: &'a Octree,
     filter_func: Box<Fn(&Point) -> bool + 'a>,
     node_ids: VecDeque<NodeId>,
-    node_iterator: Option<NodeIterator>,
+    node_iterator: NodeIterator,
 }
 
 impl<'a> FilteredPointsIterator<'a> {
@@ -161,7 +161,7 @@ impl<'a> FilteredPointsIterator<'a> {
             octree,
             filter_func,
             node_ids,
-            node_iterator: None,
+            node_iterator: NodeIterator::Empty,
         }
     }
 }
@@ -171,37 +171,15 @@ impl<'a> Iterator for FilteredPointsIterator<'a> {
 
     fn next(&mut self) -> Option<Point> {
         loop {
-            match &mut self.node_iterator {
-                Some(node_iterator) => {
-                    match node_iterator.next() {
-                        Some(point) => {
-                            if (self.filter_func)(&point) {
-                                return Some(point);
-                            }
-                        }
-                        None => self.node_iterator = None,
-                    };
+            while let Some(point) = self.node_iterator.next() {
+                if (self.filter_func)(&point) {
+                    return Some(point);
                 }
-                None => match self.node_ids.pop_front() {
-                    Some(node_id) => {
-                        self.node_iterator = Some(get_node_iterator(self.octree, &node_id));
-                    }
-                    None => return None,
-                },
             }
-        }
-    }
-}
-
-fn add_children(
-    parent_id: &NodeId,
-    octree_nodes: &FnvHashMap<NodeId, NodeMeta>,
-    node_ids: &mut Vec<NodeId>,
-) {
-    for child_index in 0..8 {
-        let child_id = parent_id.get_child_id(ChildIndex::from_u8(child_index));
-        if octree_nodes.contains_key(&child_id) {
-            node_ids.push(child_id);
+            self.node_iterator = match self.node_ids.pop_front() {
+                Some(node_id) => get_node_iterator(self.octree, &node_id),
+                None => return None,
+            };
         }
     }
 }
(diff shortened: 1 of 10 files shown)


feuerste:
> LGTM! Thanks for the implementation. I think that the overhead is worth the gain of the functionality. Or is there some use case that is critical that I haven't considered?

I don't think we will have a critical use case. For small octrees, which may live on disk, the difference will be marginal, for huge octrees, which have to live in the cloud, the main performance bottleneck (as you can see for the last of the three tests) will be the networking overhead.

src/octree/mod.rs
```diff
@@ -133,94 +133,92 @@ pub struct Octree {
     nodes: FnvHashMap<NodeId, NodeMeta>,
 }
 
-pub struct PointsInBoxIterator<'a> {
-    octree: &'a Octree,
-    aabb: &'a Aabb3<f64>,
-    intersecting_nodes: Vec<NodeId>,
+fn get_node_iterator(octree: &Octree, node_id: &NodeId) -> NodeIterator {
```
nnmm: Would it be awkward to make this a method of `Octree`?

nnmm:
I like how this stops the proliferation of different iterators – it makes the overall design cleaner. A few questions:
- The octree benchmark only measures the AllPointsIterator. Do you think the filtering queries stay about the same, performance-wise?
- What do you think about making the iterator structs implement both `Iterator` and `InternalIterator` – would that eliminate the performance degradation?

src/octree/mod.rs
feuerste: You mean a private or public member?
I wouldn't make it public, because this is something nobody outside should need to use directly... if private, I don't have a tendency what's better...

feuerste:
> * The octree benchmark only measures the AllPointsIterator. Do you think the filtering queries stay about the same, performance-wise?

I think they will also be a little slower, since we also have more function calls here.

> * What do you think about making the iterator structs implement both `Iterator` and `InternalIterator` – would that eliminate the performance degradation?

This would improve performance, but only for the `for_each` case, we would then also have to implement the `try_for_each` on top of this. Furthermore, we would have massive code duplication...

src/octree/node_iterator.rs
```diff
@@ -65,24 +66,32 @@ impl NodeIterator {
...
+    fn size_hint(&self) -> (usize, Option<usize>) {
+        let num_points = match self {
+            NodeIterator::WithData(ref iter) => iter.meta.num_points as usize,
+            NodeIterator::Empty => 0,
+        };
+        (num_points, Some(num_points))
     }
-    fn for_each<F: FnMut(&Point)>(self, mut f: F) {
+    fn next(&mut self) -> Option<Point> {
         let mut iter = match self {
```
nnmm: I think the slowdown comes from effectively pulling everything from here until the match statement into the innermost loop. Pulling them out again would require caching `edge_length`, `min` and `num_points as usize` in the iterator. The `WithData` indirection can probably only be avoided by directly storing `NodeIteratorWithData` in the high-level iterators. Do you think any of that would be worth doing?

feuerste: `edge_length`, `min`, and `num_points` are only double or 3d doubles, which shouldn't be much overhead... For the rest I'm not sure whether this is worth the effort.

feuerste:
@wally-the-cartographer merge

wally-the-cartographer:

feuerste:
@nnmm @catevita Thanks for your reviews!

PR #249, opened 2019-04-17
Oriented beam query

nnmm:
This is an implementation of oriented beam queries. An oriented beam is the same thing as an ortho frustum, so it is not as specific as it might sound.

Much like the box query, it only opens nodes intersected by the queried volume to be efficient. To do this, a specialized implementation of a collision check between two oriented cubes/boxes is provided.


nnmm:
ptal @catevita

src/octree/mod.rs
```diff
@@ -164,6 +164,37 @@ impl<'a> InternalIterator for PointsInBoxIterator<'a> {
...
 }
 
+pub struct PointsInOrientedBeamIterator<'a> {
+    octree: &'a Octree,
+    beam: &'a OrientedBeam,
+    intersecting_nodes: Vec<NodeId>,
+}
+
+impl<'a> InternalIterator for PointsInOrientedBeamIterator<'a> {
+    fn size_hint(&self) -> Option<usize> {
```
catevita: would it maybe be more useful to implement it already with the std Iterator trait?

nnmm: Yes, I'll have to adapt to Marco's change.

src/math.rs
```diff
@@ -67,3 +67,137 @@ impl Cube {
         )
     }
 }
+
+#[derive(Debug, Clone)]
+pub struct OrientedBeam {
```
feuerste: Do you think it makes sense to specify the OrientedBeam as similar as possible as https://docs.rs/collision/0.20.1/collision/struct.Obb.html ? This would mean that we need a Vector2 for the extents, a Vector3 as axis, and a Point3 as some start/center/translation...

src/octree/mod.rs
nnmm: Done.

src/math.rs
nnmm: That's a good question. A minimal representation would need even one coordinate less; e.g. the xy position of the world origin in beam coordinates, but probably harder to specify in the client. It's also not clear to me how the Obb represents rotations around the `axis`. Vector length?

If we're optimizing for computational efficiency, I should store the conjugate of the rotation, or the inverse transformation matrix. Do you think I should have different proto and Rust representations to achieve both?

feuerste: Hm, I think the cleanest at the moment would be just to change
```rust
    pub rotation: Quaternion<f64>,
    pub translation: Vector3<f64>,
    pub half_extent_x: f64,
    pub half_extent_y: f64,
```
to
```rust
    pub rotation: Quaternion<f64>,
    pub translation: Point3<f64>,
    pub extents: Vector2<f64>,
```
What do you think?

nnmm: I have no objections to the extent being a Vector2 but I'm unsure why the translation should be a Point3 instead of a Vector3.

feuerste: This would be in accordance with the cgmath and collision libs, where (as far as I understand it) a point has absolute Euclidean coordinates, whereas a vector is just a vector defining relative coordinates.

nnmm: Yes, and I think we want it to be a vector. That's the type that's used to transform Point3s, which is the type of points in the point cloud. See e.g. https://docs.rs/cgmath/0.17.0/cgmath/prelude/trait.Transform3.html

feuerste: Isn't `translation` the starting position of the oriented beam box?

src/math.rs
```diff
@@ -67,3 +67,137 @@ impl Cube {
...
+    // Translation and orientation together form the transform from "beam coordinates" into world coordinates
+    pub rotation: Quaternion<f64>,
```
catevita: could it be coupled, consisting in Decomposed<Vector3<f64>, Quaternion<f64>>?

nnmm: Yes, I think that makes sense.

src/math.rs
nnmm: Right, both viewpoints are valid. I find it easier to imagine translation + rotation as the transform that takes local coordinates, wherein the beam is vertical, to global coordinates. This is also the viewpoint of my use case for this query. There's also an argument that since that is also how the translation will be used (i.e. for transforming points) this representation will save an additional cast.

So, I'd suggest we go with a transform Decomposed<Vector3, Quaternion> that represents the transform from global into local ("beam") coordinates, simply because that direction will not require an additional inversion when transforming each point. Perhaps we can also make the fields private if you think there is a danger of them being misinterpreted.

feuerste: [`Decomposed`](https://docs.rs/cgmath/0.17.0/cgmath/struct.Decomposed.html) also includes a scaling factor, which we don't need. Then I would rather stick with a `Vector3` and a `Quaternion`.

nnmm:
As discussed offline, I changed the members of OrientedBeam to be more geared towards efficiently computing `intersects()` and `contains()`, meaning some redundancy, and made the members private.

nnmm:
@pifon2a Do you know why my commit gets automatically reverted?

pifon2a:
@nnmm Need more data to answer. What happened exactly?

nnmm:
@pifon2a My branch is at ddbe037, but if you look above, your account force-pushed 3042bb6, the previous commit right after I pushed to my branch.

nnmm:
Weird. When I look at the files changed tab, I see my latest changes now. I seem to remember that wasn't the case before. So the messages are probably harmless. Sorry for the ping.

pifon2a:
It could be because our merge bot Wally went insane. Thanks for noticing. I will clear up its cache and restart it. It acts from my account, that's why you see "pifon2a".

src/math.rs
```diff
@@ -67,3 +67,141 @@ impl Cube {
...
+            }
+            // Project four "vertices" of the beam onto that axis
+            let mut beam_min_proj = std::f64::MAX;
+            let mut beam_max_proj = std::f64::MIN;
+            for corner in self.corners.iter() {
+                let corner_proj = corner.dot(*sep_axis);
+                beam_min_proj = beam_min_proj.min(corner_proj);
+                beam_max_proj = beam_max_proj.max(corner_proj);
+            }
+            separated |= beam_min_proj > cube_max_proj || beam_max_proj < cube_min_proj;
```
feuerste: In case `separated` is `true`, we can break early here, right?

src/math.rs
```diff
@@ -67,3 +67,141 @@ impl Cube {
...
+        x.abs() <= self.half_extent.x && y.abs() <= self.half_extent.y
+    }
+
+    fn precompute_corners(
+        rotation: &Quaternion<f64>,
+        translation: &Vector3<f64>,
+        half_extent: &Vector2<f64>,
+    ) -> [Point3<f64>; 4] {
+        [
+            rotation.rotate_point(Point3::new(half_extent.x, half_extent.y, 0.0)) + translation,
```
feuerste: How about a small helper fn or closure taking `x` and `y`?

nnmm: Sure

src/math.rs
nnmm: Yes

nnmm:
@wally-the-cartographer merge

wally-the-cartographer:

PR #251, opened 2019-05-02
Handle special case of OrientedBeam query

nnmm:
When the orientation of the beam is aligned with a basis vector, the cross product becomes zero, and normalization produces NaNs. This taints the intersection computation and gives incorrect results (no points).


nnmm:
ptal @feuerste

nnmm:
@wally-the-cartographer merge

wally-the-cartographer:

PR #253, opened 2019-05-06
Bugfix for ECEF octrees: visible nodes traversal

catevita:
Nikolai found out that the tree traversal for visible nodes in ecef was not successful when an empty node was encountered. Here the correction.

### DIFF
--- a/src/octree/mod.rs
+++ b/src/octree/mod.rs
@@ -358,7 +358,12 @@ impl Octree {
                     unreachable!();
                 }
             };
-            if self.nodes.get(&current.node.id).map_or(0, |meta| meta.num_points) > 0 {
+            if self
+                .nodes
+                .get(&current.node.id)
+                .map_or(0, |meta| meta.num_points)
+                > 0
+            {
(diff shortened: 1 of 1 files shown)


catevita:
PTAL @nnmm

catevita:
PTAL @feuerste

nnmm:
I took a second look and think the following accomplishes the fix in a cleaner way: https://github.com/googlecartographer/point_cloud_viewer/compare/nnmm/visible_nodes_bugfix

That way, we do one less hash map lookup. Feel free to cherry-pick it into the PR.

catevita:
indeed! the latest solution you presented is also much more readable.

nnmm:
:+1:

catevita:
@nnmm if it was an approval it needs to be explicit

catevita:
@wally-the-cartographer merge

wally-the-cartographer:
Merge requested by authorized user catevita. Merge queue now has a length of 1.

PR #256, opened 2019-05-06
Bugfix octree factory

catevita:
this change let the errors  in the factory to be seen in the backtrace and not silently removed by the call of ondiskoctree.


catevita:
PTAL @nnmm

catevita:
duplicate

PR #259, opened 2019-05-09
Batch Iterator for Octree 

catevita:
Implementation and test of the batch iterator

### DIFF
--- a/src/octree/batch_iterator.rs
+++ b/src/octree/batch_iterator.rs
@@ -79,6 +79,14 @@ where
         };
         (self.func)(point_data)
     }
+
+    fn push_point_and_callback(&mut self, point: Point) -> Result<()> {
+        self.push_point(point);
+        if self.position.len() == self.position.capacity() {
+            return self.callback();
+        }
+        Ok(())
+    }
 }
 
 /// Iterator on point batches
@@ -135,6 +143,15 @@ impl<'a> BatchIterator<'a> {
         // call on the last batch
         point_stream.callback()
     }
+
+    pub fn try_for_each_batch2<F>(&mut self, mut func: F) -> Result<()>
+    where
+        F: FnMut(PointData) -> Result<()>,
+    {
+        let mut point_stream = PointStream::new(self.batch_size, &mut func);
+        self.iterator
+            .try_for_each(|point: Point| point_stream.push_point_and_callback(point))
+    }
 }
 
 #[cfg(test)]
@@ -145,9 +162,7 @@ mod tests {
     use cgmath::Point3;
     use tempdir::TempDir;
 
-    #[test]
-    #[ignore]
-    fn test_batch_iterator() {
+    fn build_test_octree(batch_size: usize) -> Box<Octree> {
         let default_point = Point {
             position: Vector3::new(-2_699_182.0, -4_294_938.0, 3_853_373.0), //ECEF parking lot porter dr
             color: Color {
@@ -159,7 +174,6 @@ mod tests {
             intensity: None,
         };
 
-        let batch_size = NUM_POINTS_PER_BATCH / 10;
         let mut points = vec![default_point; 4 * batch_size + 1];
         points[3 * batch_size].position = Vector3::new(-2_702_846.0, -4_291_151.0, 3_855_012.0); // ECEF STANFORD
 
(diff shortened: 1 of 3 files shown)


catevita:
ptal @feuerste

src/octree/batch_iterator.rs
```diff
@@ -0,0 +1,206 @@
...
+            position: self.position.split_off(0),
+            layers,
+        };
+        (self.func)(point_data)
+    }
+}
+
+/// Iterator on point batches
+pub struct BatchIterator<'a> {
+    _octree: &'a Octree,
```
feuerste: `_octree` -> `octree`?

src/octree/batch_iterator.rs
```diff
@@ -0,0 +1,206 @@
...
+    iterator: Box<Iterator<Item = Point> + 'a>,
+    batch_size: usize,
+}
+
+impl<'a> BatchIterator<'a> {
+    pub fn new(
+        octree: &'a octree::Octree,
+        location: &'a octree::batch_iterator::PointLocation,
+        size: usize,
```
feuerste: `size` -> `batch_size`

src/octree/batch_iterator.rs
```diff
@@ -0,0 +1,206 @@
...
+    ) -> Self {
+        let iter: Box<Iterator<Item = Point>> = match location {
```
feuerste: `iter` -> `iterator`

src/octree/batch_iterator.rs
```diff
@@ -0,0 +1,206 @@
...
+            PointLocation::Any() => Box::new(octree.all_points()),
+            PointLocation::Aabb(aabb) => Box::new(octree.points_in_box(aabb)),
+            PointLocation::Frustum(frustum) => Box::new(octree.points_in_frustum(frustum)),
+            PointLocation::OrientedBeam(beam) => Box::new(octree.points_in_oriented_beam(beam)),
+        };
+        BatchIterator {
```
feuerste: -> `BatchIterator { octree, iterator, batch_size }`

src/octree/batch_iterator.rs
```diff
@@ -0,0 +1,206 @@
...
+        }
+    }
+
+    /// compute a funtion while iterating on a batch of points
+    pub fn try_for_each_batch<F>(&mut self, mut func: F) -> Result<()>
+    where
+        F: FnMut(PointData) -> Result<()>,
+    {
+        let mut point_stream = PointStream::new(self.batch_size, &mut func);
+        'octree_loop: loop {
```
feuerste: Is there any specific reason why you decided not to use
`iter.try_for_each(|point: Point| point_stream.push_point_and_callback(point))?;`
from the original implementation?
This loop seems to be harder to understand...

catevita: Indeed ..My intention is to spare the call of the method(s) and the checks whether the batch function has to be run.

src/octree/batch_iterator.rs
catevita: the variable is currently unused, I do not know if it is better to remove this at all

src/octree/batch_iterator.rs
catevita: done!

src/octree/batch_iterator.rs
catevita: done!

src/octree/batch_iterator.rs
catevita: done!

catevita: ..sort of `BatchIterator { _octree: octree, iterator, batch_size }`

src/octree/batch_iterator.rs
catevita: I did a test and i spare already 50 ms for an octree of 200_000 points and batch size of 50_000
Current 343.306956ms
Try for each 394.974788ms

src/octree/batch_iterator.rs
feuerste: Yes, I would remove it

src/octree/batch_iterator.rs
feuerste: Did you rerun the test multiple times for each variant?

src/octree/batch_iterator.rs

src/octree/batch_iterator.rs

src/octree/batch_iterator.rs
feuerste: I just tried the test myself wrapping it inside a loop calling it 100 times each. The new implementation calls the `callback` one time less then the original implementation, which would need to be fixed. This difference in calling the callback actually leads to the small time difference, since printing to screen takes quite a while. I would therefore go with the old implementation.

catevita: Oh that is true, it is missing in line 132, where the error is returned! Thank you for precious time and I am applying the changes as you suggest!

feuerste:
@wally-the-cartographer merge

feuerste:
@cschuet @pifon2a wally seems to be down... Can one of you please restart it? Thanks a lot!

feuerste:
@wally-the-cartographer merge

catevita:
@wally-the-cartographer merge

PR #262, opened 2019-05-13
Change depth to breadth first search when iterating over points.

feuerste:
This will give more representative points when only querying a part of the point cloud, as returned points spatially cover more of the requested location.

### DIFF
--- a/src/octree/mod.rs
+++ b/src/octree/mod.rs
@@ -392,10 +392,10 @@ impl Octree {
     /// Returns the ids of all nodes that cut or are fully contained in 'aabb'.
     pub fn points_in_box<'a>(&'a self, aabb: &'a Aabb3<f64>) -> FilteredPointsIterator<'a> {
         let mut node_ids = VecDeque::new();
-        let mut open_list = VecDeque::new();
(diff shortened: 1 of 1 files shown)


src/octree/mod.rs
```diff
@@ -417,19 +418,20 @@ impl Octree {
         beam: &'a OrientedBeam,
     ) -> FilteredPointsIterator<'a> {
         let mut node_ids = VecDeque::new();
-        let mut open_list = vec![Node::root_with_bounding_cube(Cube::bounding(
+        let mut open_list = VecDeque::new();
+        open_list.push_back(Node::root_with_bounding_cube(Cube::bounding(
```
nnmm: Could we also use vec![...].into() here like above?

feuerste: Yes, it doesn't make much difference though in terms of readability... I adjusted it.

feuerste:
@wally-the-cartographer merge

feuerste:
@wally-the-cartographer merge

feuerste:
@wally-the-cartographer merge

feuerste:
@wally-the-cartographer merge

PR #263, opened 2019-05-14
Add OBB.

feuerste:
We want to extend iterators to take local transformations. In order to implement all of them properly, we need an object oriented bounding box, so we can transform an aabb properly.

### DIFF
--- a/src/math.rs
+++ b/src/math.rs
@@ -269,7 +269,7 @@ impl OrientedBeam {
 #[cfg(test)]
 mod tests {
     use super::*;
-    use cgmath::Zero;
+    use cgmath::{Rad, Rotation3, Zero};
     // These tests were created in Blender by creating two cubes, naming one
     // "Beam" and scaling it up in its local z direction. The corresponding
     // object in Rust was defined by querying Blender's Python API with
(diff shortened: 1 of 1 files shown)


src/math.rs
```diff
@@ -78,7 +201,7 @@ pub struct OrientedBeam {
     translation_inv: Vector3<f64>,
     half_extent: Vector2<f64>,
     corners: [Point3<f64>; 4],
-    separating_axes: [Vector3<f64>; 5],
+    separating_axes: Vec<Vector3<f64>>,
```
nnmm: What's the benefit of a vector?

nnmm: Ah, you're filtering before intersects().

src/math.rs
```diff
@@ -68,6 +71,126 @@ impl Cube {
...
+            unit_x.cross(rot_y).normalize(),
+            unit_x.cross(rot_z).normalize(),
+            unit_y.cross(rot_x).normalize(),
+            unit_y.cross(rot_y).normalize(),
+            unit_y.cross(rot_z).normalize(),
+            unit_z.cross(rot_x).normalize(),
+            unit_z.cross(rot_y).normalize(),
+            unit_z.cross(rot_z).normalize(),
+        ] {
+            let is_finite_and_non_parallel = is_finite(&axis)
```
nnmm: Is the non-parallel condition just an optimization?

src/math.rs
feuerste: Yes, so we can avoid the check later.

src/math.rs
feuerste: Yes, it's an optimization to get rid of as many axes to check before actually performing the intersection tests.

feuerste:
@wally-the-cartographer merge

feuerste:

PR #265, opened 2019-05-15
Introduce transforms of obb/oriented beam and conversion from aabb to obb.

feuerste:
This is required to support passing local transforms to iterators.


src/math.rs
```diff
@@ -225,17 +264,22 @@ impl OrientedBeam {
...
+        isometry: &Isometry3<f64>,
         half_extent: &Vector2<f64>,
     ) -> [Point3<f64>; 4] {
-        let transform =
-            |x: f64, y: f64| rotation.rotate_point(Point3::new(x, y, 0.0)) + translation;
+        let transform = |x: f64, y: f64| {
+            isometry.rotation.rotate_point(Point3::new(x, y, 0.0)) + isometry.translation
+        };
         [
             transform(half_extent.x, half_extent.y),
```
catevita: could be this transform be confused with the function in line 272? could it be more explicit like get_corner_from_offset

feuerste: Yes, definitely, I renamed it.

src/octree/batch_iterator.rs
```diff
@@ -91,36 +111,58 @@ where
...
+                        disp: global_from_local.translation,
+                    }) * frustum,
+                ),
+                PointCulling::OrientedBeam(beam) => {
+                    PointCulling::OrientedBeam(beam.transform(global_from_local))
+                }
+            },
+            None => location.culling.clone(),
         };
+        let local_from_global = location.global_from_local.clone().map(|t| t.inverse());
```
catevita: what does it happen if location.global_from_local is None?

catevita: sorry i found it, it maps Option<T> to Option<U

feuerste: `map` leaves the `None` untouched, see https://doc.rust-lang.org/std/option/enum.Option.html#method.map
The point stream internally handles it accordingly.

nnmm:
:+1:

feuerste:
@wally-the-cartographer merge

PR #266, opened 2019-05-17
tidying up src/octree/mod.rs

catevita:
splitting mod file into mod and iterator file


catevita:
PTAL @nnmm

catevita:
@wally-the-cartographer merge

PR #267, opened 2019-05-20
Add point cloud client.

feuerste:
This PR adds a point cloud client capable of dealing with multiple octrees in the same coordinate frame. In addition, a factory is passed which can be extended to handle octrees coming from other sources than local or grpc.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1511,7 +1511,7 @@ dependencies = [
 ]
 
 [[package]]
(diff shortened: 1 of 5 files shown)


feuerste:
@wally-the-cartographer merge

feuerste:
@cschuet Can you please reactivate wally and/or merge? Thanks a lot!

PR #268, opened 2019-05-21
Iterator by Node ID

catevita:
The node file id iterator allows traversal by accessing node id an meta. This is a pre step useful for node parallelization (using e.g. rayon and par_iter())

Test were moved into a single file to allow for less code duplication


catevita:
PTAL @feuerste @nnmm

src/octree/octree_iterator.rs
```diff
@@ -7,7 +7,7 @@ use collision::Aabb3;
 use std::collections::VecDeque;
 
 /// returns an Iterator over the points of the current node
-fn get_node_iterator(octree: &Octree, node_id: &NodeId) -> NodeIterator {
+pub fn get_node_iterator(octree: &Octree, node_id: &NodeId) -> NodeIterator {
```
nnmm: Why did this become public?

src/octree/octree_test.rs
```diff
@@ -0,0 +1,79 @@
...
+
+        let pool = scoped_pool::Pool::new(10);
+        let tmp_dir = TempDir::new("octree").unwrap();
+
+        build_octree(&pool, &tmp_dir, 1.0, bounding_box, points.into_iter());
+        crate::octree::on_disk::octree_from_directory(tmp_dir.into_path()).unwrap()
+    }
+
+    #[test]
+    #[ignore]
```
nnmm: I don't understand why this test is ignored now – does your change impact the batch iterator?

nnmm:
I think abstracting away traversal of the nodes into an iterator is potentially a good idea, we do that a lot. However, it's currently unused – could you comment a bit more on how this should be used?

src/octree/octree_test.rs
catevita: the test tooks something longer to be tested. should that always be running?

src/octree/octree_iterator.rs
catevita: here i forgot to remove some previous changes

catevita:
hi nikolai,

> I think abstracting away traversal of the nodes into an iterator is potentially a good idea, we do that a lot. However, it's currently unused – could you comment a bit more on how this should be used?

in another branch I am currently implementing the parallelization... PR comes soon.
Either should par_iter from rayon be directly used, or should the candidate nodes be gathered and then should the threads reading memory be parallelized.

catevita:
https://github.com/googlecartographer/point_cloud_viewer/pull/273 here the PR

src/octree/octree_iterator.rs
catevita: fixed and, most importantly, pushed to origin

src/octree/mod.rs
```diff
@@ -43,12 +43,15 @@ pub use self::factory::OctreeFactory;
...
 mod octree_iterator;
 pub use self::octree_iterator::{
-    contains, intersecting_node_ids, AllPointsIterator, FilteredPointsIterator,
+    contains, intersecting_node_ids, AllPointsIterator, FilteredPointsIterator, NodeIdIterator,
 };
 
 mod batch_iterator;
 pub use self::batch_iterator::{BatchIterator, PointCulling, PointLocation, NUM_POINTS_PER_BATCH};
 
+#[cfg(test)]
```
feuerste: Can we remove these three lines?

src/octree/octree_iterator.rs
```diff
@@ -130,3 +130,42 @@ pub fn intersecting_node_ids(
...
+            filter_func,
+        }
+    }
+}
+
+impl<'a> Iterator for NodeIdIterator<'a> {
+    type Item = NodeId;
+
+    fn next(&mut self) -> Option<NodeId> {
+        loop {
```
feuerste: What do you think about shortening this to
```rust
        while let Some(current) = self.node_ids.pop_front() {
            if (self.filter_func)(&current, &self.octree) {
                for child_index in 0..8 {
                    let child_id = current.get_child_id(ChildIndex::from_u8(child_index));
                    if self.octree.nodes.contains_key(&child_id) {
                        self.node_ids.push_back(child_id);
                    }
                }
                return Some(current);
            }
        }
        return None;
```
?

src/octree/octree_iterator.rs
```diff
@@ -130,3 +130,42 @@ pub fn intersecting_node_ids(
     }
     node_ids
 }
+
+pub struct NodeIdIterator<'a> {
+    octree: &'a Octree,
+    filter_func: Box<Fn(&NodeId, &Octree) -> bool + 'a>,
```
nnmm: My first preference would be to integrate this directly with PointCulling instead of functions, and to use the more optimized distinction between Inside, Crossing and Outside to automatically add child nodes. If we use functions, how about making this a template parameter instead of a trait object?

src/octree/mod.rs
catevita: if it is not here, it does not compile the test

src/octree/octree_iterator.rs
catevita: As mentioned offline, the iterator is more generic than the filtering of the point culling. This distinction occurs for Frustum and a specific method returns the node ids.

src/octree/octree_iterator.rs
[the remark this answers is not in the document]
catevita: the loop is necessary to guarantee that next() returns Some(item) if the filter fuction is successful.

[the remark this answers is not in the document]
catevita: nonetheless can be optimized!

catevita: done

src/octree/octree_iterator.rs
```diff
@@ -130,3 +130,42 @@ pub fn intersecting_node_ids(
     }
     node_ids
 }
+
+pub struct NodeIdIterator<'a> {
```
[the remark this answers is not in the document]
catevita: done

catevita:
> Can we actually also apply/use the new `NodeIdIterator` inside `FilteredPointsIterator` and `AllPointsIterator`?

I removed these in https://github.com/googlecartographer/point_cloud_viewer/pull/273
this PR is just implementing the NodeIdsIterator

src/octree/mod.rs
feuerste: Just tried, it also compiles without.

catevita: resuming from offline talk, it compiles, but the test in octree_test.rs can't be run,

src/octree/octree_test.rs
```diff
@@ -0,0 +1,147 @@
...
+
+        let pool = scoped_pool::Pool::new(10);
+        let tmp_dir = TempDir::new("octree").unwrap();
+
+        build_octree(&pool, &tmp_dir, 1.0, bounding_box, points.into_iter());
+        crate::octree::on_disk::octree_from_directory(tmp_dir.into_path()).unwrap()
+    }
+
+    #[test]
+    //#[ignore]
```
feuerste: Nit: Please delete this line.

src/octree/octree_iterator.rs
nnmm: Yes, but can we make this a parameter of the struct, I.e. `struct NodeIdsIterator<F>` and `filter_func: F`, and restrict it to `Fn(&NodeId, &Octree) -> bool` in the impl?

nnmm: Actually, seems like you can directly embed the trait bound into the struct definition (`where F: Fn(&NodeId, &Octree) -> bool`)

src/octree/octree_test.rs
catevita: done!

src/octree/octree_iterator.rs
catevita: done

src/octree/octree_iterator.rs
```diff
@@ -130,3 +130,47 @@ pub fn intersecting_node_ids(
...
+            octree,
+            node_ids: vec![NodeId::from_level_index(0, 0)].into(),
+            filter_func,
+        }
+    }
+}
+
+impl<'a, F> Iterator for NodeIdsIterator<'a, F>
+where
+    F: Fn(&NodeId, &Octree) -> bool + 'a,
```
nnmm: I don't think we need that here when it's already in the struct definition, do we?

nnmm: Same above

src/octree/octree_test.rs
catevita: test streamlined

src/octree/octree_iterator.rs
catevita: Unfortunately it does not work removing that

catevita:
@wally-the-cartographer merge

catevita:
@wally-the-cartographer merge

catevita:
@pifon2a Hi, sorry to bother you, Wally does not seem to be responsive, can you please restart it?

pifon2a:
@catevita The Wally is not run by me anymore. I have also left the project. Merging manually...

pifon2a:
@catevita Wolfgang and Holger know more about that.

catevita:
Thanks, for all your recent and previuos help! next time I am asking to someone else, all the best

PR #269, opened 2019-05-21
Remove unused ecef earth bounding box feature.

feuerste:
PR #240 added the option to fix the bounding box of an octree, which was never used. This PR reverts these changes, which leads to tight bounding boxes for all octrees again, so we have the option to query this bounding box quickly in the point cloud client, see also PR #270.


feuerste:
@wally-the-cartographer merge

wally-the-cartographer:

feuerste:

PR #270, opened 2019-05-21
Refactor xray quadtree building to use point cloud client.

feuerste:
Previously xrays could only be built from one single octree. Using the point cloud client allows us to specify multiple octree locations and define our own factory.

I tested it by running `build_xray_quadtree` with
- no coloring strategy
- `--coloring_strategy colored_with_intensity`
- `--coloring_strategy colored_with_height_stddev`
- `--coloring_strategy colored`

each on the master branch and this branch.
When running `diff -qr xrays/ xrays_new/` on the corresponding output directories, only `meta.pb` differed (because of the arbitrary ordering of repeated messages), all image files were exactly the same.


point_cloud_client/src/lib.rs
```diff
@@ -19,12 +21,27 @@ impl PointCloudClient {
                     .map(|octree| *octree)
             })
             .collect();
+        let mut aabb = Aabb3::zero();
+        if let Ok(octrees) = &octrees {
```
nnmm: How about collapsing these ifs with `if let Ok((first, rest)) = octrees().ok().and_then(|octs| oct.split_first())`?

point_cloud_client/src/lib.rs
```diff
@@ -19,12 +21,27 @@ impl PointCloudClient {
...
+            if let Some(first_octree) = octrees.first() {
+                aabb = *first_octree.bounding_box();
+                for octree in octrees.iter().skip(1) {
```
catevita: why does it skip the first element?

point_cloud_client/src/lib.rs
```diff
@@ -19,12 +21,27 @@ impl PointCloudClient {
...
+                    let octree_bbox = octree.bounding_box();
+                    aabb = aabb.grow(octree_bbox.min()).grow(octree_bbox.max());
```
catevita: I also saw recently that there is the method union() to merge two Aabbs

point_cloud_client/src/lib.rs
nnmm: Actually, can we use `?` in the expression above and here and bail early?

feuerste: I cleaned up the code, PTAL.

point_cloud_client/src/lib.rs
feuerste: Thanks, I modified it.

feuerste:
@wally-the-cartographer merge

wally-the-cartographer:

PR #271, opened 2019-05-21
Xrays with local transforms.

feuerste:
This PR adds the ability to pass a transform to `build_xray_quadtree`, so all processed points are treated as points in a local axis aligned bounding box. This is important, if the octree points are e.g. in ECEF coordinates, since xray generation expects a local, gravity aligned coordinate frame.

### DIFF
--- a/src/math.rs
+++ b/src/math.rs
@@ -13,7 +13,8 @@
 // limitations under the License.
 
 use cgmath::{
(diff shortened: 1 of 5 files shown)


feuerste:
@wally-the-cartographer merge

wally-the-cartographer:

PR #273, opened 2019-05-22
PointCulling as trait 

catevita:
To prepare parallelization across nodes, we split into node search first and then point iteration and filtering. PointCulling has being reformatted into a trait to perform the most common operations.


src/octree/batch_iterator.rs
```diff
@@ -153,14 +136,11 @@ impl<'a> BatchIterator<'a> {
...
-            PointCulling::Any() => Box::new(self.octree.all_points()),
-            PointCulling::Aabb(aabb) => Box::new(self.octree.points_in_box(aabb)),
-            PointCulling::Obb(obb) => Box::new(self.octree.points_in_obb(obb)),
-            PointCulling::Frustum(frustum) => Box::new(self.octree.points_in_frustum(frustum)),
-            PointCulling::OrientedBeam(beam) => Box::new(self.octree.points_in_oriented_beam(beam)),
-        };
-        iterator.try_for_each(|point: Point| point_stream.push_point_and_callback(point))?;
-        point_stream.callback()
+        let mut iterator: Box<Iterator<Item = NodeId>> =
+            get_node_id_iterator(&self.octree, &self.culling);
```
catevita: here the origin of most changes. (still wip)

nnmm: Where does get_node_id_iterator come from? I don't see it in the other PR.

catevita:
finally it compiles and tests run smoothly! this is my first trait implementation so I am already grateful for all your suggestions!
PTAL @nnmm @feuerste

src/math.rs
```diff
@@ -259,19 +289,43 @@ impl<S: BaseFloat> Obb<S> {
...
+        Box::new(Self::new(
+            isometry * &self.isometry_inv.inverse(),
+            self.half_extent,
+        ))
+    }
+}
+
 #[derive(Debug, Clone)]
-pub struct OrientedBeam {
+pub struct OrientedBeam<S> {
```
nnmm: How about making this its own small PR?

nnmm:
At a high level, do you think we could keep the `PointCulling` struct (and name the trait something else)? I think in general not having Boxed up traits is cleaner.

src/math.rs
catevita: I had to change the <f64> in <S> to allow for Sync and Send fot the trait. The other changes except for the trait are  again already in master, here unfortunately yet not visible cause the PR of NodeIdIterator is still open

catevita:
Resuming from the spontaneus pairing session with Nikolai: keeping the point culling enum struct allows to spare costly dereferencing from Arc and Box with the cost of few disambiguation lines https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=b5ad139bdfae3bcc3e93d15911c8ee0a

catevita:
this PR has been splitted up. 
See https://github.com/googlecartographer/point_cloud_viewer/pull/278 (still big)

PR #275, opened 2019-05-24
API cleanup.

feuerste:
This PR cleans up the interface by hiding non-public functions.

### DIFF
--- a/xray/src/generation.rs
+++ b/xray/src/generation.rs
@@ -543,7 +543,7 @@ pub fn build_xray_quadtree(
     let (all_nodes_tx, all_nodes_rx) = mpsc::channel();
     println!("Building level {}.", deepest_level);
 
-    let global_from_local = &local_from_global.clone().map(|t| t.inverse());
+    let global_from_local = &local_from_global.as_ref().map(|t| t.inverse());
     pool.scoped(|scope| {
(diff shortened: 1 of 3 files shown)


xray/src/generation.rs
```diff
@@ -545,6 +543,7 @@ pub fn build_xray_quadtree(
     let (all_nodes_tx, all_nodes_rx) = mpsc::channel();
     println!("Building level {}.", deepest_level);
 
+    let global_from_local = &local_from_global.clone().map(|t| t.inverse());
```
nnmm: Naive question: Why is this clone() needed?

feuerste: We create a reference to a temporary here, without clone this wouldn't be possible.

nnmm: What do you think about using `as_ref()`? Not that performance matters here, but it feels slightly cleaner.

feuerste: Yep, good idea, done.

nnmm:
:100:

feuerste:
@nnmm I think here a :100: unfortunatly isn't enough to approve :disappointed:

feuerste:
@wally-the-cartographer merge

wally-the-cartographer:

PR #278, opened 2019-06-07
PointCulling Trait and Iterator cleanup

catevita:
The iterator are splitted on 
-iterator that retrieve nodes
-iterator that scan points on a single node

The PointLocation in which the Points that have to be retrieved has been streamlined,
the methods contains and intersects are part of the trait PointCulling.
This works for an Aabb3, Obb, OrientedBeam and Frustum


catevita:
still quite a big PR, I tried my best to remove some other additional changes
PTAL @nnmm @feuerste

src/math.rs
```diff
@@ -276,19 +301,43 @@ impl<S: BaseFloat> Obb<S> {
     }
 }
 
+impl<S> PointCulling<S> for Obb<S>
```
nnmm: Nit: Why was this moved?

src/octree/batch_iterator.rs
```diff
@@ -1,31 +1,47 @@
...
 /// size for batch
 pub const NUM_POINTS_PER_BATCH: usize = 500_000;
 
-///possible kind of iterators that can be evaluated in batch of points in BatchIterator
 #[allow(clippy::large_enum_variant)]
-#[derive(Clone)]
-pub enum PointCulling {
-    Any(),
+#[derive(Debug, Clone)]
+pub enum GeoFence {
```
nnmm: I think it might be clearer if we name this PointLocation, and what's currently called PointLocation can become PointQuery. GeoFence has a specific meaning: https://en.wikipedia.org/wiki/Geo-fence

src/octree/batch_iterator.rs
```diff
@@ -151,16 +148,27 @@ impl<'a> BatchIterator<'a> {
...
-        let mut iterator: Box<Iterator<Item = Point>> = match &self.culling {
-            PointCulling::Any() => Box::new(self.octree.all_points()),
-            PointCulling::Aabb(aabb) => Box::new(self.octree.points_in_box(aabb)),
-            PointCulling::Obb(obb) => Box::new(self.octree.points_in_obb(obb)),
-            PointCulling::Frustum(frustum) => Box::new(self.octree.points_in_frustum(frustum)),
-            PointCulling::OrientedBeam(beam) => Box::new(self.octree.points_in_oriented_beam(beam)),
-        };
-        iterator.try_for_each(|point: Point| point_stream.push_point_and_callback(point))?;
-        point_stream.callback()
+        //todo (catevita) mutable function parallelization
```
nnmm: I'm not sure I understand and/or we need this. In the end, the user calls try_for_each_batch. Are you proposing the callbacks should happen on several threads in parallel? I think that's not super useful, since it requires synchronization on the user side again. Better to (as a user) e.g. request points in non-overlapping boxes from several threads.

src/octree/mod.rs
```diff
@@ -302,50 +300,35 @@ impl Octree {
...
+        let container = location.get_point_culling();
+        let filter_func = move |node_id: &NodeId, octree: &Octree| -> bool {
+            let current = &octree.nodes[&node_id];
+            container.intersects(&current.bounding_cube.to_aabb3())
+        };
+        NodeIdsIterator::new(&self, filter_func)
     }
 
-    pub fn points_in_frustum<'a>(
+    /// Returns the ids of all nodes that cut or are fully contained in 'aabb'.
```
nnmm: Nit: Let's be consistent with two vs three slashes

src/math.rs
```diff
@@ -16,11 +16,59 @@ use cgmath::{
...
+use std::convert::From;
+use std::fmt::Debug;
 use std::ops::Mul;
 
+pub trait PointCulling<S>: objekt::Clone + Debug + Sync + Send
+where
+    S: BaseFloat + Sync + Send,
+{
+    fn contains(&self, point: &Point3<S>) -> bool;
+    fn intersects(&self, aabb: &Aabb3<S>) -> bool;
```
nnmm: Could you add a TODO here to return Relation?

src/math.rs
```diff
@@ -16,11 +16,59 @@ use cgmath::{
...
+    fn transform(&self, isometry: &Isometry3<S>) -> Box<PointCulling<S>>;
```
nnmm: This can probably be impl PointCulling, I'd think.

src/octree/mod.rs
```diff
@@ -397,3 +380,54 @@ fn maybe_push_node(
         });
     }
 }
+
+#[derive(Debug, Clone)]
+pub struct Frustum<S: BaseFloat> {
+    pub matrix: Matrix4<S>,
```
nnmm: Could you remind me why this contains the second matrix? I forgot. Also, the fields probably do not need to be public.

src/octree/batch_iterator.rs
catevita: good idea, I was running short on names!

src/octree/batch_iterator.rs
catevita: As told offline, this PR only takes care of implementing the trait, the parallelization follows up next

src/octree/batch_iterator.rs
catevita: done!

src/math.rs
catevita: the methods have been moved to implement the trait - I hope to understand the question

src/octree/mod.rs
catevita: this was meant to be a minimal doc comment for the method points_in_node, do we use doc comment at all?

src/octree/mod.rs
catevita: i moved it in math, the matrix is used by the contains function

// TODO(ksavinash9) update after https://github.com/rustgd/collision-rs/issues/101 is resolved.
pub fn contains<S: BaseFloat>(projection_matrix: &Matrix4<S>, point: &Point3<S>) -> bool {

the issue in collision is not resolved yet

src/math.rs
catevita: done

src/math.rs
catevita: Unfortunately this does not work inside a trait, from [rfcs](https://github.com/rust-lang/rfcs/blob/master/text/1522-conservative-impl-trait.md)

_"impl Trait may only be written within the return type of a freestanding or inherent-impl function, not in trait definitions or any non-return type position. "_

src/math.rs
feuerste: Nit: Our formatting convention for TODOs is:
```rust
// TODO(username): Do something.
```
Can you please adjust it everywhere in this PR?

point_cloud_client/src/bin/test.rs
```diff
@@ -54,8 +53,8 @@ fn main() {
     let num_points = args.num_points;
     let point_cloud_client = PointCloudClient::new(&args.locations, OctreeFactory::new())
         .expect("Couldn't create octree client.");
-    let point_location = PointLocation {
-        culling: PointCulling::Aabb(Aabb3::new(args.min, args.max)),
```
feuerste: Any specific reason to remove the min/max restriction?
Then the whole point of having min/max arguments goes away...

point_viewer_grpc/src/bin/octree_benchmark.rs
```diff
@@ -71,14 +72,20 @@ fn server_benchmark(octree_directory: &Path, num_points: u64) {
...
-            println!("Streamed {}M points", counter / 1_000_000);
-        }
-        counter += 1;
-        if counter == num_points {
+    let all_points = PointQuery {
+        location: PointLocation::AllPoints(),
+        global_from_local: None,
+    };
+    let mut batch_iterator = BatchIterator::new(&octree, &all_points, 1_000_000);
+    let mut num_batches = 0;
```
feuerste: Nit: How about having a const variable `BATCH_SIZE`, removing `num_batches`, and just working on `counter`?

point_viewer_grpc/src/service.rs
```diff
@@ -145,12 +137,11 @@ impl proto_grpc::Octree for OctreeService {
...
-            OctreeQuery::Frustum(frustum_matrix),
-            &req.octree_id,
-            &ctx,
-            resp,
-        )
+        let point_location = PointQuery {
+            location: octree::PointLocation::Frustum(frustum_matrix),
+            global_from_local: None,
+        };
+        self.stream_points_back_to_sink(point_location, &req.octree_id, &ctx, resp)
```
feuerste: It seems `global_from_local` is `None` whenever calling `self.stream_points_back_to_sink`. How about just passing a `location` and constructing the `PointQuery` inside `stream_points_back_to_sink`?

point_viewer_grpc/src/service.rs
```diff
@@ -266,57 +264,67 @@ impl OctreeService {
...
+                            v
+                        })
+                        .collect();
+
+                    reply.colors = match p_data.layers.get(&"color".to_string()) {
+                        Some(LayerData::U8Vec4(data)) => data
+                            .iter()
+                            .map(|p| {
+                                let mut v = point_viewer::proto::Color::new();
+                                v.set_red(f32::from(p.x) / 255.);
```
feuerste: How about using `Color::to_f32` to perform the conversion?

src/math.rs
```diff
@@ -13,14 +13,63 @@
...
+where
+    S: BaseFloat + Sync + Send,
+{
+    fn contains(&self, point: &Point3<S>) -> bool;
+    // todo (catevita) return Relation
+    fn intersects(&self, aabb: &Aabb3<S>) -> bool;
+    fn transform(&self, isometry: &Isometry3<S>) -> Box<PointCulling<S>>;
+}
+
+// PointCulling for Aabb3
```
feuerste: Nit: remove comment, this is clear

src/math.rs
```diff
@@ -150,22 +199,6 @@ impl<'a, S: BaseFloat> Mul<&'a Vector3<S>> for &'a Isometry3<S> {
     }
 }
 
-impl<S: BaseFloat> Into<Decomposed<Vector3<S>, Quaternion<S>>> for Isometry3<S> {
```
feuerste: Any reason why the generics were removed here and instead having an `f64` implementation only?

src/math.rs
```diff
@@ -297,27 +347,12 @@ impl OrientedBeam {
...
-        isometry: &Isometry3<f64>,
-        half_extent: &Vector2<f64>,
-    ) -> [Point3<f64>; 4] {
-        let corner_from = |x: f64, y: f64| {
-            isometry.rotation.rotate_point(Point3::new(x, y, 0.0)) + isometry.translation
+    fn precompute_corners(isometry: &Isometry3<S>, half_extent: &Vector2<S>) -> [Point3<S>; 4] {
+        let corner_from = |x, y| {
+            isometry
+                .rotation
+                .rotate_point(Point3::new(x, y, S::from(0.0).unwrap()))
```
feuerste: Nit: Use `S::zero()` instead of `S:from(0.0)`

src/math.rs
```diff
@@ -428,3 +486,54 @@ mod tests {
...
+    fn intersects(&self, aabb: &Aabb3<S>) -> bool {
+        match self.frustum.contains(aabb) {
+            Relation::Cross => true,
+            Relation::In => true,
+            Relation::Out => false,
+        }
+    }
+
+    fn transform(&self, isometry: &Isometry3<S>) -> Box<PointCulling<S>> {
+        let matrix = Matrix4::from(Decomposed {
```
feuerste: Here we could nicely make use of the conversion methods between `Decomposed` and `Isometry3`, which were deleted above...

src/octree/batch_iterator.rs
```diff
@@ -151,16 +148,27 @@ impl<'a> BatchIterator<'a> {
...
+            .clone()
+            .map(|t| t.inverse());
+        let mut point_stream = PointStream::new(self.batch_size, local_from_global, &mut func);
+        // nodes iterator: retrieve nodes
+        let node_id_iterator = self.octree.nodes_in_location(self.point_location);
+        // operate on nodes
+        for node_id in node_id_iterator {
+            let point_iterator = self.octree.points_in_node(self.point_location, node_id);
+            for point in point_iterator {
+                if let Err(err) = point_stream.push_point_and_callback(point) {
```
feuerste: Can we just do `point_stream.push_point_and_callback(point)?;` instead?

src/math.rs
catevita: this there is indeed some automatic or manual merging gone wrong! I am restoring them

src/math.rs
catevita: done

src/math.rs
catevita: done

src/math.rs
catevita: done

point_viewer_grpc/src/bin/octree_benchmark.rs
catevita: done

point_cloud_client/src/bin/test.rs
catevita: done

catevita: i think it also was a copy-paste error

point_viewer_grpc/src/service.rs
catevita: if it is ok for you i'd like to leave this pieces of code as is, as the grpc server is not intensively used, so if we have to support query with local/global transforms the PointQuery struct is already prepared to support that.

catevita:
I did some benchmarking running ` time (./target/release/octree_benchmark ~/octree_marco/ --no-client)` on my workstation. Results are quite aligned, typically these

new:
`real	0m3.662s
user	0m3.500s
sys	0m0.160s`

old
`real	0m3.563s
user	0m3.408s
sys	0m0.152s`

src/math.rs
catevita: done

src/octree/batch_iterator.rs
catevita: done!

point_viewer_grpc/src/bin/octree_benchmark.rs
catevita: done

point_viewer_grpc/src/service.rs
catevita: I changed it, it adds some lines but the performance seems not to be affected from that

feuerste:
Can you please remove the `.DS_Store` file?

src/math.rs
```diff
@@ -13,14 +13,61 @@
...
+{
+    fn contains(&self, point: &Point3<S>) -> bool;
+    // TODO (catevita) return Relation
+    fn intersects(&self, aabb: &Aabb3<S>) -> bool;
+    fn transform(&self, isometry: &Isometry3<S>) -> Box<PointCulling<S>>;
+}
+
+impl<S> PointCulling<S> for Aabb3<S>
+where
+    S: 'static + BaseFloat + Sync + Send,
```
feuerste: What do we `'static` need for?

src/math.rs
feuerste: Supernit: Still not in the format
```
// TODO(username): Do something.
```
everywhere.

src/math.rs
feuerste: It would be clearer to move them back exactly where they were before, so they don't show up as diff...

src/math.rs
catevita: My bad, I just paid attention to the caps! now it should be finally fixed!

catevita:
> Can you please remove the `.DS_Store` file?

Indeed it shows up in the list, but it as removed file

feuerste:
> > Can you please remove the `.DS_Store` file?
> 
> Indeed it shows up in the list, but it as removed file

Ah, I see, thanks, I overlooked this...

catevita:
@wally-the-cartographer merge

catevita:
@wally-the-cartographer merge

feuerste:
@wally-the-cartographer merge

catevita:
@gaschler sorry to bother you, Wally seems to be down again, could you help me merging this branch or pointing to the right person that mantains Wally?

catevita:
thanks!

PR #219, opened 2019-02-22
octree dir: split content into smaller files

catevita:
ptal! @nnmm 
ptal! @feuerste 

in this small PR I splitted the node and mod files of the src/octree into pieces to allow for a better overview.

PR #279, opened 2019-06-07
Splitted batch iterator functionality

catevita:
splitted batch iterator functionality into 
- getting node_ids per octree according to a function e.g. point location
- getting from each node chunks of data
- computing the mut function at the end

this is the pre step that allows finally for a parallelization of the query

### DIFF
--- a/src/octree/batch_iterator.rs
+++ b/src/octree/batch_iterator.rs
@@ -1,11 +1,12 @@
-use crate::errors::*;
+use crate::errors::Result;
 use crate::math::PointCulling;
 use crate::math::{AllPoints, Isometry3, Obb, OrientedBeam};
-use crate::octree::{self, Octree};
+use crate::octree::{self, FilteredPointsIterator};
 use crate::{LayerData, Point, PointData};
 use cgmath::{Matrix4, Vector3, Vector4};
 use collision::Aabb3;
 use fnv::FnvHashMap;
+use std::collections::VecDeque;
 
 /// size for batch
 pub const NUM_POINTS_PER_BATCH: usize = 500_000;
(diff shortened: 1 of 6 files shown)


src/octree/batch_iterator.rs
```diff
@@ -107,68 +104,74 @@ where
...
-    pub fn new(
-        octree: &'a octree::Octree,
-        point_location: &'a PointLocation,
-        batch_size: usize,
-    ) -> Self {
-        BatchIterator {
-            octree,
-            point_location,
-            batch_size,
+    mut func: Callback,
```
catevita: this function will be splitted in the next pr:
1) gathering node ids in the octrees
2)parallelizing point iterator over nodes returning batches of points
3) applying the mut function in the receiver/consumer thread

catevita:
ptal @feuerste @nnmm

catevita:
ptal @feuerste @nnmm

catevita:
a number of changes of the pr in #219 got lost in the merge to master
hence i'd leave for comparison the changes against the branch and then switch after the review to master

src/octree/batch_iterator.rs
```diff
@@ -107,66 +103,71 @@ where
...
+}
 
-        Ok(())
-    }
+pub fn get_point_data_iterator<'a>(
+    octree: &'a octree::Octree,
+    point_query: &'a PointQuery,
+    batch_size: usize,
+) -> VecDeque<Box<BatchIterator<impl Fn(&Point) -> bool>>> {
+    let local_from_global = point_query.global_from_local.clone().map(|t| t.inverse());
```
nnmm: Nit: I think you can avoid the clone with `as_ref()`

nnmm:
The main question I have is whether we want an mpsc of points vs an mpsc of batches. This PR seems to be targetting an mpsc of batches, and I'd like to understand better why, since the decision to go with batches options seems to be making this change a bit complex. Is it for performance reasons?

catevita:
yes, I'd say so. every time you want to read a single point a piece of file is gonna be cached and you get the rest of the batch nearly for free. the batch iterator has been now implemented since a couple of prs. which complexity are you thinking of? This PR actually decouples the evaluation of mutable fuctions from the reading of the points, as mutable function do not generally go along with threads.

nnmm:
I think you're right that it makes sense to have an mpsc/channel of batches. It seems possible however to keep the BatchIterator and PointStream like before, but just make the callback in the PointStream a closure that sends the batch over the channel back to the main thread. The actual callback could then be executed with `try_for_each` on `rx.iter()` like in [this demo](https://github.com/googlecartographer/point_cloud_viewer/compare/master...nnmm:nnmm/mpsc_points). That seems a lot simpler, but if I'm missing anything, let me know :)

catevita:
in theory yes, it would work, but it allows misuse of the batch iterator. that's why I want to scrap passing the mut function completely.

PR #280, opened 2019-06-12
Make protobuf code generation not depend on system installation

nnmm:
Instead, protoc is expected to be contained within `target/protobuf/bin/protoc`. This eliminates confusion and brittleness around PATH ordering.

A follow-up PR will factor out the boilerplate of generating protobufs (setting the PATH to point to the above directory, wrapping in a `mod proto` etc.


nnmm:
meow @tblaette 
fyi @feuerste

README.md
```diff
@@ -10,7 +10,7 @@ It was build to serve the needs of the [Cartographer](https://github.com/googlec
 The project consist of a root crate that can build and read octrees on disk and viewer binaries to visualize the data. For Mac OS X, we assume below that you've installed [Homebrew](https://brew.sh).
 
 - Install Rust: `curl https://sh.rustup.rs -sSf | sh`. See <https://rustup.rs> for details.
-- Install protobuf > 3.0. See [ci/install_proto3.sh](https://github.com/googlecartographer/point_cloud_viewer/blob/master/ci/install_proto3.sh) for Linux or run `brew install protobuf` on Mac OS X.
+- Install `protoc` > 3.0 in `target/protobuf/bin`. The [`protobuf` Makefile target](https://github.com/googlecartographer/point_cloud_viewer/blob/master/Makefile) will compile `protoc` from source and place it there. Alternatively, you can symlink to a `protoc` installed on your system, e.g. from `brew install protobuf` on Mac OS X.
```
tblaette: make the separate steps for Linux vs. macOS clearer. Something along these lines:
- Linux: Install protoc by running the makefile
- macOS: brew install protobuf, symlink from ... to ...

Makefile
```diff
@@ -0,0 +1,20 @@
...
+ROOT_DIR := $(abspath $(dir ${MAKEFILE_PATH}))
+PROTOBUF_DIR := ${ROOT_DIR}/target/protobuf
+
+.PHONY: protobuf
+protobuf: ${PROTOBUF_DIR}/protoc
+
+${PROTOBUF_DIR}/protoc:
+	mkdir -p ${PROTOBUF_DIR}
+	$(eval TMPDIR := $(shell mktemp -d))
+	git clone https://github.com/google/protobuf.git ${TMPDIR}/protobuf
```
tblaette: check out a specific version of protobuf and not the latest

README.md
nnmm: Do you agree with the current formulation? The script works for both Mac and Linux.

Makefile
nnmm: Done. Checking out 3.6.1.

nnmm:
@wally-the-cartographer merge

point_viewer_grpc_proto_rust/build.rs
```diff
@@ -48,12 +48,15 @@ fn main() {
     println!("cargo:rerun-if-changed=src/proto.proto");
 
     let git_repo_root = find_git_repo_root();
+    let old_path = env::var("PATH");
```
feuerste: Would it make sense to have a small RAII helper for temporarily changing the path and use it everywhere, something like?
```rust
struct TempPath {
    old_path: String,
}

impl TempPath {
    pub fn new(new_path: &Str) -> Self {
        let old_path = env::var("PATH");
        env::set_var("PATH", new_path);
        TempPath{ old_path }
    }
}

impl Drop for TempPath {
    fn drop(&mut self) {
        env::set_var("PATH", self.old_path.unwrap_or("".to_string()));
    }
}
```

nnmm: Something like that, yes. I'd do that in a separate `build_proto` crate that we can use in build scripts for other crates.

feuerste: Cool, thanks

nnmm:
@cschuet Could you merge this PR?

nnmm:
Alternatively @gaschler?

nnmm:
@sebastianklose Could you merge and/or point me to the right person to ping about Wally?

nnmm:
Thank you, Andre!

PR #281, opened 2019-06-13
Expose ecef to s2 cell id conversions.

feuerste:
This will be needed in order to build point clouds sharded by s2 cells.


Cargo.toml
```diff
@@ -24,14 +24,16 @@ cgmath = "0.16.0"
 collision = "0.18.0"
 error-chain = "0.11.0"
 fnv = "1.0.6"
+libc = "0.2.43"
+nav-types = "0.3"
 num = "0.1.36"
 num-traits = "0.1.36"
 pbr = "1.0.0-alpha.1"
 protobuf = "2.0.0"
+s2 = "0.0.8"
```
nnmm: Could it be a bit confusing to have a dependency of the same name as a module that uses it?

feuerste: Probably... how about `s2_geo` or `geo`?

nnmm: `s2_geo` sounds good

feuerste: Done, renamed it.

src/s2_geo.rs
```diff
@@ -0,0 +1,120 @@
...
+
+pub fn cell_id(ecef: ECEF<f64>, level: u8) -> CellID {
+    CellID::from(lat_lng_from_ecef(ecef)).parent(u64::from(level))
+}
+
+pub fn cell_ids_rect(
+    ecef_m: ECEF<f64>,
+    extent_m: Vector2<f64>,
+    level: u8,
+    max_cells: usize,
```
nnmm: Nit: max_num_cells?

src/s2_geo.rs
```diff
@@ -0,0 +1,120 @@
...
+const LEVEL_MOD: u8 = 1;
+const EARTH_RADIUS_M: f64 = 6_371_010.0;
+
+fn earth_tangent_m_to_rad(meters: f64) -> f64 {
+    2.0 * (0.5 * meters / EARTH_RADIUS_M).atan()
+}
+
+fn lat_lng_from_ecef(ecef: ECEF<f64>) -> LatLng {
+    let wgs = WGS84::from(ecef);
+    let lat = Angle::from(Rad(wgs.latitude()));
```
nnmm: Nit: Do you think it would be helpful to have WGS LatLng to S2 LatLng conversion be a function?

src/s2_geo.rs
```diff
@@ -0,0 +1,120 @@
...
+mod tests {
+    use super::*;
+    use s2::cellid::MAX_LEVEL;
+
+    #[test]
+    fn test_cell_id() {
+        // Hofbräuhaus Munich.
+        let ecef = ECEF::new(4_177_536., 855_997., 4_727_102.);
+        assert_eq!(
+            "479e758bb8ddeb63",
```
nnmm: It was a bit confusing to me that MAX_LEVEL is much more exact than your ECEF example coordinates – maybe test a lower level?

src/s2_geo.rs
```diff
@@ -0,0 +1,120 @@
...
+}
+
+pub fn cell_ids_radius(
+    ecef_m: ECEF<f64>,
+    radius_m: f64,
+    level: u8,
+    max_cells: usize,
+) -> Vec<CellID> {
+    let wgs = WGS84::from(ecef_m);
+    let center = LatLng::new(
```
nnmm: If this is to be used to implement queries, we can't simply drop the altitude, no?

src/s2_geo.rs
nnmm: An alternative would be to have a `ToS2LatLng` trait and call its conversion function here

src/s2_geo.rs
```diff
@@ -0,0 +1,120 @@
...
+use s2::rect::Rect;
+use s2::region::RegionCoverer;
+use s2::s1::angle::{Angle, Rad};
+use s2::s1::chordangle::ChordAngle;
+use std::vec::Vec;
+
+const LEVEL_MOD: u8 = 1;
+const EARTH_RADIUS_M: f64 = 6_371_010.0;
+
+fn earth_tangent_m_to_rad(meters: f64) -> f64 {
```
nnmm: I'd be wary of pretending the earth is a sphere. For `cell_ids_radius()` it might be good to have that, but then we should be sure that it overestimates the rad, and it could maybe be made internal to that function.

src/s2_geo.rs
nnmm: It was a bit confusing to me that MAX_LEVEL is much more exact than your ECEF example coordinates – maybe test a lower level?

src/s2_geo.rs
```diff
@@ -0,0 +1,120 @@
...
+    }
+
+    #[test]
+    fn test_region_coverer() {
+        // Defines a 5x5 m region in Palo Alto.
+        let length = 5.0;
+        let ecef_pao = ECEF::new(-2699112.642037565, -4294985.7005343605, 3853368.5643771216);
+        let extent = Vector2::new(length, length);
+
+        let tokens_ground_truth = vec!["808fba97497", "808fba97499"];
```
nnmm: Does that come from the online region coverer app?

src/s2_geo.rs
feuerste: This is what the conversion between WGS84 and LatLng internally does...

src/s2_geo.rs
feuerste: I like the idea of a trait, implemented it...

src/s2_geo.rs
feuerste: Done

src/s2_geo.rs
feuerste: ECEF coordinates are double floating point... this should be precise enough, right?

src/s2_geo.rs
feuerste: Yes, I did some testing using https://s2.sidewalklabs.com/regioncoverer

src/s2_geo.rs
feuerste: Yes, when actually using it we should have a few more tests... I just wanted to get this in now and concentrate on more important things for now...

nnmm: How about you make it a nested function of `cell_ids_radius()` and optionally add a TODO? Also, I think we get the most `rad` for our `m` at the smallest radius, so let's go with the Arctic Ocean floor for the `EARTH_RADIUS_M` (https://en.wikipedia.org/wiki/Earth_radius#Geophysical_extremes)

nnmm: Best to also hide that constant in `cell_ids_radius()`

nnmm: Oh, I see it's also used in `cell_ids_rect()`. Maybe the best approach is then to suffix these three functions with `_inexact` or `_approx` and add a TODO.

src/s2_geo.rs
nnmm: Right. Let's leave it like this.

src/s2_geo.rs
nnmm: At high altitude the number of cells intersecting a fixed-size sphere will be smaller than at low altitudes. Since it's inexact anyway, we can just call this out in a comment for now.

src/s2_geo.rs
feuerste: How about using the polar radius as defined by WGS84, see https://en.wikipedia.org/wiki/Earth_radius#Global_average_radii and https://en.wikipedia.org/wiki/World_Geodetic_System#WGS84?

feuerste: I actually just changed it to exact computations, PTAL!

nnmm:
:+1: After changes discussed offline

feuerste:

feuerste:
@wally-the-cartographer merge

wally-the-cartographer:

PR #282, opened 2019-06-17
Factor out a proto_helper crate to generate proto stubs

nnmm:


nnmm:
meow @krzesi-mir

nnmm:
fyi @tblaette @feuerste @catevita

proto_helper/src/lib.rs
```diff
@@ -0,0 +1,163 @@
+use std::fs::File;
+use std::io::{Read, Write};
+use std::path::{Path, PathBuf};
+
+pub struct ProtoBuilder {
+    protoc_search_path: Option<PathBuf>,
+    files: Vec<PathBuf>,
+    import_paths: Vec<PathBuf>,
+    generate_grpc: bool,
+    transformations: Vec<Box<Fn(String) -> String>>,
```
krzesi-mir: I would add a comment explaining what the `transformations` fields are for and also specify the order in which the transformations are applied later. i.e. if `transformations = [T1, T2]` and we later apply them to `code`, would it result in `T1(T2(code))` or `T2(T1(code))`?

proto_helper/src/lib.rs
```diff
@@ -0,0 +1,163 @@
...
+
+impl ProtoBuilder {
+    // The path can be relative to the build script.
+    pub fn new() -> Self {
+        Self {
+            protoc_search_path: None,
+            files: vec![],
+            import_paths: vec![],
+            generate_grpc: false,
+            transformations: vec![],
```
krzesi-mir: shall we add `replace_clippy` to `transformations` and `grpc_transformations` here, so we don't have to call `replace_clippy` explicitly later below?

proto_helper/src/lib.rs
```diff
@@ -0,0 +1,163 @@
...
+        });
+    } else {
+        let args = protoc_rust::Args {
+            out_dir: out_dir.to_str().unwrap(),
+            includes: includes.as_slice(),
+            input: &[protobuf_file.to_str().unwrap()],
+            ..Default::default()
+        };
+        protoc_rust::run(args).expect("Error running protoc");
+        let out_path = out_dir.join(format!("{}.rs", protobuf_name.display()));
```
krzesi-mir: see the comment above

proto_helper/src/lib.rs
```diff
@@ -0,0 +1,163 @@
...
+        self.import_paths.push(import_path.as_ref().to_owned());
+        self
+    }
+
+    pub fn grpc(&mut self, generate_grpc: bool) -> &mut Self {
+        self.generate_grpc = generate_grpc;
+        self
+    }
+
+    pub fn add_transformation(&mut self, f: Box<Fn(String) -> String>) -> &mut Self {
```
krzesi-mir: nit: I think it would be easier to use this and next methods if instead of taking a boxed function as an argument, we accept a generic function as an argument and box it on our own, i.e something like.
```
    pub fn add_transformation<F: Fn(String) -> String> (&mut self, f:  F) -> &mut Self {
        self.transformations.push(Box::new(f));
        self
    }
```

proto_helper/src/lib.rs
```diff
@@ -0,0 +1,163 @@
...
+        let out_path = out_dir.join(format!("{}.rs", protobuf_name.display()));
+        inplace_modify_file(&out_path, |code| {
+            let code = replace_clippy(code);
+            transformations.iter().fold(code, |code, tr| tr(code))
+        });
+        let grpc_out_path = out_dir.join(format!("{}_grpc.rs", protobuf_name.display()));
+        inplace_modify_file(&grpc_out_path, |code| {
+            let code = replace_clippy(code);
+            grpc_transformations.iter().fold(code, |code, tr| tr(code))
+
```
krzesi-mir: nit: empty line

proto_helper/src/lib.rs
```diff
@@ -0,0 +1,163 @@
...
+        self
+    }
+
+    pub fn run(&mut self) {
+        let old_path = std::env::var("PATH");
+        if let Some(search_path) = &self.protoc_search_path {
+            std::env::set_var("PATH", search_path);
+        }
+        for imp in self.import_paths.iter_mut() {
+            std::mem::replace(imp, std::fs::canonicalize(&imp).unwrap());
```
krzesi-mir: nit: `std::fs::canonicalize(&imp)` -> `imp.canonicalize()`

proto_helper/src/lib.rs
```diff
@@ -0,0 +1,163 @@
...
+    }
+
+    // Register a file for compilation. The path must be relative to the build script.
+    pub fn add_file<P: AsRef<Path>>(&mut self, file_path: P) -> &mut Self {
+        self.files.push(file_path.as_ref().to_owned());
+        self
+    }
+
+    pub fn add_import_path<P: AsRef<Path>>(&mut self, import_path: P) -> &mut Self {
+        self.import_paths.push(import_path.as_ref().to_owned());
```
krzesi-mir: nit: would it make sense to call `canonicalize` on `import_path` here instead of in the line 64?

proto_helper/src/lib.rs
```diff
@@ -0,0 +1,163 @@
...
+
+    pub fn run(&mut self) {
+        let old_path = std::env::var("PATH");
+        if let Some(search_path) = &self.protoc_search_path {
+            std::env::set_var("PATH", search_path);
+        }
+        for imp in self.import_paths.iter_mut() {
+            std::mem::replace(imp, std::fs::canonicalize(&imp).unwrap());
+        }
+        self.files.iter().for_each(|file| {
```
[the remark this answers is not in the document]
nnmm: Yep

proto_helper/src/lib.rs
```diff
@@ -0,0 +1,163 @@
...
+// Opens a file, calls 'func' with its contents and writes the new content back.
+fn inplace_modify_file<F: FnOnce(String) -> String>(path: &Path, func: F) {
+    let mut contents = String::new();
+    File::open(&path)
+        .expect(&format!("Could not open {}", path.display()))
+        .read_to_string(&mut contents)
+        .unwrap();
+    let new_contents = func(contents);
+    File::create(&path)
+        .unwrap()
```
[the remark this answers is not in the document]
nnmm: Clippy forces me to do `unwrap_or_else`, but why not

proto_helper/src/lib.rs
nnmm: Sure, then run can take self as non-mut

proto_helper/src/lib.rs
nnmm: Neat

proto_helper/src/lib.rs
```diff
@@ -0,0 +1,163 @@
...
+    import_paths: Vec<PathBuf>,
+    generate_grpc: bool,
+    transformations: Vec<Box<Fn(String) -> String>>,
+    grpc_transformations: Vec<Box<Fn(String) -> String>>,
+}
+
+impl ProtoBuilder {
+    // The path can be relative to the build script.
+    pub fn new() -> Self {
+        Self {
```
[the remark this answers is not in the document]
nnmm: Done

proto_helper/src/lib.rs
```diff
@@ -0,0 +1,163 @@
...
+            files: vec![],
+            import_paths: vec![],
+            generate_grpc: false,
+            transformations: vec![],
+            grpc_transformations: vec![],
+        }
+    }
+
+    // The path to the directory where protoc is located. If this is not set, PATH is used.
+    pub fn protoc_search_path<P: AsRef<Path>>(&mut self, protoc_search_path: P) -> &mut Self {
```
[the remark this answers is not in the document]
nnmm: Wouldn't that be less convenient for a user who has a &str?

proto_helper/src/lib.rs
nnmm: F would need to be 'static, which seems a bit of an unpleasant restriction.

proto_helper/src/lib.rs
nnmm: Good idea

proto_helper/src/lib.rs
```diff
@@ -0,0 +1,163 @@
...
+    // protoc-grpcio (in the protoc-grpcio repo)
+
+    let protobuf_name: &Path = protobuf_file.file_stem().unwrap().as_ref();
+    let includes: Vec<&str> = import_paths.iter().map(|p| p.to_str().unwrap()).collect();
+
+    if generate_grpc {
+        // This generates both protobuf and grpc definitions
+        protoc_grpcio::compile_grpc_protos(&[protobuf_file], includes.as_slice(), out_dir)
+            .expect("Error running protoc (with grpc)");
+        let out_path = out_dir.join(format!("{}.rs", protobuf_name.display()));
```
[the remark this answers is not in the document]
nnmm: Yeah, but I wanted to be consistent with `grpc_out_path`. Do you don't think that consistency buys us anything, I can change it.

proto_helper/src/lib.rs
nnmm: Sure, but then we can't do the `Default` thing, can we? I think it's clearer this way anyway.

nnmm: I implemented Default myself

proto_helper/src/lib.rs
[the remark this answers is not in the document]
krzesi-mir: hmm, I think you are right.

proto_helper/src/lib.rs
krzesi-mir: :+1:

proto_helper/src/lib.rs
krzesi-mir: ah, yes, I didn't take this into account.

proto_helper/src/lib.rs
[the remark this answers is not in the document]
krzesi-mir: It's a very small difference so either way works for me, up to you.

catevita:
I could not find other points than those Kamil pointed out, just ping if you need some review after your changes

catevita:
@nnmm would this PR be relevant to the current work with protos?

nnmm:
No, I don't think you can use this to generate Python stubs.

PR #283, opened 2019-06-25
Parallelization of batch iterator for a single octree

catevita:


catevita:
@ptal nnmm

catevita:
PTAL @nnmm

nnmm:
I'd have only minor comments, but do you think we should merge this before implementing the worker pool for limiting the number of threads?

src/octree/batch_iterator.rs
```diff
@@ -111,28 +111,38 @@ where
...
         };
+        //println!("internal callback sending last batch...");
         (self.func)(point_data)
     }
 
-    fn push_point_and_callback(&mut self, point: Point) -> Result<()> {
-        self.push_point(point);
-        if self.position.len() == self.position.capacity() {
-            return self.callback();
+    fn push_point_and_callback<F2>(
```
nnmm: Could we rename the function to `batch_points()` or `push_points_and_callback()` since it no longer just consumes single points?

src/octree/batch_iterator.rs
```diff
@@ -111,28 +111,38 @@ where
...
+        &mut self,
+        point_iterator: FilteredPointsIterator<F2>,
+    ) -> Result<()>
+    where
+        F2: Fn(&Point) -> bool,
```
nnmm: Nit: Couldn't `F2` be `F` as well?  Or perhaps `Filter`?

src/octree/batch_iterator.rs
```diff
@@ -144,29 +154,46 @@ impl<'a> BatchIterator<'a> {
...
     {
-        //TODO(catevita): mutable function parallelization
+        let octree = self.octree;
+        // nodes iterator: retrieve nodes
+        let node_id_iterator = octree.nodes_in_location(self.point_location);
+
         let local_from_global = self
             .point_location
             .global_from_local
             .clone()
```
nnmm: Nit: as_ref instead of clone

src/octree/batch_iterator.rs
```diff
@@ -144,29 +154,46 @@ impl<'a> BatchIterator<'a> {
...
-        for node_id in node_id_iterator {
-            let point_iterator = self.octree.points_in_node(self.point_location, node_id);
-            for point in point_iterator {
-                point_stream.push_point_and_callback(point)?;
+        // operate on nodes: one thread for each node
+        crossbeam::scope(|s| {
+            let (tx, rx) = crossbeam::channel::bounded::<PointData>(2);
+            for node_id in node_id_iterator {
+                let tx = tx.clone();
+                let local_from_global = local_from_global.clone();
```
nnmm: Nit: We could just use a reference to lo local_from_global and self.point_location and avoid the clone.

src/octree/batch_iterator.rs
```diff
@@ -144,29 +154,46 @@ impl<'a> BatchIterator<'a> {
...
+                });
             }
-        }
-        // TODO(catevita): return point data through mpsc channel
-        // TODO(catevita): apply mut function to received data
 
-        Ok(())
+            rx.iter().try_for_each(func)
+        })
+        .expect("child thread panic")
```
nnmm: Nit: How about "Panic in try_for_each_batch child thread"?

src/octree/octree_test.rs
```diff
@@ -138,8 +145,8 @@ mod tests {
             .try_for_each_batch(callback_func)
             .expect_err("Test error");
 
-        assert_eq!(3, callback_count);
-        assert_eq!(3 * batch_size, delivered_points);
+        assert!(callback_count >= 3);
```
nnmm: I have an idea how we can avoid sending half-finished batches (but that can be a future PR): keeping a `mut Option<PointData>` in the receiver thread that can store incomplete batches. When you get an incomplete batch, append it to that PointData, and if the size of that PointData becomes >= batch_size, split off a "full" batch (i.e. of size batch_size), call the callback on that, and set the incomplete batch to the rest of the PointData.

src/octree/batch_iterator.rs
nnmm: Actually, if we are changing the signature anyway, what do you think about generalizing this to return a proper `Result`? Then we wouldn't need our own error type, which is not really an error anyway :)

catevita:
> I'd have only minor comments, but do you think we should merge this before implementing the worker pool for limiting the number of threads?

yes, there are some minor changes to the master that did not get merged with the previous branch, they need to be in the master before some other pr pass

src/octree/batch_iterator.rs
catevita: Filter sounds good!

src/octree/batch_iterator.rs
catevita: ok!

catevita: a reference does not work for the isometry though..

src/octree/octree_test.rs
catevita: sounds good!

src/octree/batch_iterator.rs
catevita: i added also BatchIterator in front of your sentence

src/octree/batch_iterator.rs

src/octree/batch_iterator.rs
catevita: as discussed offline, pointstream is a private struct, returning a proper result can be posponed when needed

catevita:
@wally-the-cartographer merge

catevita:
@gaschler Hi, Wally seems to be down again, can you please merge my changes? Thanks!!

catevita:
@wally-the-cartographer merge

catevita:
@cschuet hi, sorry to bother you, Wally seems to be down, could you help me to merge this changes?

catevita:
thanks!

PR #284, opened 2019-06-26
Prepare node writer/iterator for individual nodes independent of octree.

feuerste:
This makes the node writer/iterator independent from encoding and octrees. Rescaling/encoding points is now optional. With this change we should be able to directly use the node writer/iterator for individual nodes such as s2 cells on a specific level.


nnmm:
Nit: Leaf cells ;)

nnmm:
What makes it specific to leaf cells/nodes?

feuerste:
> What makes it specific to leaf cells/nodes?

Nothing, I updated/clarified the description.

src/node_writer.rs
```diff
@@ -0,0 +1,242 @@
...
+                .unwrap(),
+            intensity_writer: None, // Will be created if needed on first point with intensities.
+            stem,
+            data_writer_builder,
+            position_rescaling,
+        }
+    }
+
+    pub fn write(&mut self, p: &Point) {
+        // TODO(feuerste): This is expensive. It would be preferable not to branch on every point.
```
catevita: could this function be operating on PointData? (does not need to be implemented on this PR

src/node_writer.rs
```diff
@@ -0,0 +1,242 @@
...
+                    .write_f64::<LittleEndian>(p.position.z)
+                    .unwrap();
+            }
+        }
+
+        self.rgb_writer.write_u8(p.color.red).unwrap();
+        self.rgb_writer.write_u8(p.color.green).unwrap();
+        self.rgb_writer.write_u8(p.color.blue).unwrap();
+
+        // TODO(sirver): This is expensive. It would be preferable if we needn't branch on
```
catevita: nit: this todo is very similar to the one on top

src/node_writer.rs
feuerste: Yes, this could definitely go into another PR, if it makes sense...

feuerste:
ptal @nnmm @catevita

nnmm:
AFAIU there are two kinds of changes here: changes motivated by separating the NodeWriter from the octree, and changes that are motivated by using any Writer. Do you think those are interrelated or could the latter (which seems less urgent) be split off into another PR?

I'm also thinking about what an approach that wouldn't use as many factories/builders would look like.

feuerste:
> AFAIU there are two kinds of changes here: changes motivated by separating the NodeWriter from the octree, and changes that are motivated by using any Writer. Do you think those are interrelated or could the latter (which seems less urgent) be split off into another PR?
> 
> I'm also thinking about what an approach that wouldn't use as many factories/builders would look like.

This PR also has the nice effect that now each `DataWriter` takes care of automatically deleting empty files, previously this was done by the `NodeWriter` itself. Of course I could also hard-code this part to disk writing again, if you prefer.

feuerste:
> I'm also thinking about what an approach that wouldn't use as many factories/builders would look like.

We may use a Singleton in combination with builders, so factories wouldn't need to be passed around...

feuerste:
ptal @nnmm
I removed all factories to have less changes in this PR

src/generation.rs
```diff
@@ -34,6 +38,19 @@ use std::sync::mpsc;
 const UPDATE_COUNT: i64 = 100_000;
 const MAX_POINTS_PER_NODE: i64 = 100_000;
 
+impl CubeNodeWriter {
+    pub fn from_data_provider(
+        octree_data_provider: &OnDiskOctreeDataProvider,
+        octree_meta: &OctreeMeta,
+        node_id: &NodeId,
+    ) -> Self {
+        let path = octree_data_provider.stem(node_id);
```
nnmm: Shouldn't we factor that out?

feuerste: What exactly? `generation.rs` belongs to octree specific stuff (which will be moved into the `octree` subfolder in my next PR), so I think this can stay here...

src/read_write/node_iterator.rs
```diff
@@ -2,178 +2,192 @@ use crate::codec::{decode, fixpoint_decode};
...
-                bounding_cube,
-                position_encoding,
-                num_points,
-            },
-            point_count: 0,
-        };
-        Ok(NodeIterator::WithData(iter))
-    }
+pub struct CubeNodeReader {
+    readers: Vec<BufReader<Box<dyn Read>>>,
```
nnmm: Maybe the xyz reader should be a separate member. That would enforce that the position is always available and make the code a little more readable.

src/read_write/node_iterator.rs
```diff
@@ -2,178 +2,192 @@ use crate::codec::{decode, fixpoint_decode};
...
+            readers,
+            position_encoding,
+            bounding_cube,
+        })
+    }
+}
+
+/// Streams points from our data provider representation.
+pub struct NodeIteratorWithData {
+    reader: Box<dyn NodeReader>,
```
nnmm: I'd make this a generic to avoid a dereference each time we read a point. Alternatively, we could pack this into the NodeReader trait (as two getter functions, which would hopefully get inlined), which would avoid an extra struct.

nnmm: With the latter option, we could also make the interface to NodeReader safer and have it return an Option<Point>, right?

src/generation.rs
nnmm: I just felt like it's a pretty "large" parameter, not in a performance sense, but conceptually. For instance, it could make testing a little easier to be able to just instantiate a CubeNodeWriter with some temp file.

nnmm: On the other hand, I don't know OnDiskOctreeDataProvider very well – if my suggestion would lead to lots of redundancy, maybe it's better to leave the file name generation here.

nnmm: It would also sort of decouple the concept of a Cube with points from an octree & its naming scheme.

feuerste: As discussed offline, we will just keep it as it is, as there still is the `new` method. I made this one private to clarify its use.

src/read_write/node_iterator.rs
feuerste: I like the suggestions, implemented all of them, PTAL!

feuerste:
@wally-the-cartographer merge

wally-the-cartographer:

PR #286, opened 2019-06-27
sdl viewer readme

catevita:
just a few lines of docs to document recent user issues


catevita:
PTAL @feuerste

feuerste:
I think we should either keep all information in the top level README.md, since there already is quite some information about the sdl viewer, or move everything to this one

catevita:
I think #287 made it obsolete

PR #291, opened 2019-07-03
Add ply and raw node writers.

feuerste:
This PR prepares for writing raw nodes, which could be e.g. s2 cells, and ply nodes, which can be used e.g. for debugging.
I think more ideas from e.g. https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=f373b5ffbdc37377a7a5caad663ad9cd such as a point trait would need a separate PR.

**Timings:**
_master:_
real	60m5.493s
user	216m5.828s
sys	14m37.088s

_this PR:_
real	54m28.890s
user	176m5.276s
sys	14m53.144s


src/read_write/node_writer.rs
```diff
@@ -164,3 +196,137 @@ impl CubeNodeWriter {
...
+
+        self.point_count += 1;
+    }
+}
+
+impl Drop for PlyNodeWriter {
+    fn drop(&mut self) {
+        self.writer.write_all(b"\n").unwrap();
+        if self.writer.seek(SeekFrom::Start(4 + 32 + 15)).is_ok() {
+            let _res = write!(&mut self.writer, "{:020}", self.point_count);
```
nnmm: You didn't have the off-by-one error here?

src/read_write/node_writer.rs
```diff
@@ -164,3 +196,137 @@ impl CubeNodeWriter {
...
+        {
+            let mut ply_writer = PlyNodeWriter::new(&file_path_test);
+            PlyIterator::from_file(file_path_gt).unwrap().for_each(|p| {
+                ply_writer.write(&p);
+            });
+        }
+        let ply_gt = PlyIterator::from_file(file_path_gt).unwrap();
+        let ply_test = PlyIterator::from_file(&file_path_test).unwrap();
+        ply_gt.zip(ply_test).for_each(|(gt, test)| {
+            assert_eq!(gt.position, test.position);
```
nnmm: Awesome :)

src/read_write/node_writer.rs
feuerste: I fixed it, can show you offline...

src/read_write/node_writer.rs
```diff
@@ -164,3 +196,137 @@ impl CubeNodeWriter {
...
+        };
+
+        self.point_count += 1;
+    }
+}
+
+impl Drop for PlyNodeWriter {
+    fn drop(&mut self) {
+        self.writer.write_all(b"\n").unwrap();
+        if self.writer.seek(SeekFrom::Start(4 + 32 + 15)).is_ok() {
```
nnmm: Do you think it would be cleaner to compute a Seek offset in create_header and store it in the PlyNodeWriter?

feuerste: I updated it to use two consts, one for the part until the number of vertices, one for the number of vertices themselves.

feuerste:
> Looks good! In general, how about one file in read_write per file type containing both iterator and writer?

I moved the new code to separate files.

nnmm:
Okay :+1:

feuerste:
@wally-the-cartographer merge

wally-the-cartographer:

PR #292, opened 2019-07-04
Consolidate node layers.

feuerste:
This moves from a `NodeLayer` enum to attribute strings, so we are more flexible in the future, what kind of attributes we want to read/write.


src/read_write/ply.rs
```diff
@@ -484,27 +484,28 @@ impl PlyNodeWriter {
         }
     }
 
-    fn create_header(&mut self, layers: &[NodeLayer]) {
+    fn create_header(&mut self, attributes: &[&str]) {
```
nnmm: Wouldn't we want an attribute that is essentially (&str, attr_type) here? Otherwise this only works for known attributes.

catevita: could the attribute type be specified in the future from the node meta data?

feuerste: @nnmm Yes, you are right, once the `write` function works over `PointData` instead of `Point`, we want this. I am currently addressing this in a follow-up PR.
@catevita Yes, once we have metadata, we can utilize it, but still the writers should not know too much about metadata, the one calling them probably should...

catevita: yes, I agree

nnmm:
:+1: I also like that we're calling them attributes everywhere now.

feuerste:
Thanks for reviewing, Caterina and Nikolai!

feuerste:
@wally-the-cartographer merge

wally-the-cartographer:

wally-the-cartographer:
Refusing to merge, since the combined checks state is failure

feuerste:
Thanks for the review Caterina and Nikolai!

feuerste:
@wally-the-cartographer merge

wally-the-cartographer:

PR #294, opened 2019-07-08
Support PointData writing.

feuerste:
With this PR we have the possibility to also write `PointData`, which is more generic than only writing `Point`s of position, color, and intensity.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -72,7 +72,7 @@ impl AttributeData {
 /// General structure that contains points and attached feature attributes.
 pub struct PointData {
     pub position: Vec<Vector3<f64>>,
-    // BTreeMap, so we get deterministic iteration order.
+    // BTreeMap for deterministic iteration order.
     pub attributes: BTreeMap<String, AttributeData>,
 }
 
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -70,7 +70,7 @@ impl AttributeData {
 }
 
 /// General structure that contains points and attached feature attributes.
-pub struct PointData {
+pub struct PointsBatch {
     pub position: Vec<Vector3<f64>>,
     // BTreeMap for deterministic iteration order.
     pub attributes: BTreeMap<String, AttributeData>,
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -20,7 +20,7 @@ pub mod read_write;
 pub mod s2_geo;
 
 use cgmath::{Vector3, Vector4};
-use fnv::FnvHashMap;
+use std::collections::BTreeMap;
 
 #[derive(Debug, Clone)]
 pub struct Point {
@@ -42,6 +42,7 @@ pub fn attribute_extension(attribute: &str) -> &str {
 }
 
 /// General field to describe point feature attributes such as color, intensity, ...
+#[derive(Debug, Clone)]
 pub enum AttributeData {
     F32(Vec<f32>),
     F64Vec3(Vec<Vector3<f64>>),
@@ -71,7 +72,10 @@ impl AttributeData {
 /// General structure that contains points and attached feature attributes.
 pub struct PointData {
     pub position: Vec<Vector3<f64>>,
-    pub attributes: FnvHashMap<String, AttributeData>,
+    pub attributes: BTreeMap<String, AttributeData>,
 }
 
 pub use point_viewer_proto_rust::proto;
+
+/// size for batch
+pub const NUM_POINTS_PER_BATCH: usize = 500_000;
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -72,6 +72,7 @@ impl AttributeData {
 /// General structure that contains points and attached feature attributes.
 pub struct PointData {
     pub position: Vec<Vector3<f64>>,
+    // BTreeMap, so we get deterministic iteration order.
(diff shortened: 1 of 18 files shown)


src/lib.rs
```diff
@@ -71,7 +72,11 @@ impl AttributeData {
...
     pub position: Vec<Vector3<f64>>,
-    pub attributes: FnvHashMap<String, AttributeData>,
+    // BTreeMap for deterministic iteration order.
+    pub attributes: BTreeMap<String, AttributeData>,
 }
 
 pub use point_viewer_proto_rust::proto;
+
+/// size for batch
+pub const NUM_POINTS_PER_BATCH: usize = 500_000;
```
catevita: Nit: as this refers to the size of PointData and not anymore to the Batch Iterator, could the name reflect more this change

src/octree/octree_iterator.rs
```diff
@@ -36,7 +36,7 @@ impl NodeIterator<CubeNodeReader> {
         }
 
         if let Ok(mut data_map) = octree_data_provider.data(id, &["intensity"]) {
-            match data_map.remove("position") {
+            match data_map.remove("intensity") {
```
nnmm: Is that a fix for something that slipped in by accident in a previous PR?

feuerste: Yes, sorry, I wanted to mention it... I can also make a mini PR for this, if you prefer...

nnmm: I don't think it's a problem, I was just curious :)

feuerste: Ok

src/lib.rs
feuerste: @nnmm @catevita Or we even rename `PointData` to `PointBatch`? What do you think?

nnmm: +1

src/read_write/cube.rs
```diff
@@ -163,71 +164,104 @@ pub struct CubeNodeWriter {
...
+                    vec3_encode::<f64>(position, min, edge_length)
+                        .write_le(&mut self.xyz_writer)?;
+                }
             }
         }
 
-        p.color.write_le(&mut self.attribute_writers[0]);
-        if let Some(i) = p.intensity {
-            if self.attribute_writers.len() < 2 {
+        if self.point_count == 0 {
```
nnmm: Does PointData always contain > 0 points?

nnmm:
Basic question: The primary motivation for this approach over something like https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=f373b5ffbdc37377a7a5caad663ad9cd is performance, right?

src/read_write/cube.rs
feuerste: Theoretically it can contain anything from 0 to NUM_POINTS_PER_BATCH...

feuerste:
> Basic question: The primary motivation for this approach over something like https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=f373b5ffbdc37377a7a5caad663ad9cd is performance, right?

Yes, here we have far less indirections and still have the possibility to either write "old" points or new points batches...

src/lib.rs
feuerste: Done, I renamed it to `PointsBatch`

src/read_write/cube.rs
nnmm: Maybe this should test `self.attribute_writers.len()` instead?

feuerste: Unfortunately this would repeat the check for each iteration, if we don't write any attributes.

feuerste: This would be possible, but would be slightly slower, as we needed two check when writing no attributes at all.

feuerste: Hm, but since we almost always write attributes, it is nicer, since we can get rid of `point_count`.

nnmm:
That's true, because `write()` does reflection: looping over a str representation of the attributes and calling a 'get' method for each one which returns a general 'attribute' datatype that has to be matched against.

On the other hand, the efficiency in this approach comes from making that reflection less frequent (once per batch), but does that work well with the generation code for octrees? There we usually deal with single points, and you mentioned once that it is not straightforward to translate this code to use batches.

src/read_write/cube.rs
feuerste: Done, with `self.attribute_writers.is_empty()`

nnmm: :+1:

nnmm:
Summary of offline discussion: There is no good way around reflection on single points during node splitting. There we'll need either an `impl NodeWriter<DynamicPoint> for CubeNodeWriter` or some reflection to put a point into another `PointsBatch`.

nnmm:
For now, we'll postpone arbitrary attributes in octree generation and continue to work with `Point`s.

nnmm:
One last high-level comment: What about having a `write()` and `write_batch()` method instead of a type parameter?

feuerste:
> One last high-level comment: What about having a `write()` and `write_batch()` method instead of a type parameter?

This is a matter of taste. I actually had this before, but personally preferred a generic.

feuerste:
@wally-the-cartographer merge

feuerste:

wally-the-cartographer:

PR #295, opened 2019-07-08
Add a few attribute types

nnmm:
Also adds a generic `WriteLE` impl for `Vec<T>`, or you would need to write two impls per attribute.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -63,15 +63,17 @@ impl AttributeData {
             AttributeData::U8Vec4(data) => data.len(),
         }
     }
+
     pub fn is_empty(&self) -> bool {
         self.len() == 0
     }
+
     pub fn dim(&self) -> usize {
         match self {
-            AttributeData::I64(_) => 1,
-            AttributeData::U64(_) => 1,
-            AttributeData::F32(_) => 1,
-            AttributeData::F64(_) => 1,
+            AttributeData::I64(_)
+            | AttributeData::U64(_)
+            | AttributeData::F32(_)
+            | AttributeData::F64(_) => 1,
             AttributeData::F64Vec3(_) => 3,
             AttributeData::U8Vec4(_) => 4,
         }
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -44,7 +44,10 @@ pub fn attribute_extension(attribute: &str) -> &str {
 /// General field to describe point feature attributes such as color, intensity, ...
 #[derive(Debug, Clone)]
 pub enum AttributeData {
+    I64(Vec<i64>),
+    U64(Vec<u64>),
     F32(Vec<f32>),
+    F64(Vec<f64>),
     F64Vec3(Vec<Vector3<f64>>),
     U8Vec4(Vec<Vector4<u8>>),
 }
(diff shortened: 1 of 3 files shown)


src/read_write/node_writer.rs
```diff
@@ -129,24 +147,7 @@ impl WriteLE for Color<u8> {
     }
 }
 
-impl WriteLE for Vec<f32> {
-    fn write_le(&self, writer: &mut DataWriter) -> Result<()> {
-        let mut bytes = vec![0; 4 * self.len()];
-        LittleEndian::write_f32_into(self, &mut bytes);
```
nnmm: @feuerste  Was this optimization important?

nnmm:
meow @feuerste

src/read_write/ply.rs
```diff
@@ -454,7 +454,10 @@ impl NodeWriter<PointsBatch> for PlyNodeWriter {
                         (
                             &k[..],
                             match data {
+                                AttributeData::I64(_) => panic!("Can't write i64 attribute to PLY"),
```
feuerste: Shall we extend `DataType` with `Int64` and `Uint64` and its corresponding `longlong/int64` and `ulonglong/uint64` `from_str` methods instead? I know its not officially supported, but we could still support it to make debugging easier...

src/read_write/node_writer.rs
feuerste: I made this optimization, since the f32 implementation gets a reference, which is internally dereferenced, which is not optimal... and writing a vector of hundreds of thousands of points at once seems to be much faster.

src/read_write/ply.rs
nnmm: Okay

src/read_write/node_writer.rs
nnmm: So you'd prefer we keep the two versions of the instance?

src/read_write/ply.rs
nnmm: Done

src/read_write/node_writer.rs
feuerste: Could we keep one for `cgmath::BaseNum`?

nnmm: How would I get the correct LittleEndian method? A macro?

nnmm: Note that I kept the Vector3 instances, this is only for Vec<_>

feuerste: Ok, then we may need a macro taking the type as str  and its number of bytes and call this for all our types. And we could have another one for the non `Vec` types, too, to reduce the amount of code, right?

nnmm: Done.

nnmm:
ptal @feuerste

src/lib.rs
```diff
@@ -62,7 +68,10 @@ impl AttributeData {
     }
     pub fn dim(&self) -> usize {
         match self {
+            AttributeData::I64(_) => 1,
```
feuerste: Nit: How about
```rust
            AttributeData::I64(_) |
            AttributeData::U64(_) |
            AttributeData::F32(_) |
            AttributeData::F64(_) => 1,
```

nnmm: Done

nnmm:
:dash: Thanks for the review!

nnmm:
@wally-the-cartographer merge

wally-the-cartographer:

PR #296, opened 2019-07-10
Fix color writing and compatibility.

feuerste:
PR #294 introduced writing of alpha values of colors. Previously we only wrote rgb. Since we also only read rgb, this lead to buggy behavior. This PR unifies reading and writing again to just rgb, so we are compatible with old files.
Also the representation of colors in `AttributeData` is changed to a three-vector.


feuerste:
@wally-the-cartographer merge

feuerste:

wally-the-cartographer:

PR #297, opened 2019-07-10
Basic split writer.

feuerste:
Introduces a basic s2 cell based split writer taking a base path and s2 level to split at, which internally splits and writes a passed points batch into files inside the base path with the keys corresponding to the s2 cell as filenames.


src/s2_geo.rs
```diff
@@ -8,6 +9,16 @@ use std::vec::Vec;
 
 const LEVEL_MOD: u8 = 1;
 
+pub trait ECEFExt<N> {
+    fn from(_: Self) -> ECEF<N>;
```
nnmm: I'm curious, wouldn't you normally write `from(self)`?

feuerste: Yes, this is nicer, I adjusted it everywhere.

nnmm:
I'd like to understand better why this should be a trait instead of a struct. My first thought is whether we actually need more than one splitting function. The second is that IMO in a perfectly composable world, we wouldn't need a `PlySplitWriter` and `RawSplitWriter` but have a struct `S2Splitted<T: NodeWriter>`, maybe even `S2Splitted<T: NodeWriter, L: S2Level>` that implements `NodeWriter` itself. I guess the difficulty with that is that the splitter needs to create new inner writers, and the `CubeNodeWriter` prevents us from making `new(path)` a trait method *. Then again, do we really want to support using this with anything other than `RawNodeWriter`?

* One way around is to have a `SimpleNodeWriter` extension to `NodeWriter` that introduces an additional `new_in_dir(path: PathBuf) -> Self` method that would not be implemented by `CubeNodeWriter`.

feuerste:
Yes, the difficulty are the inner writers, since their construction needs different parameters. I thought about getting around this by providing something like a "get cube/encoding for key" function, which can be passed to the constructor of a possible `CubeSplitWriter`. And yes, it was also my thinking that we can use the splitter also during octree generation for writing cubes.

nnmm:
What do you think about adding an encoding parameter to the `new()` method. I.e. either 
```
pub enum Encoding {
    Plain,
    ScaledCentered(Vector3<f64>, f64, PositionEncoding),
}

pub trait NodeWriter<P> {
    fn new(dir: &Path, codec: Encoding) -> Self;
    fn write(&mut self, p: &P) -> Result<()>;
}
```
or, to eliminate the inner match against each point
```
struct ScaledCentered<Dtype> {
    origin: Vector3<f64>,
    edge_length: f64,
}

trait Codec {
    fn encode_pos(v: &Vector3<f64>) -> impl WriteLE;
}

impl Codec for ScaledCentered<u16> {...}

pub trait NodeWriter<P, C: Codec> {
    fn new(dir: &Path, codec: C) -> Self;
    fn write(&mut self, p: &P) -> Result<()>;
}
```
?

feuerste:
I agree that we don't need as much flexibility for now... I refactored the writers in #300 and implemented an easier version here.
@nnmm ptal, first #300 and then this one...

src/read_write/s2.rs
```diff
@@ -0,0 +1,76 @@
...
+    writers: HashMap<String, W>,
+    stem: PathBuf,
+    split_level: u8,
+}
+
+impl<W> S2Splitter<W>
+where
+    W: NodeWriter<PointsBatch>,
+{
+    pub fn new(path: impl Into<PathBuf>, split_level: u8) -> Self {
```
nnmm: Is the split_level the reason why you didn't `impl NodeWriter for S2Splitter`?

src/read_write/s2.rs
```diff
@@ -0,0 +1,76 @@
...
+        let stem = path.into();
+        S2Splitter {
+            writers,
+            stem,
+            split_level,
+        }
+    }
+
+    pub fn write(&mut self, points_batch: &PointsBatch) -> Result<()> {
+        let mut out = HashMap::new();
```
nnmm: Nit: Consider naming this `batches_by_s2cell` and/or adding a comment that this is a buffer with one entry per S2Cell.

sdl_viewer/Cargo.toml
```diff
@@ -25,7 +25,7 @@ gl_generator = "0.8.0"
 cgmath = "0.17.0"
 clap = "2.33.0"
 collision = "0.20.1"
-lru-cache = "^0.1.1"
```
feuerste: lru-cache is not maintained any more, so I adjusted it here, too

feuerste:
PTAL @nnmm 
I integrated an open mode to specify whether to truncate the files when opening them and unified the application of the encoding to all writers.

src/read_write/ply.rs
```diff
@@ -485,7 +489,7 @@ impl NodeWriter<PointsBatch> for PlyNodeWriter {
         }
 
         for (i, pos) in p.position.iter().enumerate() {
-            pos.write_le(&mut self.writer)?;
+            pos.write_le_encoded(&self.encoding, &mut self.writer)?;
```
nnmm: Isn't it a bit nonsensical to write data as anything but LE? Third-party tools won't be able to read that, and I'm not aware of a use-case for having our own compressed PLY format.

src/read_write/ply.rs
```diff
@@ -525,61 +530,80 @@ impl NodeWriter<Point> for PlyNodeWriter {
...
-                self.point_count,
-                width = HEADER_NUM_VERTICES.len()
-            );
+        if self.point_count == 0 {
+            return;
         }
+        self.writer.write_all(b"\n").unwrap();
+        self.writer.flush().unwrap();
+
+        let mut file = self.writer.get_mut().get_mut();
```
nnmm: Could you comment on why this was changed?

src/read_write/ply.rs
feuerste: It is still LE and third party tools can still read it, it's just in other coordinates. I here preferred interface consistency, but we can also fail the writing in case an encoding other than Plain was specified... but then again, the interface doesn't make so much sense to me...

src/read_write/ply.rs
feuerste: Since we now have an append mode, we need to update the number of vertices written. It was much easier to directly use the underlying `File` writer (which implements both the `Read` and `Write` traits) for this instead of adding read functionality to the `DataWriter`, which internally holds a `BufWriter`.

src/read_write/ply.rs
nnmm: Good point, you're right that it makes sense to combine the ScaledToCube with PLY. I think should we add more encoding options in the future, we can return a Result<Self> from the `NodeWriter`'s `from()` function that may return `UnsupportedEncoding` or similar.

feuerste: Yes, this may be an option.
Furthermore, we could also think of including the encoding in ply comments, so we wouldn't need metadata along with the ply... what do you think?

src/read_write/node_writer.rs
```diff
@@ -197,7 +215,74 @@ impl WriteLEPos for AttributeData {
     }
 }
 
+pub trait WriteLEEncoded {
```
nnmm: NIt: Could this go into codec.rs or would that not make sense in the current dependency graph? Also, how about naming this `WriteEncoded` since future encodings (like compression) might not really be LE anymore?

feuerste: I thought about it, but since it takes the `DataWriter` as parameter, I decided to leave it here to be consistent with the other two traits... I will rename it.

src/read_write/ply.rs
nnmm: Hmm, could `PlyNodeWriter::new()` also just open the file (if any) in read mode and initialize `point_count` from the contents?

feuerste: Yes, this would be another alternative, but then we would need two internal variables `point_count_total` and `point_count_this_session` or the like to keep track of the states...

nnmm: Where do we need to distinguish these from just their sum?

feuerste: We currently have `if self.point_count == 0` at a couple of places, these could also be replaced with a boolean variable and checking `if !self.points_written`, but at least there needs to be a way to find out whether we need to create a header... If you think it's nicer this way, I can change it.

nnmm: Is the reason we can't just write the header every time that some code assumes that only empty files can be safely deleted?

feuerste: As discussed offline, I updated the logic to read in the point count.

feuerste:

feuerste:
@wally-the-cartographer merge

wally-the-cartographer:

PR #300, opened 2019-07-11
Writer refactoring

feuerste:
This is in preparation for #297, where we don't want to overcomplicate things.
Here we
- Merge `RawNodeWriter` and `CubeNodeWriter` into one `RawNodeWriter`, both taking an encoding enum as argument
- Move `PositionEncoding` from `octree` to `read_write`
- and a few minor changes...


src/read_write/node_iterator.rs
```diff
@@ -17,25 +17,39 @@ use std::io::Result;
...
 }
 
 impl<R> NodeIterator<R>
 where
     R: NodeReader,
 {
-    pub fn new(reader: R) -> Self {
-        if reader.num_points() == 0 {
-            return NodeIterator::Empty;
+    pub fn new(reader: R, num_points: usize) -> Self {
```
nnmm: Could you comment on why this changed? Does that mean you have to declare beforehand how many points you want to read?

feuerste: Unfortunately this has always been there, see e.g. https://github.com/googlecartographer/point_cloud_viewer/blob/master/src/octree/generation.rs#L195 for its usage. I actually just moved it from the reader to the iterator, so at least the reader doesn't need to know it beforehand...

src/octree/generation.rs
```diff
@@ -36,7 +37,7 @@ use std::sync::mpsc;
 const UPDATE_COUNT: i64 = 100_000;
 const MAX_POINTS_PER_NODE: i64 = 100_000;
 
-impl CubeNodeWriter {
+impl RawNodeWriter {
```
nnmm: The name might not be descriptive enough now that it's not necessarily "Raw" (unencoded) anymore – how about `PerAttributeNodeWriter`?

src/read_write/node_iterator.rs
nnmm: Ah, okay.

src/octree/generation.rs
feuerste: Yeah, agree, didn't find a better name before... I meant raw in the sense of not having a file format with header... `PerAttribute...` is fine with me, too, just a little long...

feuerste:

feuerste:
@wally-the-cartographer merge

wally-the-cartographer:

PR #303, opened 2019-07-16
BufReader: buffer size set to 1MB 

catevita:
BufReader wraps and impl Read struct in order to optimize the read() calls (calling less often), while caching the contents in an internal buffer. The default size of 8kb is quite tiny for point clouds, 1MB is a quite good limit for a buffer, especially in multi-threaded, multi-buffer implementation.

The benchmark test gives pretty much the same results on local disk, improvements should be noticed acquiring the data from remote.


catevita:
PTAL @feuerste

feuerste:
Can you please update the description of this PR and explain its reasoning a bit?

src/read_write/mod.rs
```diff
@@ -38,3 +38,5 @@ pub use self::raw::{RawNodeReader, RawNodeWriter};
 
 mod s2;
 pub use self::s2::S2Splitter;
+
+pub const BUFFER_CAPACITY_1MB: usize = 1024 * 1024;
```
feuerste: How about just having a `pub const MB: u64 = 1024 * 1024;` in `src`, or, if it should be local to read/write, remove the `_1MB` suffix?

catevita: i changed it into pub const MB: usize = 1024 * 1024; in src

catevita: .. usize because capacity mostly requires usize type

src/read_write/ply.rs
```diff
@@ -403,9 +403,11 @@ impl PlyIterator {
         }
 
         // We align the buffer of this 'BufReader' to points, so that we can index this buffer and know
-        // that it will always contain full points to parse.
+        // that it will always contain full points to parse. Capacity should be approximately 1MB
+        let num_points_in_mb: usize = BUFFER_CAPACITY / num_bytes_per_point;
```
feuerste: This should probably not use `BUFFER_CAPACITY`, as it may be different to one MB.

src/lib.rs
```diff
@@ -88,5 +88,7 @@ pub struct PointsBatch {
 
 pub use point_viewer_proto_rust::proto;
 
-/// size for batch
-pub const NUM_POINTS_PER_BATCH: usize = 500_000;
+/// size for batch - with ca 24 bytes per point ~< 0.5 MB total
```
feuerste: I don't understand this comment, 24*128*1024 gives about 3.1 MB?

catevita:
the modifications are not necessary anymore.

PR #304, opened 2019-07-16
Use std::slice::from_ref to prettify code

nnmm:
This is a handy function that I think all of us overlooked when we made the BatchIterator accept `&[Octree]`: https://doc.rust-lang.org/std/slice/fn.from_ref.html


nnmm:
ptal @catevita 
fyi @feuerste

feuerste:
@wally-the-cartographer merge

PR #309, opened 2019-07-29
S2 Metadata

catevita:
First Version: trying to keep the most of the structure compatible with octrees


catevita:
PTAL @feuerste @nnmm

point_viewer_proto_rust/src/proto.proto
```diff
@@ -75,23 +75,50 @@ message NodeId {
...
+}
 
+
+message Node {
   PositionEncoding position_encoding = 2;
   int64 num_points = 3;
   NodeId id = 4;
 }
 
+enum AttributeData{
```
feuerste: How about `AttributeType` or `AttributeDataType`?

feuerste: `enum AttributeData{` -> `enum AttributeData {`

point_viewer_proto_rust/src/proto.proto
```diff
@@ -75,23 +75,50 @@ message NodeId {
...
+    I64 = 0;
+    U64 = 1;
+    F32 = 2;
+    F64 = 3;
+    U8Vec3 = 4;
+    F64Vec3 = 5;
+}
+
+message Attribute{
```
feuerste: `message Attribute{` -> `message Attribute {`

point_viewer_proto_rust/src/proto.proto
```diff
@@ -75,23 +75,50 @@ message NodeId {
...
+  string name = 1;
+  AttributeData attr_type = 2;
+}
+
+message S2Cell {
+  PositionEncoding position_encoding = 1;
+  uint64 cell_id = 2;
```
feuerste: `cell_id` -> `id`?

point_viewer_proto_rust/src/proto.proto
```diff
@@ -75,23 +75,50 @@ message NodeId {
...
+  uint64 num_points = 3;
```
feuerste: `uint64` -> `int64`?

point_viewer_proto_rust/src/proto.proto
```diff
@@ -75,23 +75,50 @@ message NodeId {
...
+}
+
+
 message Meta {
   int32 version = 1;
   AxisAlignedCuboid bounding_box = 4;
   double resolution = 3;
-  repeated Node nodes = 5;
+  repeated Node nodes = 5; 
+  repeated S2Cell cells = 6;
```
feuerste: Do you think a comment saying `nodes` and `cells` are mutually exclusive and used either for octrees or s2 cells would make sense?

src/read_write/codec.rs
```diff
@@ -39,25 +40,25 @@ impl PositionEncoding {
...
+    pub fn from_proto(proto: proto::PositionEncoding) -> Result<Self> {
         match proto {
-            proto::Node_PositionEncoding::INVALID => {
-                Err(ErrorKind::InvalidInput("Invalid PositionEncoding".to_string()).into())
+            proto::PositionEncoding::Uint8 => Ok(PositionEncoding::Uint8),
+            proto::PositionEncoding::Uint16 => Ok(PositionEncoding::Uint16),
+            proto::PositionEncoding::Float32 => Ok(PositionEncoding::Float32),
+            proto::PositionEncoding::Float64 => Ok(PositionEncoding::Float64),
+            proto::PositionEncoding::INVALID => {
+                Err(ErrorKind::Proto("Proto: PositionEncoding is invalid".to_string()).into())
```
feuerste: Do we really need a new error type for this or can we just leave it with `InvalidInput`?

src/read_write/attribute_reader.rs
```diff
@@ -0,0 +1,22 @@
...
+    };
+    Ok((proto.attr_name, attr_type))
+}
+
+fn read_source<T>(attr_source: Box<dyn Read>) -> io::Result<Vec<T>>{
+   let bufreader = BufReader::<T>::new(attr_source);
+   let mut vec = Vec::<T>::new();
+   bufreader.read_all(&mut vec)?;
+   vec
+}
```
feuerste: Nit: Add newline

point_viewer_proto_rust/src/proto.proto
catevita: I thought I'd use proto::AttributeData -> AttributeData like proto::PositionEncoding.. any thoughts?

feuerste: The difference is that `AttributeData` really contains data, while this `enum` here only describes the data type it contains.

point_viewer_proto_rust/src/proto.proto
```diff
@@ -75,23 +75,50 @@ message NodeId {
...
+    U64 = 1;
+    F32 = 2;
+    F64 = 3;
+    U8Vec3 = 4;
+    F64Vec3 = 5;
+}
+
+message Attribute{
+  string name = 1;
+  AttributeData attr_type = 2;
```
feuerste: `attr_type` -> `type`?

src/read_write/codec.rs
catevita: the description of invalid input says ("The input file is not supported or invalid") does not fit really good. shall I remove "file" from the description?

feuerste: Yep, let's just remove file and stick with the old error type.

point_viewer_proto_rust/src/proto.proto
catevita: type is reserved, hence data_type

point_viewer_proto_rust/src/proto.proto
```diff
@@ -75,23 +75,51 @@ message NodeId {
...
+}
 
+
+message Node {
   PositionEncoding position_encoding = 2;
   int64 num_points = 3;
   NodeId id = 4;
 }
 
+enum AttributeType{
```
feuerste: Missing space before opening braces

point_viewer_proto_rust/src/proto.proto
```diff
@@ -75,23 +75,51 @@ message NodeId {
...
+    I64 = 0;
+    U64 = 1;
+    F32 = 2;
+    F64 = 3;
+    U8Vec3 = 4;
+    F64Vec3 = 5;
+}
+
+message AttributeDescription{
```
feuerste: Missing space before opening braces

point_viewer_proto_rust/src/proto.proto
```diff
@@ -75,23 +75,51 @@ message NodeId {
...
 
+
+message Node {
   PositionEncoding position_encoding = 2;
   int64 num_points = 3;
   NodeId id = 4;
 }
 
+enum AttributeType{
+    I64 = 0;
```
feuerste: 0 is reserved for unset (https://github.com/lyft/avsoftware/blob/master/STYLE.md#protobuf-coding-style) or unspecified (https://developers.google.com/protocol-buffers/docs/style#enums), so we should add it here...

point_viewer_proto_rust/src/proto.proto
```diff
@@ -75,23 +75,51 @@ message NodeId {
...
+    F64Vec3 = 5;
+}
+
+message AttributeDescription{
+  string name = 1;
+  AttributeType data_type = 2;
+}
+
+message S2Cell {
+  PositionEncoding position_encoding = 1;
```
feuerste: Shall we add this later, once we use it, or just ignore it for now?

point_viewer_proto_rust/src/proto.proto
feuerste: How about shortening this to just `Attribute`?

point_viewer_proto_rust/src/proto.proto
feuerste: Since it's called `data_type` below, how about unifying it everywhere to `AttributeDataType`?

src/read_write/s2.rs
```diff
@@ -101,3 +103,39 @@ where
         self.writers.get_mut(cell_id).unwrap()
     }
 }
+
+#[allow(dead_code)]
```
feuerste: Shall we add these functions later, once we use them?

point_viewer_proto_rust/src/proto.proto
catevita: as every field is optional, i think this mirrors also node contents as well and will hopefully be soon in use

src/read_write/s2.rs
catevita: I'd like to use them in avmg for some checks in point cloud generation reduce step

feuerste: Ok, then we should remove the `dead_code` allowance and add the two methods to the `pub use self::s2::{...}` section in `src/read_write/mod.rs`

src/read_write/s2.rs
```diff
@@ -101,3 +103,39 @@ where
...
+pub fn s2_cloud_to_meta_proto(
```
feuerste: What about the `bounding_box` (and possibly `resolution`)? I think we need a way to fill these fields, too, otherwise it may make more sense to have a separate `S2Meta`, which would also clarify the difference between nodes and s2cells...

src/read_write/attribute_reader.rs
```diff
@@ -0,0 +1,22 @@
+use crate::proto;
+use crate::AttributeData;
+use::io::{BufReader, Read, Result};
+
+pub fn attr_from_proto(proto : proto::Attribute, attr_source: Box<dyn Read>)-> io::Result<(String, AttributeData)> {
+    let attr_type = match proto.attr_type {
+       proto::AttributeType::I64 => AttributeDat0a::I64(read_source(attr_source)?),
```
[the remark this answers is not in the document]
catevita: That was not meant to be in the PR, deleted now

src/read_write/s2.rs
catevita: indeed in the future they could be an interesting information about s2 cell points, Do we mean with bounding box the current one from the current inserted points, or the s2cell approximation into a bounding box? Resolution can be also expressed as a kind of average surface density in the current populated s2cell bounding box.

feuerste:
@nnmm Can you please also have a look at the proto changes? I think we should all agree on them...

point_viewer_proto_rust/src/proto.proto
```diff
@@ -75,23 +75,47 @@ message NodeId {
...
+}
 
+message Node {
   PositionEncoding position_encoding = 2;
   int64 num_points = 3;
   NodeId id = 4;
 }
 
+enum AttributeDataType {
+    INVALID_DATA_TYPE = 0;
```
feuerste: Nit: `INVALID` would be enough...

nnmm: Can you explain why this is needed?

point_viewer_proto_rust/src/proto.proto
```diff
@@ -75,23 +75,47 @@ message NodeId {
...
+    I64 = 1;
```
nnmm: Nit: How about numbering them according to a schema like unsigned integers = `1X`, signed integers = `2X`, floats = `3X`, vec2 = `1YX` with `Y ` like above, vec3 = `2YX`, so that when we add datatypes they're not all jumbled? You could also add all the ints from u/i8 to u/i64.

point_viewer_proto_rust/src/proto.proto
feuerste: See https://github.com/googlecartographer/point_cloud_viewer/pull/309#discussion_r308626031

src/lib.rs
```diff
@@ -22,6 +22,12 @@ pub mod s2_geo;
 use cgmath::Vector3;
 use std::collections::BTreeMap;
 
+// Version 9 -> 10: Change in NodeId proto from level (u8) and index (u64) to high (u64) and low
+// (u64). We are able to convert the proto on read, so the tools can still read version 9.
+// Version 10 -> 11: Change in AxisAlignedCuboid proto from Vector3f min/max to Vector3d min/max.
+// We are able to convert the proto on read, so the tools can still read version 9/10.
+pub const CURRENT_VERSION: i32 = 11;
```
nnmm: Shouldn't the addition of attributes warrant a new version? To migrate, we can maybe just set attributes to color and intensity.

point_viewer_proto_rust/src/proto.proto
```diff
@@ -75,23 +75,47 @@ message NodeId {
...
+enum PositionEncoding {
     INVALID = 0;
     Uint8 = 1;
     Uint16 = 2;
     Float32 = 3;
     Float64 = 4;
-  }
+}
 
+message Node {
```
nnmm: Nit: If we're making breaking changes anyway, could this become OctreeNode?

src/read_write/s2.rs
nnmm: Resolution doesn't make sense in combination with no PositionEncoding, so I think we should just leave it unset. The bounding box doesn't need to be exact, but we should probably have it since it's used in the client, right?

nnmm:
In general the proto changes look good to me.

point_viewer_proto_rust/src/proto.proto
catevita: it is needed by proto for value 0, and the by compiling it complained having 2 INVALID in the same file (which i also found curious though

point_viewer_proto_rust/src/proto.proto
catevita: the schema does not completely work with multiplication but with offset addition, otherwise without you get some real jam. 
enum AttributeDataType {

    INVALID_DATA_TYPE = 0;
    U8 = 1;
    **U16 = 2;**
    U32 = 3;
    U64 = 4;
    **I8 = 2*1;**
    I16 = 2*2;
    I32 = 2*3;
    I64 = 2*4;

    F32 = 3*3;
    F64 = 3*4; //max value 12
    
    U8Vec3 = 12*2 + U8 ;
    F64Vec3 = 6;
}

nnmm: I'm not sure I understand – I'm proposing that unsigned integers are 10, 11, 12, ... and signed integers are 20, 21, 22, ..., etc., U8Vec3 would then be 210.

point_viewer_proto_rust/src/proto.proto
catevita: done

src/read_write/s2.rs
catevita: here i am still in doubt what to do exactly here

src/bin/upgrade_octree.rs
```diff
@@ -79,8 +79,11 @@ fn main() {
         match meta.version {
             9 => upgrade_version9(&args.directory, meta),
             10 => upgrade_version10(&args.directory, meta),
-            other if other == octree::CURRENT_VERSION => {
-                println!("Octree at current version {}", octree::CURRENT_VERSION);
+            other if other == point_viewer::CURRENT_VERSION => {
```
nnmm: Ideally we'd have `upgrade_version11` in this PR

nnmm: Ah, nevermind, I saw that it's backwards-compatible. That's good.

src/read_write/codec.rs
```diff
@@ -137,3 +139,15 @@ where
 {
     num::cast::<T, f64>(value).unwrap() * edge_length + min
 }
+
+pub fn attribute_to_proto(attribute: &AttributeData) -> proto::AttributeDataType {
```
nnmm: Nit: This could be a method of AttributeDataType

nnmm: In src/lib.rs

src/read_write/s2.rs
nnmm: Do we even read the resolution anywhere?

nnmm: I just checked and I don't think so. So it's not too dangerous to leave it unset. As for bounding box, I'm fine with leaving it unset as well.

point_viewer_proto_rust/src/proto.proto
catevita: now i got it! I did it with just a tighter offset

src/read_write/codec.rs
catevita: good idea, done!

point_viewer_proto_rust/src/proto.proto
```diff
@@ -75,23 +75,53 @@ message NodeId {
...
+  int64 num_points = 2;
+}
+
 message Meta {
   int32 version = 1;
   AxisAlignedCuboid bounding_box = 4;
   double resolution = 3;
-  repeated Node nodes = 5;
+  // Either nodes or cells is allowed, both is invalid
+  repeated OctreeNode nodes = 5;
```
feuerste: How about something like
```rust
message OctreeNodes {
  repeated OctreeNode nodes = 1;
}

message S2Cells {
  repeated S2Cell cells = 1;
}

oneof data {
  OctreeNodes octree_nodes = 1;
  S2Cells s2_cells = 2;
}
```
to make the either/or relationship clear?

point_viewer_proto_rust/src/proto.proto
```diff
@@ -119,13 +119,28 @@ message S2Cell {
...
+message OctreeMeta {
+  AxisAlignedCuboid bounding_box = 1;
+  double resolution = 2;
+  repeated OctreeNode nodes = 3;
+}
+
+message S2Meta {
+  repeated S2Cell cells = 1;
+  repeated Attribute attributes = 2;
+}
```
nnmm: Should we also have a `level` field?

feuerste: I would add it, once we need it... can you already think of a use case at the moment?

catevita: the level can be also indirectly given from the s2 cell name, which should be unique up to level 30.

src/lib.rs
```diff
@@ -23,6 +23,14 @@ pub mod read_write;
 use cgmath::Vector3;
 use std::collections::BTreeMap;
 
+// Version 9 -> 10: Change in NodeId proto from level (u8) and index (u64) to high (u64) and low
+// (u64). We are able to convert the proto on read, so the tools can still read version 9.
+// Version 10 -> 11: Change in AxisAlignedCuboid proto from Vector3f min/max to Vector3d min/max.
+// We are able to convert the proto on read, so the tools can still read version 9/10.
+// Version 11 -> 12: Change in Meta names, to have one version for s2 or octree data.
+// We are able to convert the proto on read, so the tools can still read version 9/10.
```
nnmm: Nit: This line should have version 11

src/lib.rs
```diff
@@ -65,6 +73,32 @@ pub enum AttributeDataType {
...
+            AttributeDataType::I64 => proto::AttributeDataType::I64,
+            AttributeDataType::U64 => proto::AttributeDataType::U64,
+            AttributeDataType::F32 => proto::AttributeDataType::F32,
+            AttributeDataType::F64 => proto::AttributeDataType::F64,
+            AttributeDataType::U8Vec3 => proto::AttributeDataType::U8Vec3,
+            AttributeDataType::F64Vec3 => proto::AttributeDataType::F64Vec3,
+        }
+    }
+
+    pub fn from_attribute(attribute: &AttributeData) -> Self {
```
nnmm: How about moving this into a `data_type()` method of `AttributeData`?

src/octree/mod.rs
```diff
@@ -134,57 +139,89 @@ pub struct NodeData {
...
+                    resolution: meta_proto.deprecated_resolution,
+                    bounding_box,
+                };
+                nodes_proto = meta_proto.get_deprecated_nodes();
+            }
+            CURRENT_VERSION => {
+                if !meta_proto.has_octree() {
+                    return Err(ErrorKind::InvalidInput("No octree meta found".to_string()).into());
+                }
+                let octree_meta = Some(meta_proto.get_octree()).unwrap();
```
nnmm: `Some(x).unwrap()` – probably left over from refactoring? :)

src/octree/mod.rs
```diff
@@ -134,57 +139,89 @@ pub struct NodeData {
...
+                    meta_proto.version, CURRENT_VERSION
+                );
+                bounding_box = bounding_box_to_aabb(meta_proto.get_deprecated_bounding_box());
+                meta = OctreeMeta {
+                    resolution: meta_proto.deprecated_resolution,
+                    bounding_box,
+                };
+                nodes_proto = meta_proto.get_deprecated_nodes();
+            }
+            11 => {
```
nnmm: Seems like 11 is the same as above, so could we make the above `9 | 10 | 11`?

src/octree/mod.rs
```diff
@@ -134,57 +139,89 @@ pub struct NodeData {
...
+        v
+    });
+    Aabb3::new(
+        Point3::new(min.x, min.y, min.z),
+        Point3::new(max.x, max.y, max.z),
+    )
+}
+
 impl Octree {
     // TODO(sirver): This creates an object that is only partially usable.
```
nnmm: Do you know what Holger meant by that? Is that still the case?

src/octree/mod.rs
catevita: no, it's intentional. you are left with a sort of 'pointer', you have to unwrap it to access the structure within

nnmm:
:+1: LGTM

src/octree/mod.rs
catevita: I do not know as well.. fyi @SirVer, whenever you have time to explain the meaning behind that

catevita:
@nnmm I guess you have to explicitly approve the PR selecting field and button

src/bin/upgrade_octree.rs
```diff
@@ -66,6 +66,23 @@ fn upgrade_version10(directory: &Path, mut meta: proto::Meta) {
...
 }
 
+fn upgrade_version11(directory: &Path, mut meta: proto::Meta) {
+    println!("Upgrading version 11 => 12.");
+    let mut octree = proto::OctreeMeta::new();
+
+    octree.set_bounding_box(meta.take_deprecated_bounding_box());
+
+    octree.set_resolution(meta.deprecated_resolution);
+    meta.deprecated_resolution = 0.0;
```
feuerste: What happens, if we don't set this to `0.0`, i.e. can we just leave it?

src/octree/mod.rs
```diff
@@ -134,57 +139,76 @@ pub struct NodeData {
...
     pub fn from_data_provider(data_provider: Box<dyn DataProvider>) -> Result<Self> {
+        let bounding_box;
```
feuerste: How about
```rust
    let (bounding_box, meta, nodes_prot) = match meta_proto.version {
            9 | 10 | 11 => {
                println!(
                    "Data is an older octree version: {}, current would be {}. \
                     If feasible, try upgrading this octree using `upgrade_octree`.",
                    meta_proto.version, CURRENT_VERSION
                );
                let bounding_box = bounding_box_to_aabb(meta_proto.get_deprecated_bounding_box());
                (
                    bounding_box, 
                    OctreeMeta {
                        resolution: meta_proto.deprecated_resolution,
                        bounding_box,
                    },
                    meta_proto.get_deprecated_nodes(),
                )
            }
            CURRENT_VERSION => {
                if !meta_proto.has_octree() {
                    return Err(ErrorKind::InvalidInput("No octree meta found".to_string()).into());
                }
                let octree_meta = meta_proto.get_octree();
                let bounding_box = bounding_box_to_aabb(octree_meta.get_bounding_box());
                (
                    bounding_box,
                    OctreeMeta {
                        resolution: octree_meta.resolution,
                        bounding_box,
                    },
                    octree_meta.get_nodes(),
                )
            }
            _ => return Err(ErrorKind::InvalidVersion(meta_proto.version).into()),
    }
```
?

src/bin/upgrade_octree.rs
catevita: I just copied the way it is in version 9 with deprecated level. btw take() also gets the contents by vale and replaces it with new(), so it sorts of clears the contents without having borrowing problems

catevita:
:dash:

PR #310, opened 2019-07-30
Use build.rs to install protoc

feuerste:
With this PR one less manual step is required to setup protoc, which also makes it easier to depend on this package elsewhere.
See also #280


.gitmodules
```diff
@@ -0,0 +1,4 @@
+[submodule "third_party/protobuf"]
+	path = third_party/protobuf
+	url = https://github.com/google/protobuf.git
+	ignore = dirty
```
tblaette: Are you fixing the version/tag that is checked out of the submodule anywhere? The previous shell script always checked out version 3.6.1.

build.rs
```diff
@@ -0,0 +1,30 @@
...
+    let manifest_dir = env::var("CARGO_MANIFEST_DIR").unwrap();
+    // Build and install proto3
+    Config::new("third_party/protobuf/cmake")
+        .define("CMAKE_BUILD_TYPE", "Release")
+        .define(
+            "CMAKE_INSTALL_PREFIX",
+            format!("{}/target/protobuf", manifest_dir),
+        )
+        .define("CMAKE_POSITION_INDEPENDENT_CODE", "True")
+        .define("protobuf_BUILD_TESTS", "False")
```
tblaette: Is it ok to use `True`/`False` and `ON`/`OFF` interchangeably?

.gitmodules
feuerste: Yes, if you scroll to the very bottom, you will see `Submodule protobuf added at 48cb18`... `48cb18` corresponds to `v3.6.1`

build.rs
feuerste: Yep, there is a long list evaluating to 0/1, see https://cmake.org/cmake/help/latest/manual/cmake-generator-expressions.7.html#logical-operators

feuerste:
Thanks for the review, Tom!

feuerste:
:dash:

feuerste:
@wally-the-cartographer merge

PR #311, opened 2019-08-01
Remove ninja from build deps.

feuerste:
This is a small follow-up to #310


feuerste:
Thanks for the review!

feuerste:
@wally-the-cartographer merge

PR #312, opened 2019-08-02
Fix protoc generation and dependencies.

feuerste:
Previously depending on `target/protobuf/bin` to be in the path during build time was dangerous, as build dependencies were not resolved in a way that we could rely on the `protoc` binary already available. This PR fixes this by introducing a separate lib taking care of everything.


feuerste:
Thanks for your review!

feuerste:
@wally-the-cartographer merge

PR #314, opened 2019-08-05
Move data provider out of octree

nnmm:
The data provider is actually independent of the point cloud representation (as long as it has nodes). This PR strips the "Octree" prefix from `DataProvider` and moves the module out of `octree`. The trait interface had to be changed slightly to take `node_id` as a`&str` instead of an `octree::NodeId`, but it was always converted to a String anyway.


nnmm:
ptal @feuerste @catevita

point_viewer_grpc/src/lib.rs
```diff
@@ -30,12 +30,12 @@ use std::sync::Arc;
 
 pub mod service;
 
-pub struct GrpcOctreeDataProvider {
+pub struct GrpcDataProvider {
     client: OctreeClient,
```
catevita: should be the client also renamed with something non-octree-related?

point_viewer_grpc/src/lib.rs
```diff
@@ -46,7 +46,7 @@ impl GrpcOctreeDataProvider {
             .connect(addr);
         let client = OctreeClient::new(ch);
 
-        Ok(GrpcOctreeDataProvider { client, octree_id })
+        Ok(GrpcDataProvider { client, octree_id })
```
catevita: should  ```octree_id``` become just an ```id```?

point_viewer_grpc/src/lib.rs
```diff
@@ -108,7 +108,7 @@ impl GrpcOctreeDataProvider {
     }
 }
 
-impl OctreeDataProvider for GrpcOctreeDataProvider {
+impl DataProvider for GrpcDataProvider {
     fn meta_proto(&self) -> Result<Meta> {
         let mut req = proto::GetMetaRequest::new();
         req.set_octree_id(self.octree_id.clone());
```
catevita: ```set_octree_id``` -> ```set_id``` ?

nnmm:
Thanks for the comments, I realized that actually `GrpcDataProvider` is octree-specific, so we should wait with all the renaming until it is general enough to handle any point cloud. I also changed its name back to `GrpcOctreeDataProvider`.

nnmm:
@wally-the-cartographer merge

PR #315, opened 2019-08-08
Make batch_iterator not directly depend on octree

nnmm:
`BatchIterator` is now based on a new `PointCloud` trait that can be implemented by S2-based pointclouds in the future. Since `impl Trait` is disallowed in trait definitions, I removed the type parameter from `FilteredPointsIterator` and specialized it to only filter by `PointCulling`'s `contains` method. 

I moved the `batch_iterator` module to the top level as well and renamed `intersects` to `intersects_aabb3`, to contrast it with the planned `intersects_s2_cell` method.


nnmm:
meow @feuerste 
fyi @catevita

src/batch_iterator.rs
```diff
@@ -129,18 +125,28 @@ where
...
+    type PointIter: Iterator<Item = Point>;
+    fn nodes_in_location(&self, query: &PointQuery) -> Vec<Self::Id>;
+    fn points_in_node(&self, query: &PointQuery, node_id: Self::Id) -> Self::PointIter;
+}
+
 /// Iterator on point batches
-pub struct BatchIterator<'a> {
-    octrees: &'a [octree::Octree],
+pub struct BatchIterator<'a, C> {
+    octrees: &'a [C],
```
feuerste: Should we rename this to `point_cloud`?

src/batch_iterator.rs
```diff
@@ -129,18 +125,28 @@ where
...
 }
 
-impl<'a> BatchIterator<'a> {
+impl<'a, C> BatchIterator<'a, C>
+where
+    C: PointCloud,
+{
     pub fn new(
-        octrees: &'a [octree::Octree],
+        octrees: &'a [C],
```
feuerste: -> `point_cloud`?

nnmm: I did that, but apparently saved to the old location. Committed now.

src/batch_iterator.rs
```diff
@@ -129,25 +125,35 @@ where
     }
 }
 
+pub trait PointCloud: Sync {
+    type Id: ToString + Send;
+    type PointIter: Iterator<Item = Point>;
+    fn nodes_in_location(&self, query: &PointQuery) -> Vec<Self::Id>;
```
feuerste: Would it make sense to return an iterator here as well?

nnmm: Sure

nnmm: Then I have to specialize NodeIdsIterator too

nnmm:
ptal @krzesi-mir – Marco and Caterina are out :)

src/batch_iterator.rs
nnmm: Actually, it becomes ugly because NodeIdsIterator borrows Octree, so you'd need to write `type PointIter = NodeIdsIterator<'a>`, but then you'd need to declare `'a` in the trait definition. Or use GATs, but they seem very unstable. I'd rather leave it like this for now.

feuerste: 👍 Alright then, let's leave it as it is...

src/octree/octree_iterator.rs
```diff
@@ -66,32 +66,30 @@ fn get_node_iterator(octree: &Octree, node_id: &NodeId) -> NodeIterator<RawNodeR
 }
 
 /// iterator over the points of a octree node that satisfy the condition expressed by a boolean function
-pub struct FilteredPointsIterator<F> {
-    filter_func: F,
+pub struct FilteredPointsIterator {
+    culling: Box<dyn PointCulling<f64>>,
```
catevita: this means that the filtering now is restricted only by culling/geolocation and not by any other function (idk..eg all points white, all points with intensity > x..)

nnmm: For general filtering, there is `filter()`, so I don't think we need to have that here. This change is needed because you cannot name the type of a `FilteredPointsIterator<F>` when writing the instance of `PointCloud`.

src/octree/octree_iterator.rs
```diff
@@ -66,32 +66,30 @@ fn get_node_iterator(octree: &Octree, node_id: &NodeId) -> NodeIterator<RawNodeR
...
 
-impl<F> Iterator for FilteredPointsIterator<F>
-where
-    F: Fn(&Vector3<f64>) -> bool,
-{
+impl Iterator for FilteredPointsIterator {
     type Item = Point;
 
     fn next(&mut self) -> Option<Point> {
         while let Some(point) = self.node_iterator.next() {
```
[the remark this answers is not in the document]
nnmm: Good point, changed it.

point_viewer_grpc/src/bin/octree_benchmark.rs
```diff
@@ -100,8 +99,9 @@ fn server_benchmark(
         location: PointLocation::AllPoints(),
         global_from_local: None,
     };
+    let octree_slice: &[Octree] = std::slice::from_ref(&octree);
     let mut batch_iterator = BatchIterator::new(
-        std::slice::from_ref(&octree),
+        octree_slice,
```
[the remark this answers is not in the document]
nnmm: In that case, I'd leave it like this.

nnmm:
:dash:

nnmm:
@wally-the-cartographer merge

nnmm:
@wally-the-cartographer merge

PR #317, opened 2019-08-16
Add u8 to attribute data.

feuerste:
This is needed to integrate labels from gray level segmentation images into attributes.


nnmm:
There might be a few more places where it would be good to add u8, compare https://github.com/googlecartographer/point_cloud_viewer/pull/295/files

feuerste:
@wally-the-cartographer merge

PR #318, opened 2019-08-20
Also increase ulmit for X-Ray generation

SirVer:
This should fix AVMAPPING-4886.


SirVer:
@wally-the-cartographer merge

PR #320, opened 2019-08-21
Fix X-Ray wrongly being partially transparent.

SirVer:
This bug was introduced in #198. When data is available for an X-Ray pixel, it should not be partially transparent, but instead fully opaque.


SirVer:
@wally-the-cartographer merge

PR #321, opened 2019-09-03
Make SDL viewer be quieter

nnmm:
Printing on every tick when moving makes it very hard to read any other output.


nnmm:
ptal @SirVer 
I do not find this console output particularly valuable, let me know if you do.

nnmm:
@wally-the-cartographer merge

PR #322, opened 2019-09-03
Implement a read_batch() method for RawNodeReader

nnmm:
Lots of duplicated code for now, will be structured better by introducing reading traits.

### DIFF
--- a/src/read_write/mod.rs
+++ b/src/read_write/mod.rs
@@ -39,6 +39,13 @@ pub use self::raw::{RawNodeReader, RawNodeWriter};
 mod s2;
 pub use self::s2::S2Splitter;
 
+use std::io::{BufReader, Read};
+
+pub struct AttributeReader {
+    pub dtype: crate::AttributeDataType,
+    pub reader: BufReader<Box<dyn Read>>,
+}
+
 /// We open a lot of files during our work. Sometimes users see errors with 'cannot open more
 /// files'. This utility function attempt to increase the rlimits for the number of open files per
 /// process here, but fails silently if we are not successful.
--- a/src/read_write/mod.rs
+++ b/src/read_write/mod.rs
@@ -42,7 +42,7 @@ pub use self::s2::S2Splitter;
 use std::io::{BufReader, Read};
 
 pub struct AttributeReader {
-    pub dtype: crate::AttributeDataType,
+    pub data_type: crate::AttributeDataType,
     pub reader: BufReader<Box<dyn Read>>,
 }
 
--- a/src/read_write/raw.rs
+++ b/src/read_write/raw.rs
@@ -18,14 +18,12 @@ use crate::read_write::{
     decode, fixpoint_decode, AttributeReader, DataWriter, Encoding, NodeReader, NodeWriter,
     OpenMode, PositionEncoding, WriteEncoded, WriteLE,
 };
-use crate::{
-    attribute_extension, AttributeData, AttributeDataType, Point, PointsBatch, NUM_POINTS_PER_BATCH,
-};
+use crate::{attribute_extension, AttributeData, AttributeDataType, Point, PointsBatch};
 use byteorder::{LittleEndian, ReadBytesExt};
 use cgmath::Vector3;
 use num_traits::identities::Zero;
 use std::collections::{BTreeMap, HashMap};
-use std::io::{self, BufReader, Read};
+use std::io::{self, BufReader, ErrorKind, Read};
 use std::path::PathBuf;
 
 pub struct RawNodeReader {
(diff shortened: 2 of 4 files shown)


nnmm:
ptal @feuerste 
fyi @catevita

nnmm:
Should be ready for review now.

src/read_write/raw.rs
```diff
@@ -107,45 +109,228 @@ impl NodeReader for RawNodeReader {
...
+            point.intensity = Some(ir.reader.read_f32::<LittleEndian>()?);
         }
 
         Ok(point)
     }
 }
 
+impl RawNodeReader {
+    #[allow(dead_code)]
+    fn read_batch(&mut self) -> Option<PointsBatch> {
```
feuerste: How about generalizing `NodeReader` over the return type `<P>` instead and having two implementations of `read`, one for `Point` and one for `PointsBatch`?
Also, returning a result instead of an option may be more appropriate in case the underlying data is corrupt...

src/read_write/raw.rs
```diff
@@ -107,45 +109,228 @@ impl NodeReader for RawNodeReader {
...
+
+        let _err: io::Result<()> = match self.encoding {
+            Encoding::Plain => (0..NUM_POINTS_PER_BATCH).try_for_each(|_| {
+                let x = self.xyz_reader.read_f64::<LittleEndian>()?;
+                let y = self.xyz_reader.read_f64::<LittleEndian>()?;
+                let z = self.xyz_reader.read_f64::<LittleEndian>()?;
+                batch.position.push(Vector3::new(x, y, z));
+                Ok(())
+            }),
+            Encoding::ScaledToCube(min, edge_length, PositionEncoding::Uint8) => {
```
feuerste: Similar to above, how about an inner `match` against `position_encoding`?

src/read_write/raw.rs
```diff
@@ -107,45 +109,228 @@ impl NodeReader for RawNodeReader {
...
+            }
+        };
+
+        self.attribute_readers
+            .iter_mut()
+            .for_each(|(key, AttributeReader { dtype, reader })| {
+                match dtype {
+                    AttributeDataType::U8 => {
+                        let mut attr = Vec::with_capacity(NUM_POINTS_PER_BATCH);
+                        let _err: io::Result<()> = (0..NUM_POINTS_PER_BATCH).try_for_each(|_| {
```
feuerste: How about using the `read_x_into` variants of https://docs.rs/byteorder/1.3.2/byteorder/trait.ReadBytesExt.html instead of loops? This should be faster...

src/read_write/raw.rs
```diff
@@ -107,45 +109,228 @@ impl NodeReader for RawNodeReader {
...
+                        batch
+                            .attributes
+                            .insert(key.to_owned(), AttributeData::F64Vec3(attr));
+                    }
+                };
+            });
+
+        let num_points = batch.position.len();
+        if num_points == 0 {
+            return None;
```
feuerste: Does it hurt, if the batch is empty?

src/read_write/raw.rs
```diff
@@ -107,45 +109,228 @@ impl NodeReader for RawNodeReader {
...
+        }
+
+        // If the attributes differ in length, something was wrong with the files.
```
feuerste: See comment above, I would prefer an error...

src/read_write/raw.rs
nnmm: What was the reasoning for doing it this way instead of just two methods in the same trait?

src/read_write/raw.rs
nnmm: Iterators should return None to indicate that iteration has finished, otherwise for_each etc. don't terminate.

nnmm: Okay, it's not an Iterator instance, so that doesn't hold. Hmm.

nnmm:
:+1: Thanks for making all these changes!

src/read_write/raw.rs
nnmm: Result or assert?

src/read_write/raw.rs
feuerste: I agree that the choice of generalization or several functions is a matter of taste, I think the main reason at that time was, that for generalization not all writers had to implement everything, so we could migrate step by step.

src/read_write/raw.rs
feuerste: Error should be fine I think

src/read_write/raw.rs
nnmm: Okay. For positions it's okay to loop for now though?

feuerste: Yeah, for positions and vectors of positions we need traits first, I think.

src/read_write/raw.rs
```diff
@@ -42,19 +42,21 @@ impl NodeReader for RawNodeReader {
...
         // out to be marginally slower.
-        match self.position_encoding {
-            PositionEncoding::Uint8 => {
+        match self.encoding {
+            Encoding::Plain => {
+                point.position.x = self.xyz_reader.read_f64::<LittleEndian>()?;
+                point.position.y = self.xyz_reader.read_f64::<LittleEndian>()?;
+                point.position.z = self.xyz_reader.read_f64::<LittleEndian>()?;
+            }
+            Encoding::ScaledToCube(min, edge_length, PositionEncoding::Uint8) => {
```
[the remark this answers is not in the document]
nnmm: Done.

src/read_write/mod.rs
```diff
@@ -39,6 +39,13 @@ pub use self::raw::{RawNodeReader, RawNodeWriter};
 mod s2;
 pub use self::s2::S2Splitter;
 
+use std::io::{BufReader, Read};
+
+pub struct AttributeReader {
+    pub dtype: crate::AttributeDataType,
```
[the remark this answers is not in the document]
nnmm: Done.

src/read_write/raw.rs
nnmm: Done.

nnmm:
How about the following behavior: The user knows from the meta proto or something the length of the file and requests a suitable number of points from the function. Then, failures to read that many points become an error instead of ignored (ignoring seems like a bad idea).

nnmm:
I did that. PTAL!

PR #324, opened 2019-09-05
Use simpler radial conversion to S2

nnmm:
Do not convert to LatLng and reinterpret, instead just project radially onto the sphere.

The s2_geo module is currently not used. I'd propose to implement intersection logic in math.rs instead. Let me know if you would prefer to keep parts of it.


nnmm:
meow @feuerste

src/read_write/s2.rs
```diff
@@ -40,8 +47,17 @@ where
...
+                let msg = format!(
+                    "Point ({}, {}, {}) is not a valid ECEF point",
+                    pos.x, pos.y, pos.z
+                );
+                return Err(Error::new(ErrorKind::InvalidInput, msg));
+            }
+            let s2_point = Point::from_coords(pos.x, pos.y, pos.z);
             let s2_cell_batch = batches_by_s2_cell
-                .entry(cell_id(ECEFExt::from(*pos), S2_SPLIT_LEVEL))
+                .entry(CellID::from(s2_point).parent(u64::from(S2_SPLIT_LEVEL)))
```
feuerste: Nit: Make the const `S2_SPLIT_LEVEL` above `u64`, so we don't need the conversion

PR #333, opened 2019-09-10
S2 Meta writing prerequisites

nnmm:
* Changed the CellId in the proto to unsigned, since that is what the S2 library uses
* Keep track of attributes & points per node in S2Splitter

### DIFF
--- a/src/read_write/s2.rs
+++ b/src/read_write/s2.rs
@@ -132,6 +132,8 @@ where
         self.writers.get_mut(cell_id).unwrap()
     }
 
+    /// Records the list of attributes seen in the first batch, and checks
+    /// that the following batches contain the same attributes.
     fn check_attributes(&mut self, batch: &PointsBatch) -> Result<()> {
         let mut attr_iter = batch
             .attributes
--- a/src/read_write/s2.rs
+++ b/src/read_write/s2.rs
@@ -1,7 +1,8 @@
-use crate::proto;
 use crate::read_write::{Encoding, NodeWriter, OpenMode};
-use crate::{AttributeData, PointsBatch, CURRENT_VERSION};
+use crate::s2_cells::{S2CellMeta, S2Meta};
+use crate::{AttributeData, AttributeDataType, PointsBatch};
 use cgmath::InnerSpace;
+use fnv::FnvHashMap;
 use lru::LruCache;
 use s2::cellid::CellID;
 use s2::point::Point;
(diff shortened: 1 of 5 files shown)


nnmm:
ptal @catevita @feuerste

src/read_write/s2.rs
```diff
@@ -120,35 +131,37 @@ where
...
-            attr_meta
-        })
-        .collect();
-    s2_meta.set_attributes(::protobuf::RepeatedField::<proto::Attribute>::from_vec(
-        attributes_meta,
-    ));
-    meta.set_s2(s2_meta);
-    meta
-}
+    fn check_attributes(&mut self, batch: &PointsBatch) -> Result<()> {
```
catevita: nit: maybe it is worth to comment this function, it inits with the first batch the attributes and then it checks for the following that the attributes are the same. this will not work in case of sparse attributes in the future

src/read_write/s2.rs
```diff
@@ -120,35 +131,37 @@ where
...
+            .iter()
+            .map(|(name, data)| (name, data.data_type()));
+        if self.attributes_seen.is_empty() {
+            self.attributes_seen
+                .extend(attr_iter.map(|(key, val)| (key.to_owned(), val)));
+            Ok(())
+        } else {
+            attr_iter.try_for_each(|(name, dtype)| {
+                self.attributes_seen
+                    .get(name)
```
catevita: NIT: if a field of the pointsbatch has not be seen yet, it could be added here

src/read_write/s2.rs
```diff
@@ -37,16 +40,21 @@ where
         let writers = LruCache::new(MAX_NUM_NODE_WRITERS);
         let already_opened_writers = HashSet::new();
         let stem = path.into();
+        let cell_stats = FnvHashMap::default();
```
feuerste: Nit: I would stay with the default `HashMap` in the future. Since Rust 1.36 it has a pretty fast implementation, see https://blog.rust-lang.org/2019/07/04/Rust-1.36.0.html#a-new-hashmapk,-v%3E-implementation.

src/s2_cells/mod.rs
```diff
@@ -0,0 +1,52 @@
...
+use std::collections::HashMap;
+use std::convert::TryInto;
+
+#[derive(Copy, Clone)]
+pub struct S2CellMeta {
+    pub num_points: u64,
+}
+
+pub struct S2Meta {
+    pub cells: FnvHashMap<CellID, S2CellMeta>,
```
feuerste: Nit: `HashMap`?

src/read_write/s2.rs
nnmm: Yes, the HashMap algorithm itself is faster, but it still uses SipHash, FnvHashMap uses a different hash function, but the same HashMap algorithm optimized for small keys, so you get the benefits of both. If you prefer I can make this a HashMap though.

src/read_write/s2.rs
nnmm: Good idea, added a comment.

PR #335, opened 2019-09-11
Implement the PointCloud trait for S2Cells

nnmm:
* The point cloud trait aims to contains everything that's needed to function as a backend for the point cloud client (not necessarily the sdl_viewer)
* Beam method is unimplemented
* The point cloud trait is still reading `Point`s, not `PointsBatch`es, TBD


nnmm:
ptal @catevita @feuerste

src/s2_cells/mod.rs
```diff
@@ -50,3 +63,157 @@ impl S2Meta {
...
+}
+
+impl S2Cells {
+    /// Wrapper arround cells_in_convex_hull for Obbs
+    fn cells_in_obb(
+        &self,
+        obb: &Obb<f64>,
+        global_from_local: Option<&Isometry3<f64>>,
+    ) -> Vec<S2CellId> {
+        let obb = match global_from_local {
```
feuerste: Nit: Instead of returning a `Cow` of `obb`, can we just return the `obb.corners` directly as `points`, since the `obb` isn't used elsewhere?

src/s2_cells/mod.rs
```diff
@@ -50,3 +63,157 @@ impl S2Meta {
...
+
+    /// Returns all cells that intersect the convex hull of the given points
+    fn cells_in_convex_hull<IterPoint>(&self, points: IterPoint) -> Vec<S2CellId>
+    where
+        IterPoint: IntoIterator<Item = Point3<f64>>,
+    {
+        // We could choose either a covering rect or a covering cap as a convex hull
+        let point_cells = points
+            .into_iter()
+            .map(|p| CellID::from(S2Point::from_coords(p.x, p.y, p.z)))
```
feuerste: Do we want the cell ids to be in a specific s2 level?

src/s2_cells/mod.rs
```diff
@@ -50,3 +63,157 @@ impl S2Meta {
...
+            .collect();
+        let mut cell_union = CellUnion(point_cells);
+        cell_union.normalize();
```
feuerste: Referring to the comment above, is this what we want? We will potentially have cells in a different levels after this operation....

src/s2_cells/mod.rs
nnmm: `creates a temporary which is freed while still in use` – unless we just clone.

src/s2_cells/mod.rs
nnmm: As fine as possible to not make the approximation unnecessarily conservative, which I assume is what `CellID::from` gives us.

PR #339, opened 2019-09-12
Create AttributeDataType from proto

catevita:


src/lib.rs
```diff
@@ -87,6 +88,25 @@ impl AttributeDataType {
...
+    pub fn from_proto(attr_proto: proto::AttributeDataType) -> Result<Self> {
+        let attr = match attr_proto {
+            proto::AttributeDataType::U8 => AttributeDataType::U8,
+            proto::AttributeDataType::I64 => AttributeDataType::I64,
+            proto::AttributeDataType::U64 => AttributeDataType::U64,
+            proto::AttributeDataType::F32 => AttributeDataType::F32,
+            proto::AttributeDataType::F64 => AttributeDataType::F64,
+            proto::AttributeDataType::U8Vec3 => AttributeDataType::U8Vec3,
+            proto::AttributeDataType::F64Vec3 => AttributeDataType::F64Vec3,
+            _ => {
```
nnmm: What happens if we leave this out? Does it still compile?

catevita: it will complain that not all the types have been covered

nnmm: I see, I thought those were all, but the proto has a few more that we'd have to add to our `AttributeDataType` enum. To get more type safety, how about, we replace the `_` with listing all the proto data types, and still return an `Err` for those? `use proto::AttributeDataType::*` might help there.

PR #340, opened 2019-09-13
Don't write header for empty batches.

feuerste:
Calling `write` with empty points batches led to creation of multiple headers. This PR fixes this.

### DIFF
--- a/src/read_write/ply.rs
+++ b/src/read_write/ply.rs
@@ -465,6 +465,9 @@ impl NodeWriter<PointsBatch> for PlyNodeWriter {
     }
 
     fn write(&mut self, p: &PointsBatch) -> io::Result<()> {
+        if p.position.is_empty() {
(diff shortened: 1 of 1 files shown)


src/read_write/ply.rs
```diff
@@ -465,6 +465,9 @@ impl NodeWriter<PointsBatch> for PlyNodeWriter {
     }
 
     fn write(&mut self, p: &PointsBatch) -> io::Result<()> {
+        if p.position.is_empty() {
```
nnmm: How about `if self.point_count == 0 && !p.position.is_empty()` below?

feuerste: I personally prefer the early break, but do not insist on it :)

nnmm: Okay , it's fine like this too.

PR #341, opened 2019-09-17
S2Meta: Added conversion methods from proto and DataProvider

catevita:
actually the only modifications are from line 73 to 114. I am wondering why the diff is bloated up


catevita:
PTAL @nnmm

src/s2_cells/mod.rs
```diff
@@ -1,221 +1,263 @@
...
+            attributes_meta,
+        ));
+        meta.set_s2(s2_meta);
+        meta
+    }
+
+    pub fn from_proto(meta_proto: proto::Meta) -> Result<Self> {
+        // check if the meta is meant to be for S2 point cloud
+        if !(meta_proto.version == CURRENT_VERSION && meta_proto.has_s2()) {
+            return Err(ErrorKind::InvalidInput("No s2 meta found".to_string()).into());
```
nnmm: How about "Point cloud is not an S2 point cloud"?

nnmm: But this could be confusing if the error actually comes from the version check ... maybe that should have a separate message?

src/s2_cells/mod.rs
```diff
@@ -1,221 +1,263 @@
...
+                cell_id,
+                S2CellMeta {
+                    num_points: cell.num_points,
+                },
+            );
+        });
+
+        // attributes
+        let mut attributes = HashMap::default();
+        // s2_meta_proto.attributes.iter().try_for_each( |attr: proto::Attribute| -> Result<()> {
```
nnmm: Please remove the commented out code

src/s2_cells/mod.rs
```diff
@@ -1,221 +1,263 @@
...
+        //     let d_type: proto::AttributeDataType = attr.get_data_type();
+        //     let attr_type: crate::AttributeDataType = crate::AttributeDataType::from_proto(d_type)?;//
+        //     attributes.insert(
+        //         attr.name,
+        //         attr_type,
+        //     )
+        // })?;
+        for attr in s2_meta_proto.attributes.iter() {
+            let d_type: proto::AttributeDataType = attr.get_data_type();
+            let attr_type: crate::AttributeDataType = crate::AttributeDataType::from_proto(d_type)?; //
```
nnmm: Nits: For consistency, probably better to add AttributeDataType to the `use` above. How about just `let attr_type = AttributeDataType::from_proto(attr.get_data_type())`?

src/s2_cells/mod.rs
```diff
@@ -1,221 +1,263 @@
...
+            attributes.insert(attr.name.to_owned(), attr_type);
+        }
+
+        Ok(S2Meta { cells, attributes })
+    }
+
+    pub fn from_data_provider(data_provider: Box<dyn DataProvider>) -> Result<Self> {
```
nnmm: I think we shouldn't take `data_provider` by value because then you can't use it anywhere else after. How about omething like `from_data_provider(data_provider: &dyn DataProvider)`? You can still use that if the caller has a Box.

catevita: I've just copied the signature in octree/mod.rs, do you think that it is worth to change the signature also there (in a future pr?)

src/s2_cells/mod.rs
```diff
@@ -1,221 +1,268 @@
...
+        ));
+        meta.set_s2(s2_meta);
+        meta
+    }
+
+    pub fn from_proto(meta_proto: proto::Meta) -> Result<Self> {
+        // check if the meta is meant to be for S2 point cloud
+        match meta_proto.version {
+            CURRENT_VERSION => {
+                if !(meta_proto.version == CURRENT_VERSION && meta_proto.has_s2()) {
```
nnmm: I'd do it without nesting:
```
if meta_proto.version != CURRENT_VERSION {
    return ... format!("Point cloud has unsupported version {}", meta_proto.version) ...
}
if !meta_proto.has_s2() {
    return Err(ErrorKind::InvalidInput("Not an S2 point cloud"));
}
```

src/s2_cells/mod.rs
nnmm: I think it's fine there, because the DataProvider gets moved into the octree, but we just need a reference.

nnmm:
LGTM with two small nits:
* import AttributeDataType
* Capitalize S2 in error messages but lowercase "point cloud"

catevita:
@nnmm fixed! I need the explicit approval

PR #342, opened 2019-09-17
Extract data from attribute data.

feuerste:
This allows to extract the underlying data vector from the `AttributeData` enum.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -212,6 +212,22 @@ pub struct PointsBatch {
     pub attributes: BTreeMap<String, AttributeData>,
 }
 
+impl PointsBatch {
+    pub fn attribute_vec<'a, T>(&'a self, key: impl AsRef<str>) -> Option<&'a Vec<T>>
+    where
+        Option<&'a Vec<T>>: From<&'a AttributeData>,
+    {
+        self.attributes.get(key.as_ref()).and_then(Option::from)
+    }
+
+    pub fn attribute_vec_mut<'a, T>(&'a mut self, key: impl AsRef<str>) -> Option<&'a mut Vec<T>>
+    where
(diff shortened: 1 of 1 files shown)


nnmm:
Nice! What do you think about named functions vs From vs TryFrom?
With named functions I think you might have to think a little less about how it works, and not need type annotations. But you can't write generic code that uses them without macros. TryFrom could have a message about not having the expected format.

feuerste:
Thinking about it a little more, I personally don't like named functions so much, but `TryFrom` because of its error capabilities. Fine with you, if we use `TryFrom`?

feuerste:
I would also remove the non-ref version, since it's not currently used anyway...

nnmm:
Yep, I'm fine with `TryFrom`. Personally I think it's not hard to imagine the by-value version being used, but we can of course re-add it when needed.

feuerste:
`TryFrom` turned out to be not as trivial as expected. For
```rust
impl PointsBatch {
    pub fn attribute_vec<'a, T>(
        &'a self,
        key: impl AsRef<str>,
    ) -> std::result::Result<&'a Vec<T>, &'a str>
    where
        &'a Vec<T>: TryFrom<&'a AttributeData>,
    {
        let val = self.attributes.get(key.as_ref()).ok_or_else(|| "")?;
        <&'a Vec::<T>>::try_from(val)
    }
}
```
we get
```
expected type `std::result::Result<_, &'a str>`
found type `std::result::Result<_, <&'a std::vec::Vec<T> as std::convert::TryFrom<&'a AttributeData>>::Error>`
```
errors, which would only be resolvable by separating `get` from `try_from` with different result types. I kind of like the current behavior though.

PR #343, opened 2019-09-19
Increase annoyingly low rotation speed in sdl_viewer

nnmm:
This only affects rotating the camera in sdl_viewer with the keyboard, not with the mouse.


nnmm:
meow @catevita

PR #344, opened 2019-09-19
Integration test for end-to-end comparison of octrees and S2 point clouds

nnmm:
Right now, this only includes creation of two point clouds from the same points. It will be extended to cover querying as well.


nnmm:
meow @catevita @feuerste

src/bin/generate_and_query/main.rs
```diff
@@ -0,0 +1,80 @@
...
+use random_points::{Batched, RandomPointsOnEarth};
+use std::path::PathBuf;
+use std::time::Instant;
+use structopt::StructOpt;
+
+mod random_points;
+
+#[derive(StructOpt, Debug)]
+struct CommandlineArguments {
+    /// Output directory to write the octree into.
```
catevita: maybe change to " write the point cloud" (as s2 and octree are subdirectories of that)

src/bin/generate_and_query/random_points.rs
```diff
@@ -0,0 +1,162 @@
...
+    }
+
+    pub fn bbox(&mut self) -> Aabb3<f64> {
+        let local_min = Point3::new(-self.half_width, -self.half_width, -self.half_height);
+        let local_max = Point3::new(self.half_width, self.half_width, self.half_height);
+        // take center of global bbox as initial value for min and max corner
+        let mut ecef_min = Point3::from_vec(self.ecef_from_local.w.truncate());
+        let mut ecef_max = Point3::from_vec(self.ecef_from_local.w.truncate());
+        Aabb3::new(local_min, local_max)
+            .to_corners()
```
feuerste: How about using https://docs.rs/collision/0.20.1/collision/trait.Aabb.html#tymethod.transform instead?

src/bin/generate_and_query/random_points.rs
```diff
@@ -0,0 +1,162 @@
...
+        Matrix4::from(frame)
+    }
+}
+
+impl Iterator for RandomPointsOnEarth {
+    type Item = Point;
+
+    fn next(&mut self) -> Option<Point> {
+        let pos = self.next_pos();
+        // println!("pos {:?}, bbox {:?}", pos, self.bbox());
```
feuerste: Nit: Remove comment.

catevita:
should this binary land in some sort of /benches directory for benchmarking?

catevita:
maybe we could then move there all the other kind of benchmarks

nnmm:
I'm a bit unsure whether we should declare it as a benchmark ... wouldn't that just benchmark the execution time of the whole binary, instead of octree/s2 building separately?

nnmm:
Yes, that's the main purpose. Does `tests` get special treatment by cargo?

feuerste:
See https://doc.rust-lang.org/cargo/reference/manifest.html#integration-tests for a bit of explanation.

nnmm:
Okay, sounds like it can not have command line arguments then, but I think it makes sense to make it an integration test.

nnmm:
Changed it, ptal. However, now there's no way to modify the flags anymore. `cargo test generate_and_query -- --help` prints some unrelated help text four times for me.

feuerste:
> Changed it, ptal. However, now there's no way to modify the flags anymore. `cargo test generate_and_query -- --help` prints some unrelated help text four times for me.

Ok, shall we then just remove all options and make them `const`s?

nnmm:
Yep, we can do that for now. On thing I don't know how to handle within an integration test though is benchmarking. I'd like to benchmark multiple parts later: Both octree and S2 generation, and different kinds of queries. How can we avoid repeating the setup (creating the random pointclouds) for the latter?

feuerste:
https://doc.rust-lang.org/unstable-book/language-features/custom-test-frameworks.html unfortunately doesn't seem to be available yet, but how about having the common setup in a `lazy_static`?

nnmm:
Benchmark seem to not work at all inside `tests/`, only in `src/` and `benches/`, so I moved it to `benches`.
I'm using TempDir for cleanup, and lazy_static to avoid regenerating things for tests (for the benchmark, regenerating it is of course intentional).

src/math.rs
```diff
@@ -554,3 +555,26 @@ where
...
+// https://en.wikipedia.org/wiki/Geographic_coordinate_conversion#From_ECEF_to_ENU
+pub fn local_frame_from_lat_lng(lat: f64, lon: f64) -> Matrix4<f64> {
+    const PI_HALF: Deg<f64> = Deg(90.0);
+    let lat_lng_alt = WGS84::new(lat, lon, 0.0);
+    let origin = ECEF::from(lat_lng_alt);
+    let origin_vector = Vector3::new(origin.x(), origin.y(), origin.z());
+    println!("origin_vector {:?}", origin_vector);
+    let rotation_matrix = Matrix3::from_angle_z(-PI_HALF)
+        * Matrix3::from_angle_y(Deg(lat_lng_alt.latitude_degrees()) - PI_HALF)
+        * Matrix3::from_angle_z(Deg(-lat_lng_alt.longitude_degrees()));
```
catevita: I still have to understand this rotation thoroughly.. usually rot z, rot y and rot x are applied. from ECEF to ENU this font explains that two rotations suffice to get from ecef to enu and backwards. https://gssc.esa.int/navipedia/index.php/Transformations_between_ECEF_and_ENU_coordinates

benches/generate_and_query/main.rs
```diff
@@ -0,0 +1,138 @@
...
+use test::Bencher;
+
+mod random_points;
+
+#[derive(Debug)]
+struct Arguments {
+    // Minimal precision that this point cloud should have.
+    // This decides on the number of bits used to encode each node.
+    resolution: f64,
+    // Width of the area to generate points for (used for local x and y), in meters
```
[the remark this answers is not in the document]
nnmm: Done

PR #346, opened 2019-09-23
Return position with attributes from S2Meta

catevita:
- [x] return position and attributes meta in one iterator
- [x] prevent accidental modification of s2 meta by making struct members private
- [x] fixing some clippy warnings for AttributeDataType

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -65,7 +65,7 @@ pub enum AttributeDataType {
 }
 
 impl AttributeDataType {
-    pub fn to_proto(&self) -> proto::AttributeDataType {
+    pub fn to_proto(self) -> proto::AttributeDataType {
         match self {
             AttributeDataType::U8 => proto::AttributeDataType::U8,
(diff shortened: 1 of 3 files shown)


catevita:
PTAL @feuerste fyi @nnmm

src/lib.rs
```diff
@@ -105,7 +105,7 @@ impl AttributeDataType {
         Ok(attr)
     }
 
-    pub fn size_of(&self) -> usize {
+    pub fn size_of(self) -> usize {
```
feuerste: Why consume `self`? We just want to know its size...

feuerste: Ah, that's because above `Copy` was introduced. What do we need `Copy` for?

nnmm: I don't think it was needed per se, but for small structs, it's a small convenience and efficiency win with no downsides.

catevita: Copy was added for the attribute+xyz iterator returned from S2Meta

PR #347, opened 2019-09-24
Tiled texture loader for terrain rendering

nnmm:
This is pretty simple right now, it keeps all the tiles in memory and stitches them together to an image for a given ROI.

### DIFF
--- a/sdl_viewer/src/graphic/tiled_texture_loader.rs
+++ b/sdl_viewer/src/graphic/tiled_texture_loader.rs
@@ -1,0 +1,107 @@
+use byteorder::{LittleEndian, ReadBytesExt};
+use image::{GenericImage, GenericImageView, ImageBuffer, LumaA, Pixel, Rgba};
+use num_integer::Integer;
+use std::collections::HashMap;
+use std::fs::File;
+use std::io::BufReader;
+use std::path::Path;
+
+/// Little trait that just dispatches to the correct read_xyz_into call for the image type
+pub trait ReadLittleEndian {
+    fn read_from<R: ReadBytesExt>(&mut self, rdr: &mut R) -> Result<(), std::io::Error>;
+}
+
+impl ReadLittleEndian for ImageBuffer<LumaA<f32>, Vec<f32>> {
+    fn read_from<R: ReadBytesExt>(&mut self, rdr: &mut R) -> Result<(), std::io::Error> {
+        rdr.read_f32_into::<LittleEndian>(self)
+    }
+}
+
+impl ReadLittleEndian for ImageBuffer<Rgba<u8>, Vec<u8>> {
+    fn read_from<R: ReadBytesExt>(&mut self, rdr: &mut R) -> Result<(), std::io::Error> {
+        rdr.read_exact(self)
+    }
+}
+
+type TilePos = (i32, i32);
+
+pub struct TiledTextureLoader<P: Pixel> {
+    tile_size: u32,
+    tiles: HashMap<TilePos, ImageBuffer<P, Vec<P::Subpixel>>>,
+}
+
+impl<P> TiledTextureLoader<P>
+where
+    P: Pixel + 'static,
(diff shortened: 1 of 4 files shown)


nnmm:
meow @krzesi-mir

sdl_viewer/src/graphic/tiled_texture_loader.rs
```diff
@@ -0,0 +1,107 @@
...
+}
+
+impl<P> TiledTextureLoader<P>
+where
+    P: Pixel + 'static,
+    ImageBuffer<P, Vec<P::Subpixel>>: ReadLittleEndian,
+{
+    pub fn new<I, Q>(tile_size: u32, tile_iter: I) -> Result<Self, std::io::Error>
+    where
+        I: Iterator<Item = ((i32, i32), Q)>,
```
krzesi-mir: nit: use `TilePos` instead of `(i32, i32)`?

sdl_viewer/src/graphic/tiled_texture_loader.rs
```diff
@@ -0,0 +1,107 @@
...
+        let (min_tile_x, min_mod_x) = min_x.div_mod_floor(ts);
+        let (min_tile_y, min_mod_y) = min_y.div_mod_floor(ts);
+        // Same for the upper right corner
+        let (max_tile_x, max_mod_x) = max_x.div_mod_floor(ts);
+        let (max_tile_y, max_mod_y) = max_y.div_mod_floor(ts);
+        let mut output_buffer = ImageBuffer::new(width as u32, height as u32);
+        for tile_x in min_tile_x..=max_tile_x {
+            for tile_y in min_tile_y..=max_tile_y {
+                // At the left border, start with the correct offset
+                let x_off_src = if tile_x == min_tile_x { min_mod_x } else { 0 };
```
krzesi-mir: do you think it would make sense to add a new function/method/closure, for example `get_offset_and_length`, to avoid code duplication?

sdl_viewer/src/graphic/tiled_texture_loader.rs
```diff
@@ -0,0 +1,107 @@
...
+            assert!(previous_entry.is_none());
+        }
+
+        Ok(TiledTextureLoader { tile_size, tiles })
+    }
+
+    /// Loads the specified region of the sparse texture into a ImageBuffer
+    pub fn load(
+        &self,
+        min_x: i64,
```
krzesi-mir: Wouldn't it make sense for `min_x` and `min_y` to have the same type as the indices of `self.tiles`? I.e. either all of them being `i32` or all of them `i64`. Also, do you allow negative numbers or could we use unsigned integers here?

sdl_viewer/src/graphic/tiled_texture_loader.rs
nnmm: I don't know, it would need many variables to be passed: `tile_x`, `min_tile_x`, `min_mod_x`, `max_tile_x`, `max_mod_x`. I think it would be a little easier to read when we leave it like this.

nnmm: But no strong opinions

krzesi-mir: Yeah, I agree, it might be too much.

sdl_viewer/src/graphic/tiled_texture_loader.rs
nnmm: Negative positions are allowed, but we could make all of them `i64` for sure.

sdl_viewer/src/graphic/tiled_texture_loader.rs
nnmm: Yep

nnmm:
Should I ask Marco to approve?

krzesi-mir:
:+1:

krzesi-mir:
lgtm

PR #348, opened 2019-09-24
Implement nodes querying for OrientedBeam.

feuerste:
This solution simply intersects the oriented beam with the min and max earth radius to create an OBB and treats the query as an OBB query.

### DIFF
--- a/src/math.rs
+++ b/src/math.rs
@@ -250,6 +250,7 @@ impl<S: BaseFloat> Isometry3<S> {
 
 #[derive(Debug, Clone)]
 pub struct Obb<S> {
+    pub isometry: Isometry3<S>,
     pub isometry_inv: Isometry3<S>,
     pub half_extent: Vector3<S>,
     pub corners: [Point3<S>; 8],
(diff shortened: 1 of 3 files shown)


src/math.rs
```diff
@@ -260,10 +267,27 @@ impl<S: BaseFloat> From<&Aabb3<S>> for Obb<S> {
...
-            Isometry3::new(Quaternion::one(), EuclideanSpace::to_vec(aabb.center())),
-            aabb.dim() / (S::one() + S::one()),
-        )
+        Self::from(&aabb)
+    }
+}
+
+impl<S: BaseFloat> From<&OrientedBeam<S>> for Obb<S> {
+    fn from(beam: &OrientedBeam<S>) -> Self {
+        let beam_isometry = beam.isometry_inv.inverse();
```
nnmm: Nit: Is this maybe worth putting in a TODO to get rid of the dual inversion?

src/math.rs
```diff
@@ -260,10 +267,27 @@ impl<S: BaseFloat> From<&Aabb3<S>> for Obb<S> {
...
+        let earth_radius_mean = S::from(0.5 * (EARTH_RADIUS_MIN_M + EARTH_RADIUS_MAX_M)).unwrap();
+        let z_off = earth_radius_mean - beam_isometry.translation.magnitude();
+        let pt_off = Point3::new(S::zero(), S::zero(), z_off);
+        let translation = beam_isometry.rotation.rotate_point(pt_off) + beam_isometry.translation;
```
nnmm: Nit: If you use a Vector3 in the line above, that's just multiplication with the beam isometry

feuerste:
Thanks for the review!

PR #349, opened 2019-09-24
Rename and reorganize iterators.

feuerste:
This is a small preparation step for introducing iterators over points batches. Changes are

- `BatchIterator` is renamed to `ParallelIterator`
- `NodeIterator` is renamed to `PointIterator`
- `FilteredPointsIterator` is renamed to `FilteredIterator`
- the `NodeReader` trait is removed, since it was only used once


feuerste:
Thanks for the review!

PR #350, opened 2019-09-25
Implement ply batch iterator.

feuerste:
Timings on lego small for building an octree:
Old: 0m34.951s
New: 2m36.307s
However, when additionally taking #356 into account, we get the following measurements:
New: 0m39.238s
So the strategy for merging would be to merge this one and #356 right after each other, so I need approvals for both of them.


src/read_write/ply.rs
```diff
@@ -315,83 +331,134 @@ impl PlyIterator {
...
                 }
-                other => {
-                    println!("Will ignore property '{}' on 'vertex'.", other);
+                _ => {
                     use self::DataType::*;
                     match prop.data_type {
-                        Uint8 | Int8 => readers.push(create_skip_fn!(&mut num_bytes_per_point, 1)),
-                        Uint16 | Int16 => {
-                            readers.push(create_skip_fn!(&mut num_bytes_per_point, 2))
+                        Uint8 => {
```
catevita: does the ply now also support more than color, position and intensity? Great!

nnmm:
How is the octree generation going to work with batches?

src/read_write/ply.rs
feuerste: Yes, it reads everything we can put into our `AttributeData`. I didn't implement multidimensional attributes other than color and position though.

feuerste:
> How is the octree generation going to work with batches?

As far as I saw it, the iterator is used at two places:

1. Bounding box generation, which is pretty easy to change from points to batches
2. Subsampling of children: Here all points are read into memory anyway. So we need to add functionality to merge batches and then split up batches modulo 8, which should be not too difficult.

src/read_write/ply.rs
```diff
@@ -222,25 +223,25 @@ fn parse_header<R: BufRead>(reader: &mut R) -> Result<(Header, usize)> {
     ))
 }
 
-type ReadingFn = fn(nread: &mut usize, buf: &[u8], val: &mut Point);
+type ReadingFn = fn(nread: &mut usize, read_buf: &[u8], write_buf: &mut [u8], write_pos: usize);
 
 // The two macros create a 'ReadingFn' that reads a value of '$data_type' out of a reader, and
```
nnmm: Is this comment still up to date? There is no `$data_type` in the code.

src/read_write/ply.rs
```diff
@@ -273,29 +274,57 @@ macro_rules! read_casted_property {
     };
 }
 
+macro_rules! push_reader {
```
nnmm: Is `create_and_return_reading_fn!` used exclusively by `read_casted_property!`, and is `read_casted_property!` used exclusively by `push_reader!`? If so, why shouldn't we combine them into one macro?

feuerste: We can combine `push_reader` and `read_casted_property`, but the other doesn't make too much sense, as `create_and_return_reading_fn` is used 10 times inside `read_casted_property`, which would blow up the code...

src/read_write/ply.rs
```diff
@@ -408,48 +407,117 @@ impl PlyIterator {
...
+) -> PointsBatch {
+    let (mut x_vec, mut y_vec, mut z_vec) = (Vec::new(), Vec::new(), Vec::new());
+    let (mut r_vec, mut g_vec, mut b_vec) = (Vec::new(), Vec::new(), Vec::new());
+    let mut attributes = BTreeMap::new();
+    for reader in readers {
+        match &reader.prop.name as &str {
+            "x" => x_vec = to_vec::<f64>(&reader.buf, batch_size),
+            "y" => y_vec = to_vec::<f64>(&reader.buf, batch_size),
+            "z" => z_vec = to_vec::<f64>(&reader.buf, batch_size),
+            "r" | "red" => r_vec = reader.buf[..batch_size].to_vec(),
```
nnmm: Can we read other "vector" attributes than color and position? If so, where is that defined?

nnmm: Ah I see, you answered that above.

nnmm: Is that a TODO? If so, consider adding a TODO comment.

src/read_write/ply.rs
```diff
@@ -408,48 +407,117 @@ impl PlyIterator {
...
         })
     }
 }
 
+fn to_vec<T>(slice: &[u8], cur_batch_size: usize) -> Vec<T>
+where
+    T: Clone,
+{
+    let ptr = slice.as_ptr() as *const T;
+    unsafe { std::slice::from_raw_parts(ptr, cur_batch_size).to_vec() }
```
nnmm: What do you think about omitting a copy here by using https://doc.rust-lang.org/std/vec/struct.Vec.html#method.from_raw_parts and std::mem::replace?

src/read_write/ply.rs
```diff
@@ -222,25 +223,25 @@ fn parse_header<R: BufRead>(reader: &mut R) -> Result<(Header, usize)> {
...
+            $assign(write_buf, write_pos, $reading_fn(read_buf) as _);
             *nread += $num_bytes;
         }
     }};
 }
 
 macro_rules! read_casted_property {
-    ($data_type:expr, $assign:expr, &mut $size:ident) => {
-        match $data_type {
+    ($prop:expr, $assign:expr, &mut $size:ident) => {
```
nnmm: What's the advantage of changing this?

nnmm:
At a high level, couldn't the `buf` inside `PropertyReader` be an `AttributeData` of the right type? Then we wouldn't need to cast and copy.

Maybe we don't even need all those macros then?

src/read_write/ply.rs
feuerste: I changed it back to `$data_type`

src/read_write/ply.rs
feuerste: Done.

src/read_write/ply.rs
feuerste: I changed the implementation to use `AttributeData`, so we don't need any `unsafe` methods any more.

src/read_write/ply.rs
feuerste: I've overseen it, changed it back.

feuerste:
I changed everything to work on `AttributeData`, so no vector casting is needed any more. I think the macros are still needed because of the simultaneous assignment and casting while reading, as mentioned in the comment above `create_and_return_reading_fn`.

nnmm:
I think we should be able to get rid of the match in `read_casted_property!` (at which point we can probably remove that macro). In `push_reader!` we have `$dtype`, from which we can construct the `LittleEndian::read_xyz` function name and get the `$num_bytes` by using `std::mem::size_of::<xyz>()`.

feuerste:
Please be aware that `$dtype` is not equal the type we read in `LittleEndian::read_*` for x, y, and z. Furthermore, there is no `LittleEndian::read_u8` function.

feuerste:
Thanks for your reviews!

PR #351, opened 2019-09-26
Integration testing: Add querying test

nnmm:
* Adds the `from_data_provider()` method, the last missing piece to read S2 point clouds.
* The test is currently failing and marked ignored until we fix querying.


nnmm:
Please review, @catevita @feuerste

benches/generate_and_query/main.rs
```diff
@@ -136,3 +156,58 @@ fn num_points_in_s2() {
     let num_points: u64 = meta.get_s2().get_cells().iter().map(|c| c.num_points).sum();
     assert_eq!(num_points, Arguments::default().num_points as u64);
 }
+
+fn query_and_sort<C>(point_cloud: &C, query: &PointQuery) -> Vec<Point>
```
feuerste: How about working on batches of points and use the `ParallelIterator` with `num_threads = 1` internally? Of course we would need a `sort_unstable_by` implementation for `PointsBatch`, but it would be easy with https://github.com/jeremysalwen/rust-permutations.

nnmm: The advantage of this would be that it's more end-to-end? The sort impl would be unneeded if we just return positions, right?

nnmm: Actually, I think we should provide both methods. Then we can compare/benchmark between parallel iterator and direct point fetching.

PR #353, opened 2019-09-27
sdl_viewer: Terrain metadata deserialization

nnmm:
The terrain is represented as a set of tiles, each saved in one file. This is a simple binary format for metadata including the coordinate frame of the terrain, resolution, and tile size.

An alternative would be to use serde_json which is already a dependency, however we can't derive `Serialize`/`Deserialize` because of the `Isometry3` and `Vector3`, and versioning is probably not as easy.


nnmm:
ptal @krzesi-mir

sdl_viewer/src/terrain_drawer/read_write.rs
```diff
@@ -0,0 +1,56 @@
...
+
+pub type TextureLoaders = (TiledTextureLoader<LumaA<f32>>, TiledTextureLoader<Rgba<u8>>);
+
+#[allow(dead_code)]
+impl Metadata {
+    // Custom serialization of some values
+    pub fn from_dir<P: AsRef<std::path::Path>>(dir: P) -> Result<Self, Error> {
+        let meta_path = dir.as_ref().join("meta.json");
+        let reader = File::open(meta_path)?;
+        let meta = serde_json::from_reader(reader).map_err(|e| {
```
krzesi-mir: nit: return here instead of using an intermediate variable `meta`?

sdl_viewer/src/terrain_drawer/read_write.rs
```diff
@@ -0,0 +1,56 @@
...
+            let msg = format!("Could not parse meta.json: {}", e.to_string());
+            Error::new(ErrorKind::InvalidData, msg)
+        })?;
+        Ok(meta)
+    }
+
+    pub fn read_tiles<P: AsRef<std::path::Path>>(&self, dir: P) -> Result<TextureLoaders, Error> {
+        let height_tiles = self.tile_positions.iter().map(|xy| {
```
krzesi-mir: nit: do `.map(|(x, y)| { ... ` instead of `.map(|xy| { ...` so we don't need line `35`?

sdl_viewer/src/terrain_drawer/read_write.rs
```diff
@@ -0,0 +1,56 @@
...
+            let (x, y) = xy;
+            let height_file_path = dir.as_ref().join(format!("x{:08}_y{:08}.height", x, y));
+            (*xy, height_file_path)
+        });
+        let color_tiles = self.tile_positions.iter().map(|xy| {
+            let (x, y) = xy;
```
krzesi-mir: nit: same as the comment above

sdl_viewer/src/terrain_drawer/read_write.rs
```diff
@@ -0,0 +1,56 @@
...
+    pub tile_size: u32,
+    pub world_from_terrain: Isometry3<f64>,
+    pub origin: Vector3<f64>,
+    pub resolution_m: f64,
+    pub tile_positions: Vec<TilePos>,
+}
+
+pub type TextureLoaders = (TiledTextureLoader<LumaA<f32>>, TiledTextureLoader<Rgba<u8>>);
+
+#[allow(dead_code)]
```
krzesi-mir: Do we need this attribute?

sdl_viewer/src/terrain_drawer/read_write.rs
```diff
@@ -0,0 +1,56 @@
...
+impl Metadata {
+    // Custom serialization of some values
+    pub fn from_dir<P: AsRef<std::path::Path>>(dir: P) -> Result<Self, Error> {
```
krzesi-mir: nit: use `std::io::Result`?

sdl_viewer/src/terrain_drawer/read_write.rs
```diff
@@ -0,0 +1,56 @@
...
+            let (x, y) = xy;
+            let color_file_path = dir.as_ref().join(format!("x{:08}_y{:08}.color", x, y));
+            (*xy, color_file_path)
+        });
+        let height_tiles = TiledTextureLoader::new(self.tile_size, height_tiles)?;
+        let color_tiles = TiledTextureLoader::new(self.tile_size, color_tiles)?;
+        Ok((height_tiles, color_tiles))
+    }
+
+    pub fn write<P: AsRef<std::path::Path>>(&self, dir: P) -> Result<(), Error> {
```
krzesi-mir: nit: use `std::io::Result`?

sdl_viewer/src/terrain_drawer/read_write.rs
```diff
@@ -0,0 +1,56 @@
...
+        let meta_path = dir.as_ref().join("meta.json");
+        let reader = File::open(meta_path)?;
+        let meta = serde_json::from_reader(reader).map_err(|e| {
+            let msg = format!("Could not parse meta.json: {}", e.to_string());
+            Error::new(ErrorKind::InvalidData, msg)
+        })?;
+        Ok(meta)
+    }
+
+    pub fn read_tiles<P: AsRef<std::path::Path>>(&self, dir: P) -> Result<TextureLoaders, Error> {
```
krzesi-mir: nit: use `std::io::Result`?

sdl_viewer/src/terrain_drawer/read_write.rs
nnmm: Yeah, for now. Will be removed in the next PR.

sdl_viewer/src/terrain_drawer/read_write.rs
nnmm: Don't you think the parenthesizing would hurt readibility a bit if we wrap it in a Ok()?

sdl_viewer/src/terrain_drawer/read_write.rs
nnmm: Done

sdl_viewer/src/terrain_drawer/read_write.rs
nnmm: Done

sdl_viewer/src/terrain_drawer/read_write.rs
krzesi-mir: `serde_json::from_reader` returns a `Result` so I don't think we have to wrap in in `Ok`, or do we?

nnmm: Oh yeah

nnmm:
@feuerste Can I get an approval?

nnmm:
Thanks!

PR #354, opened 2019-09-27
sdl_viewer: Single layer of a terrain drawer

nnmm:
This struct encapsulates a terrain texture that gets updated as the camera moves. That is, it ties together loading incremental texture data (`TiledTextureLoader`) and updating it on the OpenGL side (`GlMovingWindowTexture`).

A terrain drawer might involve several layers – for visualizing several patches of terrain at once, or multilevel terrain (overpasses). Hence, this struct also includes uniforms that are set on a per-layer basis: each layer might have a different frame, resolution, etc.


nnmm:
ptal @krzesi-mir

sdl_viewer/src/terrain_drawer/layer.rs
```diff
@@ -0,0 +1,232 @@
...
+use crate::graphic::uniform::GlUniform;
+use crate::graphic::GlProgram;
+use crate::terrain_drawer::read_write::Metadata;
+use cgmath::{Decomposed, Matrix4, Vector2, Vector3, Zero};
+use image::{ImageBuffer, LumaA, Rgba};
+use point_viewer::math::Isometry3;
+use std::convert::TryInto;
+use std::rc::Rc;
+
+const F64_MAX_INT: i64 = 9_007_199_254_740_992;
```
krzesi-mir: nit: Call it `F64_MAX_SAFE_INT`, write it as `2_i64.pow(std::f64::MANTISSA_DIGITS)`, and add a comment explaining what it actually is? (name taken from https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/MAX_SAFE_INTEGER)

krzesi-mir: nit: add F64_MIN_INT (or F64_MIN_SAFE_IT)?

sdl_viewer/src/terrain_drawer/layer.rs
```diff
@@ -0,0 +1,232 @@
...
+    texture_size: i64,
+}
+
+#[allow(dead_code)]
+impl TerrainLayer {
+    pub fn new<P: AsRef<std::path::Path>>(
+        program: &GlProgram,
+        path: P,
+        texture_size: u32,
+    ) -> Result<Self, std::io::Error> {
```
krzesi-mir: nit: use `io::Result`?

sdl_viewer/src/terrain_drawer/layer.rs
```diff
@@ -0,0 +1,232 @@
...
+        println!("Loading terrain.");
```
krzesi-mir: It's almost surely out of scope of this PR, but how about using some lightweight logging library instead of `println!`?

sdl_viewer/src/terrain_drawer/layer.rs
```diff
@@ -0,0 +1,232 @@
...
+            terrain_pos.y,
+            texture_size.try_into().unwrap(),
+            texture_size.try_into().unwrap(),
+        );
+        let heightmap = GlMovingWindowTexture::new(
+            &program,
+            Rc::clone(&program.gl),
+            "height",
+            texture_size,
+            0,
```
krzesi-mir: nit: add a comment `// texture_unit` here so we know what `0` means in this context?

sdl_viewer/src/terrain_drawer/layer.rs
```diff
@@ -0,0 +1,232 @@
...
+            terrain_pos.y,
+            texture_size.try_into().unwrap(),
+            texture_size.try_into().unwrap(),
+        );
+        let colormap = GlMovingWindowTexture::new(
+            &program,
+            Rc::clone(&program.gl),
+            "color",
+            texture_size,
+            1,
```
krzesi-mir: nit: add a comment `// texture_unit` here so we know what `1` means in this context?

sdl_viewer/src/terrain_drawer/layer.rs
```diff
@@ -0,0 +1,232 @@
...
+    // the vertical strip is determined by the movement in x direction.
+    pub fn update(&mut self, cur_world_pos: Vector3<f64>) {
+        let cur_pos = self
+            .grid_coordinates
+            .terrain_pos_for_camera_pos(cur_world_pos);
+        let moved = cur_pos - self.terrain_pos;
+        self.terrain_pos = cur_pos;
+        // Convert to f64, because GLSL doesn't understand i64
+        assert!(
+            cur_pos.x < F64_MAX_INT && cur_pos.x > -F64_MAX_INT,
```
krzesi-mir: nit: define and use `F64_MIN_INT` here and below?

sdl_viewer/src/terrain_drawer/layer.rs
```diff
@@ -0,0 +1,232 @@
...
+
+    // We already have the data between self.terrain_pos and self.terrain_pos + texture_size
+    // Only fetch the "L" shape that is needed, as separate horizontal and vertical strips.
+    // Don't get confused, the horizontal strip is determined by the movement in y direction and
+    // the vertical strip is determined by the movement in x direction.
+    pub fn update(&mut self, cur_world_pos: Vector3<f64>) {
+        let cur_pos = self
+            .grid_coordinates
+            .terrain_pos_for_camera_pos(cur_world_pos);
+        let moved = cur_pos - self.terrain_pos;
```
krzesi-mir: nit: I'd call it `shift`, but `moved` works well too.

sdl_viewer/src/terrain_drawer/layer.rs
```diff
@@ -0,0 +1,232 @@
...
+    terrain_from_world: Isometry3<f64>,
+}
+
+impl GridCoordinateFrame {
+    fn new(program: &GlProgram, metadata: Metadata, texture_size: u32) -> Self {
+        let u_origin = GlUniform::new(&program, "terrain_origin_m", metadata.origin);
+        let u_world_from_terrain = GlUniform::new(
+            &program,
+            "terrain_to_world",
+            Matrix4::from({
```
krzesi-mir: nit: would `Matrix4::from(Decomposed<_,_>::from(metadata.world_from_terrain.clone()))` work here?

krzesi-mir: nit: I think if you compute `terrain_from_world` befpre `u_world_from_terrain` then you wouldn't have to clone `metadata.world_from_terrain` (and save 8 characters :P).

nnmm: Unfortunately no, and `into::<Decomposed<_, _>>()` also doesn't work.

sdl_viewer/src/terrain_drawer/layer.rs
nnmm: That's not yet a constant function sadly.

sdl_viewer/src/terrain_drawer/layer.rs
nnmm: I just removed it for now. I'm also not super happy with random prints in the code.

sdl_viewer/src/terrain_drawer/layer.rs
nnmm: Nice!

krzesi-mir: How about this: `Matrix4::from(`**<**`Decomposed<_,_>`**>**`::from(metadata.world_from_terrain.into()))`?

sdl_viewer/src/terrain_drawer/layer.rs
krzesi-mir: ohh, too bad.

sdl_viewer/src/terrain_drawer/layer.rs
```diff
@@ -0,0 +1,233 @@
...
+            cur_pos.x < F64_MAX_SAFE_INT && cur_pos.x > F64_MIN_SAFE_INT,
+            "Terrain location not representable."
+        );
+        assert!(
+            cur_pos.y < F64_MAX_SAFE_INT && cur_pos.y > F64_MIN_SAFE_INT,
+            "Terrain location not representable."
+        );
+        self.u_terrain_pos.value = cur_pos.cast().unwrap();
+
+        let hori_strip = if moved.y > 0 {
```
krzesi-mir: How about something like:
```
let (pos_y, size_y) = if moved.y > 0 {
     (self.terrain_pos.y + self.texture_size, moved.y.try_into().unwrap())
} else {
    (cur_pos.y, moved.y.abs().try_into().unwrap())
};
let hori_strip = self.load(cur_pos.x, pos_y, self.texture_size.try_into().unwrap(), size_y);
```
? Similarly below.

sdl_viewer/src/terrain_drawer/layer.rs
nnmm: Sweet, that works! What's this syntax called?

sdl_viewer/src/terrain_drawer/layer.rs
nnmm: Tbh I find the current version a little easier on the brain, just because there's fewer local variables. Do you find it clearer or is it about not repeating code?

sdl_viewer/src/terrain_drawer/layer.rs
krzesi-mir: I don't know. Also, I think `Decomposed::<_,_>::from(...)` should work as well. I don't know which one is more idiomatic, though.

sdl_viewer/src/terrain_drawer/layer.rs
krzesi-mir: It's just a little more concise. If you find the current version easier to read, let's keep it.

PR #356, opened 2019-09-30
Introduce batches for all iterators.

feuerste:
This PR changes all iterators used by the point cloud client to return `PointsBatche`s instead of `Point`s. This enables reading of all attributes by the client instead of just position, color, and intensity.

Please review with `Hide whitespace changes` turned on.


nnmm:
Actually it seems like the pts format does contain the number of points, which would mean we might be able to make `NumberOfPoints` return just `usize`: https://github.com/patrikhuber/eos/blob/master/examples/data/image_0010.pts

src/read_write/pts.rs
```diff
@@ -49,21 +58,27 @@ impl Iterator for PtsIterator {
...
+                parts[6].parse::<u8>().unwrap(),
+            ));
+            point_count += 1;
+        }
+        if point_count == 0 {
+            None
+        } else {
+            Some(PointsBatch {
+                position,
+                attributes: vec![("color".to_string(), AttributeData::U8Vec3(color))]
```
nnmm: Nitpick: For just one value, it might be preferable to just insert once into a HashMap.

src/octree/generation.rs
```diff
@@ -66,35 +67,48 @@ fn split<P>(
...
+            .iter()
+            .map(|p| octree::ChildIndex::from_bounding_cube(&bounding_cube, &p))
+            .collect();
+        for (array_index, child_writer) in children.iter_mut().enumerate() {
+            let mut child_batch = batch.clone();
+            let keep: Vec<_> = child_indices
+                .iter()
+                .map(|i| i.as_u8() == array_index as u8)
+                .collect();
+            child_batch.retain(&keep);
```
nnmm: Could be made a little more efficient by making retain take an iterator, right?

nnmm: But not needed unless you think this function is particularly hot

nnmm: Or even better, do not clone and remove, but start from an empty point_batch and add points

src/octree/generation.rs
```diff
@@ -199,27 +214,33 @@ fn subsample_children_into(
...
-        point_iterator.for_each(|p| points.push(p));
+        let mut batch = node_iterator.next().unwrap();
+        node_iterator.for_each(|mut b| batch.append(&mut b).unwrap());
+        let (keep_parent, keep_child): (Vec<bool>, Vec<bool>) = (0..batch.position.len())
+            .map(|i| {
+                let in_parent = i % 8 == 0;
+                (in_parent, !in_parent)
+            })
+            .unzip();
+        let mut parent_batch = batch.clone();
```
nnmm: Same considerations for efficiency here, but optional unless you think this is a critical section

src/octree/generation.rs
feuerste: You are right, an iterator would be a little more efficient. However, I wanted to `assert` equal lengths of the keep vector and the internal vector, so for this we would need an `ExactSizeIterator`, which I didn't like as a prerequisite...
I actually also tested an implementation without `retain` just consisting of using `split_off` and `append`, which was much slower than cloning and removing. If you have a better idea, I'll be happy to test it, otherwise, I would like to keep it the way it is.

src/octree/generation.rs
feuerste: See comment above

src/read_write/pts.rs
feuerste: I personally like it to have as little mutables as possible. If fine with you, I would like to keep it as it is...

feuerste:
> Actually it seems like the pts format does contain the number of points, which would mean we might be able to make `NumberOfPoints` return just `usize`: https://github.com/patrikhuber/eos/blob/master/examples/data/image_0010.pts

Are you sure that this is the correct format? Did you find an example containing points divided into 7 parts(3 position, 1 unknown, 3 rgb)? How about just entirely removing pts reading?

feuerste:
@nnmm The pts iterator was entirely removed, ptal!

PR #357, opened 2019-09-30
Fix for clippy nightly.

feuerste:
See e.g. https://travis-ci.org/googlecartographer/point_cloud_viewer/jobs/591478379 why it failed in nightly.


feuerste:
Thanks for the review

PR #360, opened 2019-10-02
Remove unused pts reading.

feuerste:
As discussed offline with @SirVer , nobody is using the pts file reading code any more.


feuerste:
@nnmm FYI

src/read_write/input_file_iterator.rs
```diff
@@ -21,12 +21,10 @@ use std::path::PathBuf;
...
     Ply(PathBuf),
-    Pts(PathBuf),
 }
 
 pub enum InputFileIterator {
     Ply(PlyIterator),
-    Pts(PtsIterator),
 }
 
 impl Iterator for InputFileIterator {
```
SirVer: seems like InputFileIterator is now unnessecary.

feuerste: True, I removed it, too, PTAL!

PR #363, opened 2019-10-07
Verify frustum query equivalence between point cloud flavors

nnmm:
Plus subsuming `get_octree()`, `get_s2_cells()`, `Arguments::default()` and `RandomPointsOnEarth::new()` into a `setup()` function.


nnmm:
ptal @feuerste

PR #364, opened 2019-10-07
Point cloud client for s2 without bounding box.

feuerste:
The bounding box function needs to be implemented after s2 meta contains bounding box information.


feuerste:
Thanks for your review!

PR #365, opened 2019-10-07
Change back to include bounding box for s2 cells.

feuerste:
Since the meta changes are not in production yet, we can change it back to include the bounding box for both s2 cells and octrees.


src/octree/mod.rs
```diff
@@ -147,41 +124,20 @@ pub struct NodeData {
...
-    )
-}
-
 impl Octree {
     // TODO(sirver): This creates an object that is only partially usable.
     pub fn from_data_provider(data_provider: Box<dyn DataProvider>) -> Result<Self> {
         let meta_proto = data_provider.meta_proto()?;
+        if meta_proto.version < CURRENT_VERSION {
+            println!(
+                "Data is an older octree version: {}, current would be {}. \
```
catevita: maybe there should be point cloud instead of octree, because from version 12 also a s2 point cloud are possible

src/bin/upgrade_octree.rs
```diff
@@ -97,6 +95,7 @@ fn main() {
             9 => upgrade_version9(&args.directory, meta),
             10 => upgrade_version10(&args.directory, meta),
             11 => upgrade_version11(&args.directory, meta),
+            12 => upgrade_version12(&args.directory, meta),
             other if other == point_viewer::CURRENT_VERSION => {
```
catevita: maybe there should be point cloud instead of octree, because from version 12 also a s2 point cloud are possible

point_cloud_client/src/lib.rs
```diff
@@ -28,40 +28,43 @@ impl PointCloudClient {
...
-            PointClouds::S2Cells(_s2_cells) => {
-                // TODO(feuerste): Add bounding box to s2 meta.
-                unimplemented!();
-            }
-        }
 
         Ok(PointCloudClient {
             point_clouds,
-            aabb,
+            aabb: aabb.unwrap_or_else(Aabb3::zero),
```
catevita: in ECEF is the null bounding box in the center of the earth, could this bring some errors in the Point cloud client if it is defined like this? should be better this remain undefined?

feuerste: Previously it was the same for ECEF octrees. Actually this case should never happen now anyway, since I check in the beginning, whether `locations.is_empty()`.

src/bin/upgrade_octree.rs
feuerste: Good point, I changed it.

src/octree/mod.rs
```diff
@@ -60,37 +60,14 @@ impl OctreeMeta {
...
-        .mut_max()
-        .set_z(octree_meta.bounding_box.max().z);
     octree_proto.set_resolution(octree_meta.resolution);
 
     let octree_nodes = ::protobuf::RepeatedField::<proto::OctreeNode>::from_vec(nodes);
     octree_proto.set_nodes(octree_nodes);
 
     let mut meta = proto::Meta::new();
     meta.set_version(CURRENT_VERSION);
+    meta.set_bounding_box(proto::AxisAlignedCuboid::from(&octree_meta.bounding_box));
```
nnmm: Huh, that's way neater!

nnmm:
Should the bounding_box method be part of the PointCloud trait?

feuerste:
> Should the bounding_box method be part of the PointCloud trait?

Yes, makes sense, I moved it there.

feuerste:
Thanks for the review!

PR #366, opened 2019-10-08
Fix num points for IntoIter.

feuerste:


feuerste:
Thanks for the review!

PR #367, opened 2019-10-08
Only query specified attributes.

feuerste:


xray/src/generation.rs
```diff
@@ -364,6 +378,38 @@ impl HeightStddevColoringStrategy {
     }
 }
 
+impl ColoringStrategy for HeightStddevColoringStrategy {
```
feuerste: I just moved this one closer to its declaration and added the `attributes()` impl.

src/read_write/node_iterator.rs
```diff
@@ -48,6 +52,67 @@ impl NodeIterator {
...
+                .get_attributes()
+                .iter()
+                .map(|a| {
+                    AttributeDataType::from_proto(a.data_type)
+                        .map(|data_type| (a.name.clone(), data_type))
+                })
+                .collect::<Result<HashMap<String, AttributeDataType>>>()?,
+            _ => vec![
+                ("color".to_string(), AttributeDataType::U8Vec3),
+                ("intensity".to_string(), AttributeDataType::F32),
```
catevita: some octree could be just color only, will this maybe lead to problems in line 96?

feuerste: Here we just collect the possible mappings between attribute name and attribute data type, independent of the input `attributes`. In general it would be much cleaner to also store this mapping in the octree meta proto, where we could be sure about, which attributes also exist, but I think, this may be a little too much for this PR!?

src/octree/octree_test.rs
```diff
@@ -37,7 +37,14 @@ mod tests {
         let pool = scoped_pool::Pool::new(10);
         let tmp_dir = TempDir::new("octree").unwrap();
 
-        build_octree(&pool, &tmp_dir, 1.0, bounding_box, vec![batch].into_iter());
+        build_octree(
+            &pool,
+            &tmp_dir,
+            1.0,
+            bounding_box,
+            vec![batch].into_iter(),
```
nnmm: I like ArrayVec for such cases, but ymmv

xray/src/generation.rs
```diff
@@ -154,6 +154,8 @@ pub trait ColoringStrategy: Send {
     // After all points are processed, this is used to query the color that should be assigned to
     // the pixel (x, y) in the final tile image.
     fn get_pixel_color(&self, x: u32, y: u32, background_color: Color<u8>) -> Color<u8>;
+
+    fn attributes(&self) -> Vec<String>;
```
nnmm: `'static str` for avoiding `to_string()`?

src/read_write/node_iterator.rs
```diff
@@ -48,6 +52,67 @@ impl NodeIterator {
...
+        let mut all_reads =
+            data_provider.data(&id.to_string(), &[&["position"], attributes].concat())?;
+        let position_read = all_reads
+            .remove("position")
+            .ok_or_else(|| -> Error { "No position reader available.".into() })?;
+
+        let mut attribute_readers = HashMap::new();
+        for attribute in attributes {
+            match all_reads.remove(*attribute) {
+                Some(data) => {
```
nnmm: Wouldn't the `data_provider.data()` call have failed already if this wasn't Some? If so, how about `unwrap()` with a comment?

nnmm:
With this PR, XRays do not yet use the attribute() function, correct?

xray/src/generation.rs
feuerste: Good idea, changed it.

src/read_write/node_iterator.rs
feuerste: Done.

feuerste:
> With this PR, XRays do not yet use the attribute() function, correct?

They do, in https://github.com/googlecartographer/point_cloud_viewer/pull/367/files#diff-510d3c09c443afc25a26519f35942b99R483

src/read_write/node_iterator.rs
catevita: ok!

feuerste:
Thanks a lot for your reviews!

PR #369, opened 2019-10-10
Verify beam query equivalence between point cloud flavors 

nnmm:
Still fails, but should be passing after https://github.com/googlecartographer/point_cloud_viewer/pull/368

### DIFF
--- a/benches/generate_and_query/main.rs
+++ b/benches/generate_and_query/main.rs
@@ -2,15 +2,16 @@
 
 extern crate test;
 
-use cgmath::{InnerSpace, Matrix4, PerspectiveFov, Rad, Transform, Vector3};
+use cgmath::{Decomposed, InnerSpace, Matrix4, PerspectiveFov, Quaternion, Rad, Vector2, Vector3};
 use lazy_static::lazy_static;
 use point_viewer::data_provider::{DataProvider, OnDiskDataProvider};
 use point_viewer::iterator::{PointCloud, PointLocation, PointQuery};
+use point_viewer::math::OrientedBeam;
 use point_viewer::octree::{build_octree, Octree};
 use point_viewer::read_write::{Encoding, NodeWriter, OpenMode, RawNodeWriter, S2Splitter};
 use point_viewer::s2_cells::S2Cells;
 use protobuf::Message;
-use random_points::{Batched, RandomPointsOnEarth, SEED};
+use random_points::{Batched, RandomPointsOnEarth};
 use std::fs::File;
 use std::io::BufWriter;
 use std::path::Path;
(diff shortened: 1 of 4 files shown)


nnmm:
Hooray, it indeed passes!

benches/generate_and_query/random_points.rs
```diff
@@ -12,24 +12,26 @@ pub struct RandomPointsOnEarth {
...
 }
 
-pub const SEED: u64 = 80_293_751_232;
-
 impl RandomPointsOnEarth {
     pub fn new(width: f64, height: f64, size: usize, seed: u64) -> Self {
         let mut rng = StdRng::seed_from_u64(seed);
         let lat = rng.gen_range(-90.0, 90.0);
         let lon = rng.gen_range(-180.0, 180.0);
         let ecef_from_local = local_frame_from_lat_lng(lat, lon);
```
feuerste: I just looked at it again. Actually the fix is here, we need
```rust
let ecef_from_local = local_frame_from_lat_lng(lat, lon).inverse();
```
Then you can change back to `rect_bound()` below, as it's not a bug in s2.

feuerste:
@nnmm Please also rebase this one on master, once #370 is merged, so we have an improved index comparison, as previously discussed, which also helps for beam query comparison.

benches/generate_and_query/random_points.rs
nnmm: Hmm, why did changing it to cap_bound fix it then?

feuerste: I didn't investigate this further, but what I found ot is:
* Comparison of the behavior of the c++ and the rust version of `rect_bound`, which was exactly the same for the failing values.
* I looked at the `ecef_from_local` translations, which only made sense for the inverse.

benches/generate_and_query/main.rs
```diff
@@ -49,6 +52,7 @@ impl Default for Arguments {
             num_threads: 10,
             num_points: 1_000_000,
             batch_size: 5000,
+            seed: 80_293_751_232,
```
feuerste: I changed the seed to `80_293_751_234`, since it wasn't behaving as nicely as this one, but you can play around a bit, if you want...

benches/generate_and_query/random_points.rs
nnmm: So `local_frame_from_lat_lng()` actually doesn't return `ecef_from_local` but `local_from_ecef`?

feuerste: yes, see also https://github.com/lyft/avmapping-geometry/commit/395c119a92af7b3ad4a24d2760092e8124146e88#diff-fcd43e43678aa473c40a0a7be1f9f558R90

nnmm:
Thanks for the comments! Fixed.

PR #370, opened 2019-10-11
Improve index comparison.

feuerste:
This is a follow up to #368 and implements proper index comparison, including an assert for too many points to be indexed.


feuerste:
:dash: 
Thanks for the review!

PR #371, opened 2019-10-17
Add compile proto method.

feuerste:
This can potentially be reused elsewhere to compile protos.


feuerste:
Thanks for your review!

PR #373, opened 2019-10-18
Fix frustum query, clarify transformations.

feuerste:
This PR improves the following:
- The frustum query now consists of an isometry describing the transformation from frustum to global coordinates and a perspective transformation.
- The frustum query got a unit test.
- The isometry now transforms points instead of vectors, which is in accordance with cgmath and nalgebra.
- Clarifies the transformation direction of struct variables of OBB, oriented beam, and frustum.

### DIFF
--- a/src/iterator.rs
+++ b/src/iterator.rs
@@ -3,7 +3,7 @@ use crate::math::PointCulling;
 use crate::math::{AllPoints, Frustum, Isometry3, Obb, OrientedBeam};
 use crate::read_write::{Encoding, NodeIterator};
 use crate::PointsBatch;
-use cgmath::Point3;
+use cgmath::{EuclideanSpace, Point3};
 use collision::Aabb3;
 use crossbeam::deque::{Injector, Steal, Worker};
 use std::collections::BTreeMap;
@@ -116,7 +116,7 @@ where
                 batch.position = batch
                     .position
                     .iter()
-                    .map(|p| local_from_global * p)
+                    .map(|p| (local_from_global * &Point3::from_vec(*p)).to_vec())
                     .collect();
             }
             self.buf.append(&mut batch)?;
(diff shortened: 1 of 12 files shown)


src/iterator.rs
```diff
@@ -116,7 +116,7 @@ where
                 batch.position = batch
                     .position
                     .iter()
-                    .map(|p| local_from_global * p)
+                    .map(|p| (local_from_global * &Point3::from_vec(*p)).to_vec())
```
catevita: is the initialization directly into a vector not possible?

src/math.rs
```diff
@@ -252,8 +252,8 @@ impl<S: BaseFloat> Isometry3<S> {
 
 #[derive(Debug, Clone)]
 pub struct Obb<S> {
-    pub isometry: Isometry3<S>,
-    pub isometry_inv: Isometry3<S>,
+    pub from_obb: Isometry3<S>,
```
catevita: maybe here a comment like in the struct OrientedBeam would help understand the meaning of the isometry

src/math.rs
```diff
@@ -364,44 +367,46 @@ where
...
-            separating_axes: OrientedBeam::precompute_separating_axes(&isometry.rotation),
-            isometry,
+            corners: OrientedBeam::precompute_corners(&from_beam, &half_extent),
+            separating_axes: OrientedBeam::precompute_separating_axes(&from_beam.rotation),
+            from_beam,
         }
     }
 
-    fn precompute_corners(isometry: &Isometry3<S>, half_extent: &Vector2<S>) -> [Point3<S>; 4] {
+    fn precompute_corners(from_beam: &Isometry3<S>, half_extent: &Vector2<S>) -> [Point3<S>; 4] {
```
catevita: maybe here it is worth mentioning that the "corners"  are defined as the intersection of the beam with z = 0

src/math.rs
feuerste: I renamed all transformations, ptal!

src/math.rs
feuerste: Done.

feuerste:
Thanks for your review!

PR #375, opened 2019-10-22
sdl_viewer: Fix bug causing empty stripes in terrain

nnmm:
I noticed that when moving right, one column of values would get skipped, and similarly when moving up. The fix is that terrain_pos should still contain the previous position, not the one after moving, when making calls to `load()`.


nnmm:
meow @krzesi-mir

PR #376, opened 2019-10-22
Set S2 cell level to 16

nnmm:
Based on experiments with point cloud building performance, this level better balances not overfetching with avoiding overhead from too small cells.

### DIFF
--- a/src/read_write/s2.rs
+++ b/src/read_write/s2.rs
@@ -15,8 +15,8 @@ use std::path::PathBuf;
 
 /// The actual number of underlying writers is MAX_NUM_NODE_WRITERS * num_attributes.
 const MAX_NUM_NODE_WRITERS: usize = 25;
-/// Corresponds to cells of up to about 10m x 10m.
-const S2_SPLIT_LEVEL: u64 = 15;
+/// Corresponds to cells of up to about 160m x 160m.
+const S2_SPLIT_LEVEL: u64 = 16;
 
 pub struct S2Splitter<W> {
     writers: LruCache<CellID, W>,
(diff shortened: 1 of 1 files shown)


nnmm:
ptal @feuerste @catevita

catevita:
shall we put some of the result and timings somewhere into this pr or ticket? were it useful if the level was customizable -at least until the complete toolchain is active? I would also be interested on what the cell size means downwards: e.g. computing frustums in such big cells, or the need to sample from those.

feuerste:
Timings would definitely be helpful.. you have them somewhere anyways, right?

nnmm:
Making the level customizable is not super straightforward since the `NodeWriter` trait mandates the `new()` method, which doesn't have custom parameters like this one.

nnmm:
The timings are of course depending on bandwith, the tool used for uploading the files etc., but in my case they were:

73min for level 20
2 min for level 16

src/read_write/s2.rs
```diff
@@ -16,7 +16,7 @@ use std::path::PathBuf;
 /// The actual number of underlying writers is MAX_NUM_NODE_WRITERS * num_attributes.
 const MAX_NUM_NODE_WRITERS: usize = 25;
 /// Corresponds to cells of up to about 10m x 10m.
```
feuerste: Can you please update the comment?

nnmm: Done.

feuerste: Did you push?

nnmm: Well, no :D

PR #378, opened 2019-10-24
Clarify variable names, fix clippy nightly.

feuerste:


src/octree/generation.rs
```diff
@@ -159,7 +159,7 @@ fn split_node<'a, P>(
     scope: &Scope<'a>,
     octree_data_provider: &'a OnDiskDataProvider,
     octree_meta: &'a octree::OctreeMeta,
-    query_attributes: &'a HashMap<String, AttributeDataType>,
+    attribute_data_types: &'a HashMap<String, AttributeDataType>,
```
catevita: nit: in point_cloud/generation this very Hashmap is called just attributes or attributes_with_xyz

feuerste: Yes, then we should replace it there as well, but this is another repo...

feuerste: Actually here we really have attributes without xyz.

feuerste:
Thanks for your review!

PR #381, opened 2019-10-30
Fix oriented beam query.

feuerste:


nnmm:
I am a bit worried about Aabb being transformed into a larger Aabb. Would you also be fine with leaving src/iterator.rs and src/math.rs as they are for now? Or with panicking instead?

src/math.rs
```diff
@@ -42,14 +42,15 @@ where
...
-pub trait PointCulling<S>: objekt::Clone + Debug + Sync + Send
+pub trait PointCulling<S>: Debug + Sync + Send
 where
     S: BaseFloat + Sync + Send,
 {
     fn contains(&self, point: &Point3<S>) -> bool;
     // TODO(catevita): return Relation
     fn intersects_aabb3(&self, aabb: &Aabb3<S>) -> bool;
     fn transformed(&self, global_from_query: &Isometry3<S>) -> Box<dyn PointCulling<S>>;
+    fn corners(&self) -> [Point3<S>; 8];
```
nnmm: Wouldn't this be a little limiting in the future to have this in the trait?

feuerste: I currently think of culling as a bounding volume, which always has corners. I was even thinking about removing `AllPoints` from it, but it would have been too much effort in this PR.
If we need more than a bounding volume, e.g. sphere in the future, we can still take it out of the trait.

feuerste: It also nicely integrates with the changes in the s2 cells mod below...

feuerste:
> I am a bit worried about Aabb being transformed into a larger Aabb. Would you also be fine with leaving src/iterator.rs and src/math.rs as they are for now? Or with panicking instead/at least printing a warning?

As discussed offline, I reverted it to being transformed into an Obb. We also removed the oriented beam query altogether, to get rid of its hackish implementations.

feuerste:
Thanks for your review!

PR #382, opened 2019-11-07
sdl_viewer: Terrain shaders

nnmm:
This program is going to run always for the same mesh: a regular grid with spacing 1. What's varying is the textures and uniforms.

### DIFF
--- a/sdl_viewer/shaders/terrain.vs
+++ b/sdl_viewer/shaders/terrain.vs
@@ -3,7 +3,6 @@
 layout(location = 0) in ivec3 aPos;
 
 // These are constant throughout the program.
-uniform double grid_size;
 uniform dvec3 terrain_origin_m;
 uniform double terrain_res_m;
 
--- a/sdl_viewer/shaders/terrain.vs
+++ b/sdl_viewer/shaders/terrain.vs
@@ -1,0 +1,55 @@
+#version 410 core
+// The vertex positions are in grid coordinates, i.e. between 0 and grid_size (inclusive).
+layout(location = 0) in ivec3 aPos;
+
+// These are constant throughout the program.
+uniform double grid_size;
+uniform dvec3 terrain_origin_m;
+uniform double terrain_res_m;
+
+// The grid coordinates of the min corner of the currently visible terrain (the window).
+// Essentially, the terrain regular grid with resolution terrain_res_m and origin
+// terrain_origin_m, and the mesh is rendering a rectangle within that grid.
+uniform dvec2 terrain_pos;
+// These two are equal and set by the GlMovingWindowTexture.
+// They hold the coordinates (in pixels) of the min corner
+// of the texture within the sampler.
+uniform ivec2 height_texture_offset;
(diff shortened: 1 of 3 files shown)


nnmm:
ptal @krzesi-mir

sdl_viewer/shaders/terrain.vs
```diff
@@ -0,0 +1,55 @@
...
+// Essentially, the terrain regular grid with resolution terrain_res_m and origin
+// terrain_origin_m, and the mesh is rendering a rectangle within that grid.
+uniform dvec2 terrain_pos;
+// These two are equal and set by the GlMovingWindowTexture.
+// They hold the coordinates (in pixels) of the min corner
+// of the texture within the sampler.
+uniform ivec2 height_texture_offset;
+uniform ivec2 color_texture_offset;
+// The full projective transform.
+uniform dmat4 world_to_gl;
```
krzesi-mir: nit: I don't know about this repo, but in `avsoftware` we have the `x_from_y` convention for transforms: `gl_from_world`?

sdl_viewer/shaders/terrain.vs
```diff
@@ -0,0 +1,55 @@
+#version 410 core
+// The vertex positions are in grid coordinates, i.e. between 0 and grid_size (inclusive).
+layout(location = 0) in ivec3 aPos;
+
+// These are constant throughout the program.
+uniform double grid_size;
```
krzesi-mir: is `grid_size` used anywhere? Also, should it have `int` type?

sdl_viewer/shaders/terrain.vs
```diff
@@ -0,0 +1,55 @@
...
+  vec4 color;
+  uint quads;
+} vs_out;
+
+void main() {
+  // Each vertex position (aPos) maps to a pixel in the texture, but
+  // shifted (by height_texture_offset) with wraparound.
+  ivec2 texCoord = height_texture_offset + aPos.xy;
+  ivec2 texSize = textureSize(height, 0);
+  ivec2 texCoordModSize = texCoord - (texCoord / texSize) * texSize;
```
krzesi-mir: can we use https://www.khronos.org/registry/OpenGL-Refpages/gl4/html/mod.xhtml here?

nnmm: That's for floats, unfortunately

sdl_viewer/shaders/terrain.vs
nnmm: No, good point!

sdl_viewer/shaders/terrain.vs
nnmm: I think that's nicer too, but wanted to keep it consistent with the rest of point_cloud_viewer, where it uses `y_to_x`.

sdl_viewer/shaders/terrain.vs
krzesi-mir: i see

PR #384, opened 2019-11-11
Make split level configurable, set default to 20.

feuerste:


feuerste:
Thanks for your review!

PR #386, opened 2019-11-12
Move corners method outside culling, fix isometry identity constructor.

feuerste:
This will simplify the implementation of cullings without corners.

### DIFF
--- a/benches/generate_and_query/main.rs
+++ b/benches/generate_and_query/main.rs
@@ -291,7 +291,7 @@ fn check_frustum_query_equality() {
 #[test]
 fn check_obb_query_equality() {
     let (args, s2, oct, points) = setup();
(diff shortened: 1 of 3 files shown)


benches/generate_and_query/main.rs
```diff
@@ -291,7 +291,7 @@ fn check_frustum_query_equality() {
 #[test]
 fn check_obb_query_equality() {
     let (args, s2, oct, points) = setup();
-    let obb = Obb::new(Isometry3::zero(), Vector3::new(15.0, 15.0, 15.0));
+    let obb = Obb::new(Isometry3::one(), Vector3::new(15.0, 15.0, 15.0));
```
nnmm: Why did this work before?

feuerste: Honestly I don't know, it may have been fixed internally when multiplication was happening? I just know that definitely this was a bug...

feuerste:
Thanks for your review!

PR #387, opened 2019-11-12
Remove global_from_query transform from query.

feuerste:
This simplifies code around culling and will allow for other culling types, which can not be transformed.

### DIFF
--- a/xray/src/generation.rs
+++ b/xray/src/generation.rs
@@ -494,11 +494,9 @@ pub fn xray_from_points(
     let _ = point_cloud_client.for_each_point_data(&point_query, |mut points_batch| {
         seen_any_points = true;
         if let Some(query_from_global) = query_from_global {
-            points_batch.position = points_batch
-                .position
-                .iter()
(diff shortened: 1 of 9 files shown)


xray/src/generation.rs
```diff
@@ -471,21 +472,34 @@ pub struct Tile {
...
+        location,
     };
-    let _ = point_cloud_client.for_each_point_data(&point_location, |points_batch| {
+    let _ = point_cloud_client.for_each_point_data(&point_query, |mut points_batch| {
         seen_any_points = true;
+        if let Some(query_from_global) = query_from_global {
+            points_batch.position = points_batch
+                .position
+                .iter()
+                .map(|p| (query_from_global * &Point3::from_vec(*p)).to_vec())
```
nnmm: Is this performance-critical enough that we might want to use `iter_mut()` and `std::mem::replace()`?

feuerste: It's been like this before, but I changed it to a mutable loop...

feuerste:
Thanks for your review!

PR #388, opened 2019-11-12
Add s2 cell union query.

feuerste:


src/s2_cells/mod.rs
```diff
@@ -247,10 +250,14 @@ impl S2Cells {
             .collect();
         let mut cell_union = CellUnion(point_cells);
         cell_union.normalize();
-        let convex_hull = cell_union.rect_bound();
+        let rect = cell_union.rect_bound();
+        self.cells_intersecting(|cell: &Cell| rect.intersects_cell(cell))
+    }
+
+    fn cells_intersecting(&self, f: impl Fn(&Cell) -> bool) -> Vec<S2CellId> {
```
nnmm: Would it be a bit more elegant to just take a s2::Region?

feuerste: Good idea, I changed it.

src/s2_cells/mod.rs
```diff
@@ -247,10 +250,14 @@ impl S2Cells {
...
         self.cells
             .values()
-            .filter(|cell| convex_hull.intersects_cell(cell))
+            .filter(|cell| f(cell))
```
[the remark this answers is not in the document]
feuerste: Not relevant any more after above change...

feuerste:
Correct, there is no assumption about the cell union.

feuerste:
Thanks for your review!

PR #390, opened 2019-11-29
Write child first to not waste space.

feuerste:
During octree generation, we first rewrite the child, so no additional disk space is required.


feuerste:
This is done anyway, wrong thinking

PR #392, opened 2020-01-09
Support i16, i32, u16, u32 attribute data types

nnmm:
Also move out attribute code into its own module.


nnmm:
meow @feuerste

src/read_write/raw.rs
```diff
@@ -226,12 +226,19 @@ impl RawNodeReader {
                             .attributes
                             .insert(key.to_owned(), AttributeData::U8(attr));
                     }
-                    AttributeDataType::I64 => {
+                    AttributeDataType::U16 => {
```
feuerste: How about defining a macro for all the different types here, too?

nnmm: I thought about it, but I don't imagine it would help readability since there are many slightly different cases: U8 vs I8 vs (U16, U32, U64, I16, I32, I64) vs U8Vec3 vs F64Vec3.

src/read_write/node_writer.rs
```diff
@@ -114,13 +114,46 @@ macro_rules! derive_write_le_vec {
...
+
+impl WriteLE for u8 {
+    fn write_le(&self, writer: &mut DataWriter) -> Result<()> {
+        writer.write_u8(*self)
+    }
+}
+
+impl WriteLE for Vec<i8> {
+    fn write_le(&self, writer: &mut DataWriter) -> Result<()> {
+        let u8slice = unsafe { &*(self.as_slice() as *const _ as *const [u8]) };
```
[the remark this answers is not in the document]
nnmm: Done

src/read_write/raw.rs
feuerste: Hm, how about a `ReadLE` trait, which covers these cases (similar to what we do in `WriteLE`) and then the macro could just use `read_le` in combination with `batch.attributes.insert(...)`?

nnmm: Thinking about it, how about doing that in a follow-up PR that also (hopefully) improves the module structure? IMO it would make sense to split off a `little_endian` module that can read/write our data types and `Vec`s of them from/to arbitrary `Read` and `Write` instances.

feuerste: Sounds good, can you please add a TODO here and open a ticket for it?

nnmm:
Thanks for your review!

PR #393, opened 2020-01-13
Allow filtering of xrays by attribute intervals.

feuerste:
This adds a new filtering parameter to the point cloud query, so we can filter by attribute intervals while creating xrays.

### DIFF
--- a/src/iterator.rs
+++ b/src/iterator.rs
@@ -42,18 +42,18 @@ impl PointLocation {
 pub struct PointQuery<'a> {
     pub attributes: Vec<&'a str>,
     pub location: PointLocation,
-    pub filter: HashMap<&'a str, ClosedInterval>,
+    pub filter: HashMap<&'a str, ClosedInterval<f64>>,
 }
 
 /// Iterator over the points of a point cloud node within the specified PointCulling
 /// Essentially a specialized version of the Filter iterator adapter
 pub struct FilteredIterator<'a> {
     pub culling: Box<dyn PointCulling<f64>>,
-    pub filter: &'a HashMap<&'a str, ClosedInterval>,
+    pub filter: &'a HashMap<&'a str, ClosedInterval<f64>>,
     pub node_iterator: NodeIterator,
 }
 
-fn update_keep<T>(keep: &mut [bool], data: &[T], interval: &ClosedInterval)
+fn update_keep<T>(keep: &mut [bool], data: &[T], interval: &ClosedInterval<f64>)
 where
     T: ToPrimitive,
 {
--- a/src/iterator.rs
+++ b/src/iterator.rs
@@ -81,7 +81,12 @@ impl<'a> Iterator for FilteredIterator<'a> {
                 if let Some(data) = batch.attributes.get(*attrib) {
                     match data {
                         AttributeData::U8(d) => update_keep(&mut keep, d, interval),
+                        AttributeData::U16(d) => update_keep(&mut keep, d, interval),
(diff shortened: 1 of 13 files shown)


src/iterator.rs
```diff
@@ -37,26 +38,63 @@ impl<'a> PointQuery<'a> {
...
                 .position
                 .iter()
                 .map(|pos| {
                     culling.contains(&<Point3<f64> as cgmath::EuclideanSpace>::from_vec(*pos))
                 })
                 .collect();
+            for (attrib, interval) in self.filter {
+                if let Some(data) = batch.attributes.get(*attrib) {
+                    match data {
+                        AttributeData::U8(d) => update_keep(&mut keep, d, interval),
```
nnmm: Use the `match_attr_data` macro?

src/math.rs
```diff
@@ -46,6 +47,41 @@ where
...
+
+impl<T> FromStr for ClosedInterval<T>
+where
+    T: std::str::FromStr,
+{
+    type Err = T::Err;
+
+    fn from_str(s: &str) -> std::result::Result<Self, Self::Err> {
+        let bounds: Vec<&str> = s.split(',').collect();
+        assert_eq!(
```
nnmm: It would be nicer if a function with a Result couldn't panic – maybe this should have `Err = String` instead?

xray/src/generation.rs
```diff
@@ -471,37 +471,51 @@ pub struct Tile {
     pub resolution: f64,
 }
 
+pub struct XrayParameters {
+    pub point_cloud_client: PointCloudClient,
+    pub query_from_global: Option<Isometry3<f64>>,
+    pub filter: HashMap<String, ClosedInterval<f64>>,
```
nnmm: Nit: `filter` sounds like a verb, maybe something like `filter_intervals`?
Also, why are these fields specifically XrayParameters but not `image_size` or `coloring_strategy`?

src/iterator.rs
feuerste: Difficult because of the `unimplemented!` vector types.

src/math.rs
```diff
@@ -46,6 +47,41 @@ where
     clamped
 }
 
+#[derive(Debug, Clone, Copy)]
+pub struct ClosedInterval<T> {
+    pub lower_bound: T,
```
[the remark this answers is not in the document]
feuerste: Done.

src/math.rs
feuerste: It would be nicer indeed, but I didn't find an easy way to implement it...

xray/src/generation.rs
feuerste: I renamed it.
They are shared between `xray_from_points` and `build_xray_quadtree`, but `image_size`/`coloring_strategy` are not.

src/math.rs
feuerste: Can I merge in any case?

nnmm: Sure. But as for how to implement it with Err = String, you could of course do something like 
```rust
let mut parts = s.split(',');
let lower_bound = parts.next().ok_or("Empty string provided".into());
let upper_bound = parts.next().ok_or("Interval needs to have two comma-separated values".into());
if let Some(_) = parts.next() {
    return Err("Interval may not have more than two comma-separated values".into())
}
```

feuerste: I implemented an ParseClosedIntervalError, which is a little cleaner overall...

feuerste:
Thanks for your review!

PR #394, opened 2020-01-13
XRays: Support colormaps besides Jet

nnmm:
This allows using a different colormap that interpolates between a light blue and black for stddev xrays.

Should we also generalize regular XRays and Intensity-colored xrays with colormaps?


nnmm:
ptal @feuerste

xray/src/generation.rs
```diff
@@ -96,6 +49,15 @@ arg_enum! {
     }
 }
 
+arg_enum! {
+    #[derive(Debug)]
+    #[allow(non_camel_case_types)]
+    pub enum ColormapArgument {
+        jet,
+        purpley,
```
feuerste: maybe `purplish`?

tblaette: how about `purplig`? ;-)

feuerste: Haha, honestly, purplish exists, the others don't, so I would opt for the existing one...

xray/src/generation.rs
```diff
@@ -115,9 +78,21 @@ impl ColoringStrategyKind {
...
-                Box::new(HeightStddevColoringStrategy::new(max_stddev))
+            ColoringStrategyKind::ColoredWithHeightStddev(max_stddev, ColormapArgument::jet) => {
+                Box::new(HeightStddevColoringStrategy::<Jet>::new(max_stddev, Jet {}))
             }
+            ColoringStrategyKind::ColoredWithHeightStddev(
+                max_stddev,
+                ColormapArgument::purpley,
+            ) => Box::new(HeightStddevColoringStrategy::new(
+                max_stddev,
+                Monochrome(Color {
```
feuerste: How about defining a `const` for this color?

PR #397, opened 2020-01-27
Make point queries implement Serialize + Deserialize

nnmm:
The idea is to record real-life queries and build more accurate benchmarks from that.

This was previously blocked by s2 but I have added the trait instances upstream. Until that PR is merged, my fork is used.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -400,6 +400,17 @@ dependencies = [
 ]
 
 [[package]]
+name = "bigdecimal"
+version = "0.1.0"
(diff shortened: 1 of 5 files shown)


nnmm:
ptal @feuerste

feuerste:
How about also implementing https://github.com/googlecartographer/point_cloud_viewer/blob/master/src/s2_cells/mod.rs#L159 in this PR, as we change to your branch anyway?

nnmm:
Sure, done

PR #399, opened 2020-01-28
Refactor point cloud querying tests & add benchmarks

nnmm:
* Refactor to have code that can be shared between tests and benchmarks in a separate crate. This also makes it possible to depend on point_cloud_client, and makes the tests defined there now run by default by CI.
* Add benchmarks for queries. Current results:

```
test all_query_octree                    ... bench:  33,192,320 ns/iter (+/- 12,097,922)
test all_query_s2                        ... bench:  10,631,450 ns/iter (+/- 3,461,396)
test box_query_octree                    ... bench:   7,926,596 ns/iter (+/- 1,712,528)
test box_query_s2                        ... bench:   2,166,735 ns/iter (+/- 143,230)
test cell_union_query_octree             ... bench:  15,079,310 ns/iter (+/- 6,557,257)
test cell_union_query_s2                 ... bench:   1,086,969 ns/iter (+/- 85,365)
test frustum_query_octree                ... bench:  10,151,745 ns/iter (+/- 5,664,278)
test frustum_query_s2                    ... bench:   1,554,161 ns/iter (+/- 192,991)
test obb_query_octree                    ... bench:  10,024,159 ns/iter (+/- 3,689,549)
test obb_query_s2                        ... bench:   4,191,844 ns/iter (+/- 965,474)
```


nnmm:
meow @catevita

point_cloud_test/src/lib.rs
```diff
@@ -0,0 +1,130 @@
...
+impl Eq for Arguments {}
+
+impl Default for Arguments {
+    fn default() -> Arguments {
+        Arguments {
+            resolution: 0.001,
+            width: 200.0,
+            height: 20.0,
+            num_threads: 10,
+            num_points: 1_000_000,
```
catevita: as told offline, the point density might be also relevant for the query speed, maybe we can have some relevant use cases that can also be important to measure

point_cloud_test/src/lib.rs
```diff
@@ -0,0 +1,130 @@
...
+            S2_DIR = Some(s2_dir);
+            ARGUMENTS = Some(args.clone());
+        });
+        // If somebody called this twice with different arguments, caching doesn't make sense
+        assert_eq!(ARGUMENTS, Some(args.clone()));
+        let s2_path_buf = S2_DIR.as_ref().unwrap().path().to_owned();
+        let octree_path_buf = OCTREE_DIR.as_ref().unwrap().path().to_owned();
+        (s2_path_buf, octree_path_buf)
+    };
+    let data = SyntheticData::new(10.0, 10.0, args.num_points, args.seed);
```
catevita: nit: would it be helpful to move the numbers into costants?

point_cloud_test/src/queries.rs
```diff
@@ -0,0 +1,39 @@
...
+
+pub fn get_abb_query(data: SyntheticData) -> PointLocation {
+    let abb = data.bbox();
+    PointLocation::Aabb(abb)
+}
+
+pub fn get_obb_query(data: SyntheticData) -> PointLocation {
+    let obb = Obb::new(
+        data.ecef_from_local().clone(),
+        Vector3::new(15.0, 15.0, 15.0),
```
catevita: nit: also here, would it be useful to move these number into costants?

point_cloud_test/src/queries.rs
```diff
@@ -0,0 +1,39 @@
...
+use point_viewer::math::{Frustum, Obb, S2Point};
+
+pub fn get_abb_query(data: SyntheticData) -> PointLocation {
+    let abb = data.bbox();
+    PointLocation::Aabb(abb)
+}
+
+pub fn get_obb_query(data: SyntheticData) -> PointLocation {
+    let obb = Obb::new(
+        data.ecef_from_local().clone(),
```
catevita: nit: maybe add a comment to explain how the obb is constructed (i.e. for the test purpose it is currently taken in the middle of the generated point cloud

point_cloud_test/src/lib.rs
nnmm: Not sure what you're suggesting – increasing that number, or decreasing it? I plan on making a separate binary anyway that can read those parameters from the command line, which can be used to explore what factors make querying slow or fast.

point_cloud_test/src/lib.rs
nnmm: Actually, these should be args.width and args.height. Fixed.

point_cloud_test/src/queries.rs
nnmm: Made them data-dependent.

point_cloud_test/src/queries.rs
nnmm: Done.

PR #400, opened 2020-02-03
Update all versions to be future-proof.

feuerste:


feuerste:
Thanks for your review!

PR #402, opened 2020-02-04
Use stable benches.

feuerste:
So we don't have any more unstable dependencies.


feuerste:
Thanks for your review!

PR #403, opened 2020-02-04
Unify kebab case also in clap.

feuerste:
These were the last binaries still producing verbatim arguments.


feuerste:
Thanks for your review!

PR #404, opened 2020-02-13
Incorporate binning into Xray generation.

feuerste:


xray/src/generation.rs
```diff
@@ -127,7 +128,27 @@ pub trait ColoringStrategy: Send {
...
+trait BinnedColoringStrategy {
+    fn binning(&self) -> &Binning;
+    fn bins(&self, points_batch: &PointsBatch) -> Vec<i64> {
+        if let Some((attrib_name, size)) = self.binning() {
+            if let Some(attr_data) = points_batch.attributes.get(attrib_name) {
+                macro_rules! rhs {
+                    ($dtype:ident, $data:ident, $size:expr) => {
+                        $data
+                            .iter()
+                            .map(|e| *e as i64 / *$size as i64)
```
krzesi-mir: Wouldn't we want something like `(*e as f64 / *$size) as i64` here? Right now I think this will panic if `size < 1` (I don't know if you want to suport such a use case).

xray/src/generation.rs
```diff
@@ -127,7 +128,27 @@ pub trait ColoringStrategy: Send {
...
 
-    fn attributes(&self) -> Vec<&'static str>;
+    fn attributes(&self) -> HashSet<String>;
+}
+
+trait BinnedColoringStrategy {
+    fn binning(&self) -> &Binning;
+    fn bins(&self, points_batch: &PointsBatch) -> Vec<i64> {
+        if let Some((attrib_name, size)) = self.binning() {
+            if let Some(attr_data) = points_batch.attributes.get(attrib_name) {
```
krzesi-mir: q: don't we want to panic if the point cloud doesn't have `attrib_name` attribute?

xray/src/generation.rs
```diff
@@ -127,7 +128,27 @@ pub trait ColoringStrategy: Send {
     // the pixel (x, y) in the final tile image.
     fn get_pixel_color(&self, x: u32, y: u32, background_color: Color<u8>) -> Color<u8>;
 
-    fn attributes(&self) -> Vec<&'static str>;
+    fn attributes(&self) -> HashSet<String>;
```
krzesi-mir: Let's have a default implementation which returns an empty set so we don't have to do it twice downstairs?

xray/src/generation.rs
```diff
@@ -227,14 +257,22 @@ impl ColoringStrategy for IntensityColoringStrategy {
                 {
```
krzesi-mir: How about instead of `match self.per_column_data.entry(...)` we do `self.per_column_data.entry(...).or_default()`? We seem to be doing very similar stuff in both branches of the `match`.

xray/src/generation.rs
```diff
@@ -227,14 +257,22 @@ impl ColoringStrategy for IntensityColoringStrategy {
...
                     Entry::Occupied(mut e) => {
                         let per_column_data = e.get_mut();
-                        per_column_data.sum += intensity;
-                        per_column_data.count += 1;
+                        let data = per_column_data
+                            .entry(bins[i])
+                            .or_insert(IntensityPerColumnData { sum: 0.0, count: 0 });
```
krzesi-mir: nit: do you think it would make sense to derive the `Default` trait for `IntensityPerColumnData` and use `or_default` here?

xray/src/generation.rs
```diff
@@ -300,17 +365,23 @@ impl ColoringStrategy for PointColorColoringStrategy {
                 {
                     Entry::Occupied(mut e) => {
                         let per_column_data = e.get_mut();
-                        per_column_data.color_sum.red += clr.red;
-                        per_column_data.color_sum.green += clr.green;
-                        per_column_data.color_sum.blue += clr.blue;
-                        per_column_data.color_sum.alpha += clr.alpha;
-                        per_column_data.count += 1;
+                        let data = per_column_data.entry(bins[i]).or_insert(PerColumnData {
```
krzesi-mir: See my comments about `IntensityColorignStrategy`

xray/src/generation.rs
krzesi-mir: Or at least emit a warning

src/color.rs
```diff
@@ -43,6 +46,47 @@ impl Color<u8> {
...
+
+    fn div(self, rhs: f32) -> Self::Output {
+        Self {
+            red: self.red / rhs,
+            green: self.green / rhs,
+            blue: self.blue / rhs,
+            alpha: self.alpha / rhs,
+        }
+    }
+}
```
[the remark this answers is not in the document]
feuerste: I think in general using `derive_more` would be nice, but then we suddenly have it derived for non-float types, too, which I want to avoid.

xray/src/build_quadtree.rs
```diff
@@ -88,10 +88,16 @@ fn parse_arguments<T: Extension>() -> clap::ArgMatches<'static> {
                 .possible_values(&TileBackgroundColorArgument::variants())
                 .default_value("white"),
             clap::Arg::with_name("filter_interval")
-                .help("Filter intervals for attributes, e.g. --filter intensity=2.0,51.0")
+                .help("Filter intervals for attributes, e.g. --filter-interval intensity=2.0,51.0")
                 .long("filter-interval")
                 .takes_value(true)
                 .multiple(true),
+            clap::Arg::with_name("binning")
+                .help(
```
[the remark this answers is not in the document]
feuerste: Done.

xray/src/generation.rs
feuerste: Done.

xray/src/generation.rs
feuerste: Good point, I changed it.

xray/src/generation.rs
feuerste: I changed it to `expect` and a surrounding `match`.

xray/src/generation.rs
feuerste: Good idea, I changed it.

xray/src/generation.rs
feuerste: Done.

xray/src/generation.rs
feuerste: I changed it accordingly.

src/color.rs
```diff
@@ -43,6 +46,47 @@ impl Color<u8> {
     }
 }
 
+impl Add for Color<f32> {
```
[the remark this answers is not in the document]
krzesi-mir: :+1:

xray/src/generation.rs
```diff
@@ -127,7 +128,33 @@ pub trait ColoringStrategy: Send {
...
+                let attr_data = points_batch
+                    .attributes
+                    .get(attrib_name)
+                    .expect("Binning attribute needs to be available in points batch.");
+                macro_rules! rhs {
+                    ($dtype:ident, $data:ident, $size:expr) => {
+                        $data
+                            .iter()
+                            .map(|e| (*e as f64 / *$size) as i64)
+                            .collect::<Vec<i64>>()
```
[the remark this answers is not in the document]
feuerste: Done.

xray/src/generation.rs
```diff
@@ -127,7 +128,33 @@ pub trait ColoringStrategy: Send {
...
+                    };
+                }
+                match_1d_attr_data!(attr_data, rhs, size)
+            }
+            None => (vec![0; points_batch.position.len()]),
```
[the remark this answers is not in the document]
feuerste: Nope, I removed them.

xray/src/generation.rs
```diff
@@ -359,31 +346,13 @@ impl ColoringStrategy for PointColorColoringStrategy {
...
-                                color_sum: clr,
-                                count: 1,
-                            },
-                        );
-                        v.insert(data);
-                    }
-                }
+                    .or_default();
+                let bin_data = per_column_data.entry(bins[i]).or_default();
+                bin_data.color_sum += clr;
```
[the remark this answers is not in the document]
feuerste: Done

xray/src/generation.rs
```diff
@@ -257,11 +287,17 @@ impl ColoringStrategy for IntensityColoringStrategy {
...
+        attributes.insert("intensity".into());
+        if let Some((attr_name, _)) = &self.binning {
+            attributes.insert(attr_name.clone());
+        }
+        attributes
     }
 }
 
+#[derive(Default)]
 struct PerColumnData {
```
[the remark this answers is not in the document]
feuerste: I unified the two into one generic struct. However, a `mean()` method would be quite tricky because of the division in the case of a color.

feuerste:
Thanks for your review!

PR #405, opened 2020-02-14
Incorporate progress into xray building.

feuerste:
For massively big point clouds, we currently see almost no progress on the deepest level. This PR introduces progress reporting to get better feedback.
It also simplifies the leaf node generation logic.


feuerste:
Thanks for your review!

PR #406, opened 2020-02-14
Only create images if data is available.

feuerste:
Previously, for all levels above the leaf level images were generated in any case. This PR changes this behavior to only generate images, if at least one of its four children has some data in it.


feuerste:
Thanks for your review!

PR #408, opened 2020-02-17
Incorporate optional inpainting.

feuerste:
For very high resolutions we can easily create X-rays that are pretty sparse. In order to fill holes we can optionally activate inpainting with this PR.

### DIFF
--- a/xray/src/inpaint.rs
+++ b/xray/src/inpaint.rs
@@ -14,9 +14,9 @@ use scoped_pool::Pool;
 use std::fs;
 use std::path::{Path, PathBuf};
 use std::sync::Arc;
-use texture_synthesis::{Dims, Error, Example, Session};
+use texture_synthesis::{Dims, Example, Session};
 
-fn inpaint(image: RgbaImage, distance_px: u8) -> Result<RgbaImage, Error> {
+fn inpaint(image: RgbaImage, distance_px: u8) -> RgbaImage {
     let (width, height) = image.dimensions();
     let mask = map_colors(&image, |p| Luma([p[3]]));
     let closed_mask = close(&mask, Norm::LInf, distance_px);
@@ -30,9 +30,9 @@ fn inpaint(image: RgbaImage, distance_px: u8) -> Result<RgbaImage, Error> {
                 .set_sample_method(DynamicImage::ImageLuma8(mask)),
             Dims::new(width, height),
         )
-        .build()?;
-    let generated = texsynth.run(None);
-    Ok(generated.into_image().into_rgba())
+        .build()
+        .expect("Inpaint failed.");
+    texsynth.run(None).into_image().into_rgba()
 }
 
 fn get_inpaint_image_path(spatial_node_id: SpatialNodeId, output_directory: &Path) -> PathBuf {
@@ -128,7 +128,7 @@ fn create_inpaint_image(
     inpaint_distance_px: u8,
 ) -> ImageResult<()> {
     if let Some(mut inpaint_image) = stitched_image(spatial_node_id, output_directory)? {
-        inpaint_image = inpaint(inpaint_image, inpaint_distance_px).expect("Inpaint failed.");
+        inpaint_image = inpaint(inpaint_image, inpaint_distance_px);
         let inpaint_image_path = get_inpaint_image_path(spatial_node_id, output_directory);
         inpaint_image.save(inpaint_image_path)?;
     }
--- a/xray/src/inpaint.rs
+++ b/xray/src/inpaint.rs
@@ -89,6 +89,7 @@ impl<'a> SpatialNodeInpainter<'a> {
             let w = current.width() / 2;
             let h = current.height() / 2;
             let mut image = RgbaImage::from_pixel(4 * w, 4 * h, Rgba::from(TRANSPARENT.to_u8()));
+            // TODO(feuerste): Once we use image >= 0.23, add the ? operator
             image.copy_from(&current, w, h);
             let mut copy_subimage = |direction: Direction,
                                      from_x: u32,
@@ -99,6 +100,7 @@ impl<'a> SpatialNodeInpainter<'a> {
                                      to_y: u32|
              -> ImageResult<()> {
                 if let Some(neighbor) = self.image_from(direction)? {
+                    // TODO(feuerste): Once we use image >= 0.23, add the ? operator
                     image.copy_from(&neighbor.view(from_x, from_y, width, height), to_x, to_y);
                 }
                 Ok(())
--- a/xray/src/inpaint.rs
+++ b/xray/src/inpaint.rs
@@ -176,7 +176,7 @@ fn inpainting_step<P, F>(
     partitioning_function: P,
     spatial_node_function: F,
 ) where
-    P: FnMut(&&SpatialNodeId) -> bool + Send + Copy,
+    P: FnMut(&&SpatialNodeId) -> bool,
     F: Fn(SpatialNodeId) + Send + Copy,
 {
     let progress_bar = create_syncable_progress_bar(spatial_node_ids.len(), message);
--- a/xray/src/inpaint.rs
+++ b/xray/src/inpaint.rs
@@ -21,10 +21,14 @@ use std::path::{Path, PathBuf};
 use std::sync::Arc;
 use texture_synthesis::{Dims, Example, Session};
 
+/// Inpaints only holes with radius <= distance_px in the image, but leaves big borders untouched.
 fn inpaint(image: RgbaImage, distance_px: u8) -> RgbaImage {
     let (width, height) = image.dimensions();
+    // extract the alpha channel as sampling mask
     let mask = map_colors(&image, |p| Luma([p[3]]));
+    // fill holes in the alpha channel
     let closed_mask = close(&mask, Norm::LInf, distance_px);
+    // mark pixels to be inpainted
     let inpaint_mask = map_colors2(&closed_mask, &mask, |c, m| Luma([255 - (c[0] - m[0])]));
     let texsynth = Session::builder()
         // we perform multithreading outside, so one deterministic thread here is enough
@@ -198,7 +202,7 @@ impl<'a> Inpainting<'a> {
                     let progress_bar = Arc::clone(&progress_bar);
                     scope.execute(move || {
                         // TODO(feuerste): Move to rayon and try_for_each to handle errors properly!
-                        inpainter_function(&inpainter).unwrap();
+                        inpainter_function(&inpainter).expect("Inpainting failed.");
                         progress_bar.lock().unwrap().inc();
                     });
                 }
--- a/xray/src/inpaint.rs
+++ b/xray/src/inpaint.rs
@@ -128,7 +128,7 @@ fn interpolate_inpaint_image_horizontally(
             interpolate_sub_images(
                 current.sub_image(0, 0, width / 2, height),
                 left.view(width / 2, 0, width / 2, height),
-                |i, _, w, _| 0.5 * i as f32 / (w - 1) as f32,
+                |i, _, w, _| i as f32 / (w - 1) as f32,
             );
         }
         if let Some(right) = inpaint_image_from(spatial_node_id, output_directory, Direction::Right)
@@ -136,7 +136,7 @@ fn interpolate_inpaint_image_horizontally(
             interpolate_sub_images(
                 current.sub_image(width / 2, 0, width / 2, height),
                 right.view(0, 0, width / 2, height),
-                |i, _, w, _| 0.5 * (w - 1 - i) as f32 / (w - 1) as f32,
+                |i, _, w, _| (w - 1 - i) as f32 / (w - 1) as f32,
             );
         }
         current
@@ -153,7 +153,7 @@ fn interpolate_inpaint_image_vertically(
             interpolate_sub_images(
                 current.sub_image(0, 0, width, height / 2),
                 top.view(0, height / 2, width, height / 2),
-                |_, j, _, h| 0.5 * j as f32 / (h - 1) as f32,
+                |_, j, _, h| j as f32 / (h - 1) as f32,
             );
         }
         if let Some(bottom) =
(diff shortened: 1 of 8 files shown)


xray/src/inpaint.rs
```diff
@@ -0,0 +1,296 @@
...
+        inpaint_image_and_path_from(spatial_node_id, output_directory, None)?,
+        inpaint_image_and_path_from(spatial_node_id, output_directory, Direction::Right)?,
+    ) {
+        let (width, height) = (current.width() / 2, current.height());
+        let mut current_sub = current.sub_image(width, 0, width, height);
+        let mut right_sub = right.sub_image(0, 0, width, height);
+        for i in 0..width {
+            let right_weight = i as f32 / (width - 1) as f32;
+            for j in 0..height {
+                interpolate(i, j, right_weight, &mut right_sub, &mut current_sub);
```
krzesi-mir: Maybe I'm misunderstanding something: but shouldn't we pass `1-right_weight` instead of `right_weight` as the third argument here?

xray/src/inpaint.rs
```diff
@@ -0,0 +1,296 @@
...
+        inpaint_image_and_path_from(spatial_node_id, output_directory, None)?,
+        inpaint_image_and_path_from(spatial_node_id, output_directory, Direction::Bottom)?,
+    ) {
+        let (width, height) = (current.width(), current.height() / 2);
+        let mut current_sub = current.sub_image(0, height, width, height);
+        let mut bottom_sub = bottom.sub_image(0, 0, width, height);
+        for j in 0..height {
+            let bottom_weight = j as f32 / (height - 1) as f32;
+            for i in 0..width {
+                interpolate(i, j, bottom_weight, &mut bottom_sub, &mut current_sub);
```
krzesi-mir: same as above

xray/src/utils.rs
```diff
@@ -0,0 +1,8 @@
+use quadtree::NodeId;
+use std::path::{Path, PathBuf};
+
+pub fn get_image_path(directory: &Path, id: NodeId) -> PathBuf {
+    let mut rv = directory.join(&id.to_string());
```
krzesi-mir: nit: you probably don't need the ampersand.
nit2: I think we can simplify it to `directory.join(id.to_string()).with_extension("png")`

also, maybe it would make sense if `get_image_path` accepted a third argument: `extension`?

xray/src/inpaint.rs
```diff
@@ -0,0 +1,296 @@
...
+
+    let (first, second): (Vec<SpatialNodeId>, Vec<SpatialNodeId>) =
+        spatial_node_ids.iter().partition(partitioning_function);
+    run_partition(first);
+    run_partition(second);
+
+    progress_bar.lock().unwrap().finish_println("");
+}
+
+pub fn perform_inpainting(
```
krzesi-mir: IMHO, it might make sense to have a structure like
```
struct Inpainter {
    pool: &Pool,
    output_directory: &Path,
    inpaint_distance_px: u8,
    leaf_node_ids: &FnvHashSet<NodeId>,
}
```
with the following methods:
`inpainting_step, perform_inpainting, create_inpaint_image,  interpolate_inpaint_image_with_right, interpolate_inpaint_image_with_bottom, apply_inpainting`, and perhaps other.
I think it would make it a little easier to use/read as you wouldn't have to pass pool, directory, and leaves to every function.

xray/src/inpaint.rs
```diff
@@ -0,0 +1,296 @@
...
+    ) {
+        let (width, height) = (current.width() / 2, current.height());
+        let mut current_sub = current.sub_image(width, 0, width, height);
+        let mut right_sub = right.sub_image(0, 0, width, height);
+        for i in 0..width {
+            let right_weight = i as f32 / (width - 1) as f32;
+            for j in 0..height {
+                interpolate(i, j, right_weight, &mut right_sub, &mut current_sub);
+            }
+        }
```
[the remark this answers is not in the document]
feuerste: Done.

xray/src/inpaint.rs
feuerste: I restructured the code to use two helper structs: One for the methods applying to a single node, one for the step methods. Please take another look!

xray/src/utils.rs
feuerste: I simplified the method, but kept the hard coded extension.

xray/src/utils.rs
```diff
@@ -0,0 +1,15 @@
+use image::{ImageResult, RgbaImage};
+use quadtree::NodeId;
+use std::path::{Path, PathBuf};
+
+pub fn get_image_path(directory: &Path, id: NodeId) -> PathBuf {
+    directory.join(id.to_string()).with_extension("png")
+}
+
+pub fn image_from_path(image_path: &Path) -> Option<ImageResult<RgbaImage>> {
+    if image_path.exists() {
```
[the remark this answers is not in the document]
feuerste: Nope, no more nightly.

xray/src/inpaint.rs
```diff
@@ -0,0 +1,299 @@
...
+            image.copy_from(&current, w, h);
+            let mut copy_subimage = |dir: Direction,
+                                     from_x: u32,
+                                     from_y: u32,
+                                     width: u32,
+                                     height: u32,
+                                     to_x: u32,
+                                     to_y: u32|
+             -> ImageResult<()> {
+                if let Some(neighbor) = self.image_from(dir)? {
```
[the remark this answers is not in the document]
feuerste: Done.

xray/src/inpaint.rs
```diff
@@ -0,0 +1,299 @@
...
+                    image.copy_from(&neighbor.view(from_x, from_y, width, height), to_x, to_y);
```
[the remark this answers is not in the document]
feuerste: Which errors do you mean?

xray/src/inpaint.rs
```diff
@@ -0,0 +1,299 @@
...
+        "Creating inpaint images",
+        |_| false,
+        |inpainter| inpainter.create_inpaint_image(inpaint_distance_px),
+    );
+
+    inpainting.step(
+        "Horizontally interpolating inpaint images",
+        // Interleave interpolation to avoid race conditions when writing images
+        |spatial_node_id| spatial_node_id.x() % 2 == 0,
+        |inpainter| inpainter.interpolate_inpaint_image_with_right(),
```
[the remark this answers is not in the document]
feuerste: Done.

xray/src/inpaint.rs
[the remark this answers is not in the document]
krzesi-mir: [Image::GenericImage::copy_from](https://docs.rs/image/0.23.0/image/trait.GenericImage.html#method.copy_from) returns an error if the image is too big to be copied at the given position. It won't happen with the current implementation, but it doesn't hurt to use `?` here.

xray/src/inpaint.rs
krzesi-mir: Looks good!

xray/src/inpaint.rs
```diff
@@ -0,0 +1,262 @@
...
+                Ok(())
+            };
+            copy_subimage(Direction::TopLeft, w, h, w, h, 0, 0)?;
+            copy_subimage(Direction::Top, 0, h, 2 * w, h, w, 0)?;
+            copy_subimage(Direction::TopRight, 0, h, w, h, 3 * w, 0)?;
+            copy_subimage(Direction::Right, 0, 0, w, 2 * h, 3 * w, h)?;
+            copy_subimage(Direction::BottomRight, 0, 0, w, h, 3 * w, 3 * h)?;
+            copy_subimage(Direction::Bottom, 0, 0, 2 * w, h, w, 3 * h)?;
+            copy_subimage(Direction::BottomLeft, w, 0, w, h, 0, 3 * h)?;
+            copy_subimage(Direction::Left, w, 0, w, 2 * h, 0, h)?;
```
krzesi-mir: I checked and it's correct but I wish there was a nicer way :)

xray/src/inpaint.rs
```diff
@@ -0,0 +1,262 @@
...
+use point_viewer::color::TRANSPARENT;
+use point_viewer::utils::create_syncable_progress_bar;
+use quadtree::{Direction, NodeId, SpatialNodeId};
+use scoped_pool::Pool;
+use std::fs;
+use std::path::{Path, PathBuf};
+use std::sync::Arc;
+use texture_synthesis::{Dims, Example, Session};
+
+fn inpaint(image: RgbaImage, distance_px: u8) -> RgbaImage {
```
krzesi-mir: I think it would help if we added few comments with short explanations what some of these lines do.

xray/src/inpaint.rs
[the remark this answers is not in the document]
feuerste: Ah, now I see what you mean.
Unfortunately the version we are using returns a `bool` (see https://docs.rs/image/0.22.5/image/trait.GenericImage.html#method.copy_from). That's why clippy didn't catch it. 
How about adding a `TODO` instead and once we migrate to 0.23 I replace it with `?`?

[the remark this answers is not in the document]
krzesi-mir: ahhhh, didn't realize we are using an old version of `Image`. Yeah, let's add a TODO.

xray/src/inpaint.rs
```diff
@@ -0,0 +1,264 @@
...
+            interpolate_subimages(
+                &mut right.sub_image(0, 0, width, height),
+                &mut current.sub_image(width, 0, width, height),
+                |i, _| i as f32 / (width - 1) as f32,
+            );
+            current.save(current_path)?;
+            right.save(right_path)?;
+        }
+        Ok(())
+    }
```
krzesi-mir: I think we could replace this and the next function by one, more generic, function:
```
    fn interpolate_inpaint_image_with_neighbor(&self, direction::Direction) -> ImageResult<()> {
        if let (Some((mut current, current_path)), Some((mut neighbor, neighbor_path))) = (
            self.inpaint_image_and_path_from(None)?,
            self.inpaint_image_and_path_from(direction)?,
        ) {
            let (width, height) = (current.width() / 2, current.height());
            let current_subimage = match direction {
                  RIGHT => current.sub_image(width, 0, width, height),
                  BOTTOM => current.sub_image(0, height, width, height),
                  _ => unimplemented!()
           }
           let weighting_function = ... // the same for weighting_function
            interpolate_subimages(
                &mut right.sub_image(0, 0, width, height),
                &mut current_sub_image,
                weighting_function,
            );
            current.save(current_path)?;
            neighbor.save(neighbor_path)?;
        }
        Ok(())
    }
```

xray/src/inpaint.rs
```diff
@@ -0,0 +1,264 @@
...
+            self.pool.scoped(|scope| {
+                for spatial_node_id in spatial_node_ids {
+                    let inpainter = SpatialNodeInpainter {
+                        spatial_node_id,
+                        output_directory: self.output_directory,
+                    };
+                    let progress_bar = Arc::clone(&progress_bar);
+                    scope.execute(move || {
+                        // TODO(feuerste): Move to rayon and try_for_each to handle errors properly!
+                        inpainter_function(&inpainter).unwrap();
```
krzesi-mir: nit: `unwrap()` -> `expect("Inpainting failed.")`?

xray/src/inpaint.rs
feuerste: Done.

xray/src/inpaint.rs
feuerste: I would prefer keeping two functions. The generic one would be longer and look like this:
```rust
    fn interpolate_inpaint_image_with_neighbor(&self, direction: Direction) -> ImageResult<()> {
        if let (Some((mut current, current_path)), Some((mut neighbor, neighbor_path))) = (
            self.inpaint_image_and_path_from(None)?,
            self.inpaint_image_and_path_from(direction)?,
        ) {
            let (width, height) = match direction {
                Direction::Right => (current.width() / 2, current.height()),
                Direction::Bottom => (current.width(), current.height() / 2),
                _ => unimplemented!(),
            };
            let (current_subimage, neighbor_subimage) = match direction {
                Direction::Right => (
                    current.sub_image(width, 0, width, height),
                    neighbor.sub_image(0, 0, width, height),
                ),
                Direction::Bottom => (
                    current.sub_image(0, height, width, height),
                    neighbor.sub_image(0, 0, width, height),
                ),
                _ => unimplemented!(),
            };
            let weighting_function = match direction {
                Direction::Right => |i: u32, _: u32| i as f32 / (width - 1) as f32,
                Direction::Bottom => |_: u32, j: u32| j as f32 / (height - 1) as f32,
                _ => unimplemented!(),
            };
            interpolate_subimages(
                &mut neighbor_subimage,
                &mut current_subimage,
                weighting_function,
            );
            current.save(current_path)?;
            neighbor.save(neighbor_path)?;
        }
        Ok(())
    }
```

xray/src/inpaint.rs
feuerste: Done.

xray/src/inpaint.rs
feuerste: Thought about it a little more this morning. I think I found a quite nice generalization based on yours, so I change it.

feuerste:
> LGTM,
> I left one more nit and one suggestion how to combine two functions into one.
> Did you check if it still works after all the changes?

I checked it one more time, the outputs are still identical after all the changes.
Thanks for your review!

xray/src/inpaint.rs
```diff
@@ -132,36 +132,33 @@ impl<'a> SpatialNodeInpainter<'a> {
...
+            let neighbor_weighting_function: Box<dyn Fn(u32, u32) -> f32> = match direction {
+                Direction::Right => Box::new(|i, _| i as f32 / (width - 1) as f32),
+                Direction::Bottom => Box::new(|_, j| j as f32 / (height - 1) as f32),
+                _ => unimplemented!(),
+            };
             interpolate_subimages(
-                &mut bottom.sub_image(0, 0, width, height),
-                &mut current.sub_image(0, height, width, height),
-                |_, j| j as f32 / (height - 1) as f32,
+                &mut neighbor.sub_image(neighbor_x, neighbor_y, width, height),
```
krzesi-mir: nit: neighbor_x and neighbor_y are crrently always set to 0. Why don't we pass 0's directly here?

feuerste: In case we implemented top/left as well they wouldn't be zero, so I preferred this generic way...

krzesi-mir: makes sense :+1:

minamina11:
remove my name asap

PR #409, opened 2020-02-19
Code clean up.

feuerste:
This includes:

- Unifying progress reporting
- Progress reporting now has one report per step
- Moving background assignment into separate step

This will help integrating inpainting.


src/utils.rs
```diff
@@ -20,9 +19,9 @@ where
     Ok((s[..pos].parse()?, s[pos + 1..].parse()?))
 }
 
-pub fn create_progress_bar(total: usize, message: &str) -> Arc<Mutex<ProgressBar<Stdout>>> {
```
krzesi-mir: We could have another function, say, `create_progress_bar_arc_mutex` (i couldn't come up with a better name :) and do `Arc::new(Mutex::new(create_progress_bar(total, message))))` inside so you don't have to do it three times below.

src/utils.rs
```diff
@@ -2,7 +2,6 @@ use pbr::ProgressBar;
 use std::error::Error;
 use std::io::Stdout;
 use std::str::FromStr;
-use std::sync::{Arc, Mutex};
 use std::time::Duration;
 
 pub const PROGRESS_REFRESH_RATE_SECS: u64 = 2;
```
krzesi-mir: tiny nit: (I think) As `Duration::from_sec` is a `const fn`, we can do 
```
pub const PROGRESS_BAR_REFRESH_RATE: Duration = Duration::from_secs(2);
``` 
here instead of below.
It looks cleaner to me than how it's currently done, but it's a matter of taste, I guess.

src/utils.rs
feuerste: Done, I added a `create_syncable_progress_bar(...)` function

src/utils.rs
feuerste: Done

feuerste:
Thanks for your review!

PR #410, opened 2020-02-20
Incorporate spatial quadtree indexing.

feuerste:
This is required to query neighbors on the same quadtree level.

### DIFF
--- a/quadtree/src/lib.rs
+++ b/quadtree/src/lib.rs
@@ -214,7 +214,7 @@ impl fmt::Display for NodeId {
     }
 }
 
-pub enum SpatialNeighbor {
+pub enum Direction {
     Left,
     TopLeft,
     Top,
@@ -225,7 +225,7 @@ pub enum SpatialNeighbor {
     BottomLeft,
 }
 
-#[derive(Clone, Copy)]
+#[derive(Debug, Hash, Clone, Copy, PartialEq, Eq)]
 pub struct SpatialNodeId {
     level: u8,
     x: u64,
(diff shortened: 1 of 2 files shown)


quadtree/src/lib.rs
```diff
@@ -246,4 +332,13 @@ mod tests {
             assert_eq!(&NodeId::from_str(id).unwrap().to_string(), id);
         }
     }
+
```
krzesi-mir: Let's add one more test where we explicitly test where we explicitly test the output of `SpatialNodeId::from`? For example (I might have coordinates wrong :P)
```
assert_eq!(SpatialNodeId{level: 3, x:5, y:4}, SpatialNodeId::from(NodeId::from_str("r301").unwrap());
```

quadtree/src/lib.rs
```diff
@@ -214,9 +214,95 @@ impl fmt::Display for NodeId {
     }
 }
 
+pub enum SpatialNeighbor {
```
krzesi-mir: nit: how about renaming this `enum` to `Direction` or `SpatialNeighborKind`?

quadtree/src/lib.rs
```diff
@@ -214,9 +214,95 @@ impl fmt::Display for NodeId {
...
+/// on how to convert between coordinates and the quadkey.
+impl From<NodeId> for SpatialNodeId {
+    fn from(node_id: NodeId) -> Self {
+        let level = node_id.level;
+        let mut x = 0;
+        let mut y = 0;
+        for i in 1..=level {
+            let mask = 1 << (level - i);
+            match 3 & (node_id.index >> ((level - i) * 2)) {
+                1 => y |= mask,
```
krzesi-mir: This is different than what's in the link and your coordinates are transposed w.r.t. the coordinates in the doc, in the sense that whenever they have `y` you have `x`, and whenever they have `x` you have `y` (I hope this makes sense).
Is it intentional?

quadtree/src/lib.rs
```diff
@@ -246,4 +332,13 @@ mod tests {
...
     }
+
+    #[test]
+    fn test_conversion() {
+        for id in &["r", "r0", "r123323"] {
+            let node_id = NodeId::from_str(id).unwrap();
+            let spatial_node_id = SpatialNodeId::from(node_id);
+            assert_eq!(NodeId::from(spatial_node_id), node_id);
+        }
+    }
```
krzesi-mir: And maybe let's add another unit test for your `SpatialNodeId::neighbor` function?

quadtree/src/lib.rs
feuerste: Done, I renamed it to `Direction`.

quadtree/src/lib.rs
feuerste: Done, I had to switch x and y.

quadtree/src/lib.rs
```diff
@@ -214,9 +214,95 @@ impl fmt::Display for NodeId {
...
+        let level = spatial_node_id.level;
+        let mut index = 0;
+        for i in 1..=level {
+            index <<= 2;
+            let mask = 1 << (level - i);
+            if (spatial_node_id.y & mask) != 0 {
+                index += 1;
+            }
+            if (spatial_node_id.x & mask) != 0 {
+                index += 2;
```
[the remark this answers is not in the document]
feuerste: Done.

quadtree/src/lib.rs
```diff
@@ -214,9 +214,95 @@ impl fmt::Display for NodeId {
...
+            let mask = 1 << (level - i);
+            match 3 & (node_id.index >> ((level - i) * 2)) {
+                1 => y |= mask,
+                2 => x |= mask,
+                3 => {
+                    x |= mask;
+                    y |= mask;
+                }
+                _ => (),
+            }
```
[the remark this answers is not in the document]
feuerste: I changed it as proposed above.

quadtree/src/lib.rs
feuerste: Actually we here have a right hand coordinate frame (in contrast to the left hand frame for the Bing link), so I had to switch x and y.

quadtree/src/lib.rs
```diff
@@ -214,9 +214,95 @@ impl fmt::Display for NodeId {
...
+/// See e.g. https://docs.microsoft.com/en-us/bingmaps/articles/bing-maps-tile-system
+/// on how to convert between coordinates and the quadkey.
+impl From<NodeId> for SpatialNodeId {
+    fn from(node_id: NodeId) -> Self {
+        let level = node_id.level;
+        let mut x = 0;
+        let mut y = 0;
+        for i in 1..=level {
+            let mask = 1 << (level - i);
+            match 3 & (node_id.index >> ((level - i) * 2)) {
```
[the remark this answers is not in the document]
feuerste: Done.

quadtree/src/lib.rs
feuerste: Puh, were should I start and end here? I think the function is trivial enough not to test it...

quadtree/src/lib.rs
krzesi-mir: :+1:

quadtree/src/lib.rs
krzesi-mir: yeah you are probably right :P

feuerste:
Thanks for your review!

PR #411, opened 2020-02-26
Replace scoped-pool with rayon.

feuerste:
This will hopefully also fix https://github.com/reem/rust-scoped-pool/issues/4 which we run into when using lots of threads.


feuerste:
Thanks for your review!

PR #412, opened 2020-02-28
Disable threading for point cloud client inside X-ray creation.

feuerste:
This just led to an unnecessary high amount of threads created. We now just do threading outside, so outer (X-ray) and inner (point cloud client) threads don't fight for computation time...

### DIFF
--- a/xray/src/build_quadtree.rs
+++ b/xray/src/build_quadtree.rs
@@ -174,8 +174,11 @@ pub fn run<T: Extension>(data_provider_factory: DataProviderFactory) {
         .unwrap()
         .map(String::from)
         .collect::<Vec<_>>();
-    let point_cloud_client = PointCloudClient::new(&point_cloud_locations, data_provider_factory)
-        .expect("Could not create point cloud client.");
+    let mut point_cloud_client =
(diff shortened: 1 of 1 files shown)


xray/src/build_quadtree.rs
```diff
@@ -174,8 +174,11 @@ pub fn run<T: Extension>(data_provider_factory: DataProviderFactory) {
         .unwrap()
         .map(String::from)
         .collect::<Vec<_>>();
-    let point_cloud_client = PointCloudClient::new(&point_cloud_locations, data_provider_factory)
-        .expect("Could not create point cloud client.");
+    let mut point_cloud_client =
```
krzesi-mir: nit: how about passing `number of threads` as a new, maybe optional, parameter to `PointCloudClient::new`?

feuerste: I thought about this, but I would prefer changing the interface in a separate PR, since there are already 3 users.

krzesi-mir: ah, okay

feuerste:
Thanks for your review!

PR #413, opened 2020-02-28
Introduce PointCloudClientBuilder.

feuerste:
This is mainly to clean up interfaces and get rid of public members in `PointCloudClient`.

### DIFF
--- a/point_cloud_client/src/lib.rs
+++ b/point_cloud_client/src/lib.rs
@@ -14,19 +14,88 @@ enum PointClouds {
 pub struct PointCloudClient {
     point_clouds: PointClouds,
     aabb: Aabb3<f64>,
-    pub num_points_per_batch: usize,
-    pub num_threads: usize,
-    pub buffer_size: usize,
+    num_points_per_batch: usize,
+    num_threads: usize,
+    buffer_size: usize,
 }
 
 impl PointCloudClient {
-    pub fn new(locations: &[String], data_provider_factory: DataProviderFactory) -> Result<Self> {
-        if locations.is_empty() {
+    pub fn bounding_box(&self) -> &Aabb3<f64> {
+        &self.aabb
+    }
+
+    pub fn for_each_point_data<F>(&self, point_query: &PointQuery, mut func: F) -> Result<()>
+    where
+        F: FnMut(PointsBatch) -> Result<()>,
+    {
+        match &self.point_clouds {
+            PointClouds::Octrees(octrees) => {
+                let mut parallel_iterator = ParallelIterator::new(
+                    octrees,
+                    point_query,
+                    self.num_points_per_batch,
+                    self.num_threads,
+                    self.buffer_size,
+                );
+                parallel_iterator.try_for_each_batch(&mut func)?;
+            }
+            PointClouds::S2Cells(s2_cells) => {
+                let mut parallel_iterator = ParallelIterator::new(
+                    s2_cells,
+                    point_query,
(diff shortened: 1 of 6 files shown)


point_cloud_client/src/lib.rs
```diff
@@ -14,19 +14,88 @@ enum PointClouds {
...
+                    point_query,
+                    self.num_points_per_batch,
+                    self.num_threads,
+                    self.buffer_size,
+                );
+                parallel_iterator.try_for_each_batch(&mut func)?;
+            }
+        }
+        Ok(())
+    }
```
krzesi-mir: idea for another PR: since we are doing the same, up to the type of point cloud, thing in both branches of the match, we could probably simplify it slightly by using Boxes/Trait Objects or defining a new generic function `for_each<P: PointCloud>(&self,  point_clouds: &[PointCloud], point_query: &PointQuery, mut func: F) -> result<()>` which would do exactly what we currently do in each of the branches, and then simply call that function inside each branch of the match.

point_cloud_client/src/lib.rs
```diff
@@ -14,19 +14,88 @@ enum PointClouds {
...
+                    s2_cells,
+                    point_query,
+                    self.num_points_per_batch,
+                    self.num_threads,
+                    self.buffer_size,
+                );
+                parallel_iterator.try_for_each_batch(&mut func)?;
+            }
+        }
+        Ok(())
```
krzesi-mir: nit: I think you can remove `?;` from each branch and this `Ok())`?

point_cloud_client/src/lib.rs
```diff
@@ -14,19 +14,88 @@ enum PointClouds {
...
+    buffer_size: usize,
+}
+
+impl<'a> PointCloudClientBuilder<'a> {
+    pub fn new(locations: &'a [String], data_provider_factory: DataProviderFactory) -> Self {
+        Self {
+            locations,
+            data_provider_factory,
+            num_points_per_batch: NUM_POINTS_PER_BATCH,
+            num_threads: num_cpus::get() - 1,
```
krzesi-mir: silly question: what would happen if `num_cpus::get()` returns `1` (I don't know if it's even possible in XXI century :P) and, consequently, `num_thread = 0`? Wouldn't something stop working down the road?

feuerste: Good catch, I changed it everywhere to `std::cmp::max(1, num_cpus::get() - 1)`

point_cloud_client/src/lib.rs
```diff
@@ -14,19 +14,88 @@ enum PointClouds {
...
+pub struct PointCloudClientBuilder<'a> {
+    locations: &'a [String],
+    data_provider_factory: DataProviderFactory,
+    num_points_per_batch: usize,
+    num_threads: usize,
+    buffer_size: usize,
+}
+
+impl<'a> PointCloudClientBuilder<'a> {
+    pub fn new(locations: &'a [String], data_provider_factory: DataProviderFactory) -> Self {
```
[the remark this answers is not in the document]
feuerste: I actually integrated it into the builder, so `new` now only takes the locations.

point_cloud_client/src/lib.rs
feuerste: Done

point_cloud_client/src/lib.rs
feuerste: Good idea! Done, I integrated it into this PR.

PR #414, opened 2020-02-28
Update to latest nav_types with latest nalgebra.

feuerste:
This way we can get rid of two internal versions of nalgebra.


feuerste:
Thanks for your review!

PR #415, opened 2020-03-03
Replace cgmath and collision with nalgebra

nnmm:
* nalgebra has an `Isometry3` that can replace our hand-rolled `Isometry3`
* nalgebra does not provide an AABB type or something like [`cgmath::Perspective`](https://docs.rs/cgmath/0.17.0/cgmath/struct.Perspective.html), so we add those ourselves.
* While we're at it, consistently use the `Point3` type for point coordinates – this matches the intended semantics and gets rid of a bunch of conversions.

Criterion output against master (the octree queries will improve when we move to using `Relation` as opposed to a `bool` for intersection tests in an upcoming PR):
```
all_query_octree        time:   [25.687 ms 26.782 ms 28.270 ms]
                        change: [-3.5701% +1.8095% +7.7862%] (p = 0.53 > 0.05)
                        No change in performance detected.
Found 4 outliers among 100 measurements (4.00%)
  4 (4.00%) high severe


all_query_s2            time:   [19.869 ms 20.372 ms 20.974 ms]
                        change: [-1.0242% +2.5209% +6.2368%] (p = 0.17 > 0.05)
                        No change in performance detected.


box_query_octree        time:   [28.085 ms 28.390 ms 28.760 ms]
                        change: [-1.6742% +0.0614% +1.7783%] (p = 0.94 > 0.05)
                        No change in performance detected.


box_query_s2            time:   [20.913 ms 21.431 ms 22.051 ms]
                        change: [-0.3591% +3.5221% +7.3647%] (p = 0.07 > 0.05)
                        No change in performance detected.


frustum_query_octree    time:   [8.9604 ms 9.1048 ms 9.2342 ms]
                        change: [+1.9397% +4.2781% +6.7289%] (p = 0.00 < 0.05)
                        Performance has regressed.
Found 7 outliers among 100 measurements (7.00%)
  3 (3.00%) low mild
  3 (3.00%) high mild
  1 (1.00%) high severe


frustum_query_s2        time:   [1.9423 ms 2.2494 ms 2.6056 ms]
                        change: [+6.5215% +16.341% +26.731%] (p = 0.00 < 0.05)
                        Performance has regressed.
Found 13 outliers among 100 measurements (13.00%)
  5 (5.00%) low mild
  4 (4.00%) high mild
  4 (4.00%) high severe


obb_query_octree        time:   [18.768 ms 19.075 ms 19.410 ms]
                        change: [-5.5102% -3.4219% -1.4232%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 1 outliers among 100 measurements (1.00%)
  1 (1.00%) high severe


obb_query_s2            time:   [10.984 ms 11.296 ms 11.594 ms]
                        change: [-2.8831% +0.3239% +3.7390%] (p = 0.86 > 0.05)
                        No change in performance detected.
Found 13 outliers among 100 measurements (13.00%)
  3 (3.00%) low severe
  4 (4.00%) low mild
  1 (1.00%) high mild
  5 (5.00%) high severe


cell_union_query_octree time:   [12.990 ms 13.454 ms 13.929 ms]
                        change: [-13.005% -9.8551% -7.0474%] (p = 0.00 < 0.05)
                        Performance has improved.

cell_union_query_s2     time:   [295.79 us 298.04 us 299.97 us]
                        change: [-26.297% -22.423% -19.211%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 1 outliers among 100 measurements (1.00%)
  1 (1.00%) high severe

```


krzesi-mir:
> * nalgebra does not provide an AABB type or something like [`cgmath::Perspective`](https://docs.rs/cgmath/0.17.0/cgmath/struct.Perspective.html), so we add those ourselves.

`ncollide`, which is part of the same suite as `nalgebra`, provides `AABB` type https://docs.rs/ncollide/0.14.1/ncollide/bounding_volume/struct.AABB.html

nnmm:
True, I used that before, but ncollide is a very heavy dependency, if we're only going to use that, I don't think it's worth it. The Aabb struct is trivial.

krzesi-mir:
> True, I used that before, but ncollide is a very heavy dependency, if we're only going to use that, I don't think it's worth it. The Aabb struct is trivial.

Makes sense

src/octree/generation.rs
```diff
@@ -296,18 +289,24 @@ pub fn build_octree_from_file(
...
     resolution: f64,
-    bounding_box: Aabb3<f64>,
+    bounding_box: Aabb<f64>,
     input: impl Iterator<Item = PointsBatch> + NumberOfPoints + Send,
     attributes: &[&str],
 ) {
     attempt_increasing_rlimit_to_max();
 
-    // TODO(ksavinash9): This function should return a Result.
+    let attribute_data_types = vec![
```
nnmm: Made a small orthogonal change here to be more explicit than `Default::default()`. I'm fine with reverting it if you prefer.

src/octree/mod.rs
```diff
@@ -155,15 +141,22 @@ impl Octree {
                 meta_proto.version, CURRENT_VERSION
             );
         }
+        // These are hardcoded for now.
+        let attribute_data_types = vec![
```
nnmm: Same change as in src/octree/generation.rs

src/octree/mod.rs
```diff
@@ -224,7 +217,11 @@ impl Octree {
     }
 
     pub fn get_visible_nodes(&self, projection_matrix: &Matrix4<f64>) -> Vec<NodeId> {
-        let frustum = collision::Frustum::from_matrix4(*projection_matrix).unwrap();
+        let frustum =
```
nnmm: This is interesting: We now use our own `Frustum`'s intersection testing. Axes are precomputed.

src/octree/mod.rs
```diff
@@ -309,12 +307,25 @@ impl PointCloud for Octree {
     type Id = NodeId;
 
     fn nodes_in_location(&self, location: &PointLocation) -> Vec<Self::Id> {
-        let culling = location.get_point_culling();
-        let filter_func = move |node_id: &NodeId, octree: &Octree| -> bool {
-            let current = &octree.nodes[&node_id];
-            culling.intersects_aabb3(&current.bounding_cube.to_aabb3())
+        let node_ids_for_intersector = |isec: Intersector<f64>| {
```
nnmm: This doesn't use PointCulling anymore.

src/read_write/node_writer.rs
```diff
@@ -213,6 +213,15 @@ impl WriteLE for Vec<Vector3<f64>> {
     }
 }
 
+impl WriteLE for Vec<Point3<f64>> {
```
nnmm: This is now needed because our positions are saved as `Point3` instead of `Vector3`.

src/read_write/s2.rs
```diff
@@ -160,11 +160,12 @@ where
         }
     }
 
-    pub fn get_meta(self) -> S2Meta {
-        S2Meta::new(
+    pub fn get_meta(self) -> Option<S2Meta> {
```
nnmm: Again, a sort of orthogonal change.

src/s2_cells/mod.rs
```diff
@@ -216,23 +214,16 @@ impl S2Cells {
         self.meta.to_proto()
     }
 
-    /// Wrapper arround cells_in_convex_hull for Obbs
-    fn cells_in_cuboid<T>(&self, poly: &T) -> Vec<CellID>
+    /// Returns all cells that intersect this convex polyhedron
+    fn cells_in_convex_polyhedron<T>(&self, poly: &T) -> Vec<CellID>
```
nnmm: Deleted a useless wrapper function here

xray/src/generation.rs
```diff
@@ -119,10 +118,12 @@ pub trait ColoringStrategy: Send {
             // This means that the y-axis aligns too, but the origin of the image space must be at the
             // bottom left. Since images have their origin at the top left, we need actually have to
             // invert y and go from the bottom of the image.
-            let x = (((pos.x - bbox.min().x) / bbox.dim().x) * f64::from(image_size.x)) as u32;
-            let y =
-                ((1. - ((pos.y - bbox.min().y) / bbox.dim().y)) * f64::from(image_size.y)) as u32;
-            let z = (((pos.z - bbox.min().z) / bbox.dim().z) * NUM_Z_BUCKETS) as u32;
+            let x = (((pos.x - bbox.min().x) / (bbox.max().x - bbox.min().x))
```
nnmm: Should I do this with vector math?

Cargo.toml
```diff
@@ -13,40 +13,34 @@
 # limitations under the License.
 
 [package]
+authors = ["Holger Rapp <hrapp@google.com>"]
```
feuerste: It seems this PR undos a lot of my authors/contributors/version changes....

nnmm: That was accidental, I noticed and fixed it

feuerste: Fixed locally? At least I can't see the fix here...

nnmm:
@feuerste Sorry, I indeed made merge errors. Git is dumber than I thought. I will double-check that after the next commit, this PR has all your changes.

nnmm:
I love working from home and forgetting to commit with my Lyft email address

src/geometry/aabb.rs
```diff
@@ -1,72 +1,163 @@
...
+        &self.maxs
+    }
+
+    pub fn grow(&mut self, p: Point3<S>) {
+        self.mins = nalgebra::inf(&self.mins, &p);
+        self.maxs = nalgebra::sup(&self.maxs, &p);
+    }
+
+    pub fn contains(&self, p: &Point3<S>) -> bool {
+        nalgebra::partial_le(&self.mins, p) && nalgebra::partial_le(p, &self.maxs)
```
feuerste: `AABB3` of `collision` actually implements it like:
```rust
nalgebra::partial_le(&self.mins, p) && nalgebra::partial_lt(p, &self.maxs)
```
see https://docs.rs/collision/0.20.1/src/collision/volume/aabb/aabb3.rs.html#105-111

src/iterator.rs
```diff
@@ -31,8 +29,8 @@ impl PointLocation {
     pub fn get_point_culling(&self) -> Box<dyn PointCulling<f64>> {
         match &self {
             PointLocation::AllPoints => Box::new(AllPoints {}),
-            PointLocation::Aabb(aabb) => Box::new(*aabb),
-            PointLocation::Frustum(frustum) => Box::new(frustum.clone()),
+            PointLocation::Aabb(aabb) => Box::new(aabb.clone()),
```
krzesi-mir: nit: remove clone if you choose to derive `Copy`

src/iterator.rs
```diff
@@ -60,7 +58,7 @@ macro_rules! with_point_culling {
         match &$point_location {
             PointLocation::AllPoints => $closure(crate::math::AllPoints {}),
             PointLocation::Aabb(aabb) => $closure(aabb.clone()),
```
krzesi-mir: nit: remove clone if you choose to derive `Copy`

sdl_viewer/src/box_drawer.rs
```diff
@@ -164,13 +164,13 @@ impl BoxDrawer {
...
         &self,
-        cuboid: &Aabb3<f64>,
+        cuboid: &Aabb<f64>,
         world_to_gl: &Matrix4<f64>,
         color: &color::Color<f32>,
     ) {
-        let dim = cuboid.dim() / 2.0;
-        let scale_matrix = Matrix4::from_nonuniform_scale(dim.x, dim.y, dim.z);
-        let translation_matrix = Matrix4::from_translation(cuboid.center().to_vec());
+        let dim = cuboid.max() - cuboid.center();
```
krzesi-mir: Do
```suggestion
        let dim = cuboid.dim() / 2.0;
```
if you decide to implement `Aabb::dim()`

src/geometry/frustum.rs
```diff
@@ -1,67 +1,190 @@
...
+            r1c0, r1c1, r1c2, r1c3,
+            r2c0, r2c1, r2c2, r2c3,
+            r3c0, r3c1, r3c2, r3c3,
+        );
+        Self { matrix }
+    }
+
+    // This emulates cgmath::PerspectiveFov, which is more restricted
+    // and corresponds to nalgebra::Perspective.
+    pub fn new_fov(fovy: S, aspect: S, near: S, far: S) -> Self {
```
krzesi-mir: nit: since we are using `nalgebra` now, should we reorder the parameters to be on par with https://www.nalgebra.org/rustdoc/nalgebra/geometry/struct.Perspective3.html#method.new ?
i.e.
```suggestion
    pub fn new_fov(aspect: S, fovy: S, near: S, far: S) -> Self {
```

src/geometry/aabb.rs
```diff
@@ -1,72 +1,163 @@
...
+
+    pub fn max(&self) -> &Point3<S> {
+        &self.maxs
+    }
+
+    pub fn grow(&mut self, p: Point3<S>) {
+        self.mins = nalgebra::inf(&self.mins, &p);
+        self.maxs = nalgebra::sup(&self.maxs, &p);
+    }
+
```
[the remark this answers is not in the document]
nnmm: I mostly disliked the name – how about `diag()`?

src/math/frustum.rs
```diff
@@ -0,0 +1,220 @@
...
+
+    #[derive(Debug, Clone, Serialize, Deserialize)]
+    pub struct Perspective<S: RealField> {
+        matrix: Matrix4<S>,
+    }
+
+    impl<S: RealField> Perspective<S> {
+        pub fn new(left: S, right: S, bottom: S, top: S, near: S, far: S) -> Self {
+            assert!(
+                left <= right,
```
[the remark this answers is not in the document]
nnmm: That's correct, I pasted these asserts from cgmath, but strict is better

src/math/frustum.rs
```diff
@@ -0,0 +1,219 @@
...
+                r1c0, r1c1, r1c2, r1c3,
+                r2c0, r2c1, r2c2, r2c3,
+                r3c0, r3c1, r3c2, r3c3,
+            );
+            Self { matrix }
+        }
+
+        // This emulates cgmath::PerspectiveFov, which more restricted
+        // and corresponds to nalgebra::Perspective.
+        pub fn new_fov(fovy: S, aspect: S, near: S, far: S) -> Self {
```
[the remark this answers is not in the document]
nnmm: Done. Checking of near and far is done in the other constructor that is called.

src/math/aabb.rs
```diff
@@ -0,0 +1,133 @@
...
+where
+    S: RealField,
+{
+    fn compute_corners(&self) -> [Point3<S>; 8] {
+        self.corners()
+    }
+
+    fn compute_edges(&self) -> ArrayVec<[Unit<Vector3<S>>; 6]> {
+        let mut edges = ArrayVec::new();
+        edges.push(Vector3::x_axis());
```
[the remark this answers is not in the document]
nnmm: The arrays have a different length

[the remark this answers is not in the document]
nnmm: So we could do ArrayVec::from(Self::axes()).into_iter().collect(), but that's probably not better

src/geometry/frustum.rs
nnmm: Hmm, no strong opinions. Having this be a drop-in for cgmath::Perspective makes the diff smaller. And I'd have to be careful when reordering those as they are of the same type.

sdl_viewer/src/terrain_drawer/layer.rs
```diff
@@ -218,10 +219,18 @@ impl GridCoordinateFrame {
     /// Returns the terrain pos (i.e. the coordinate of the lower corner of the terrain) for
     /// a given camera position (in the world coordinate system).
     fn terrain_pos_for_camera_pos(&self, world_pos: Point3<f64>) -> Vector2<i64> {
-        let local_pos = &self.terrain_from_world * &world_pos;
+        let local_pos = self.terrain_from_world * world_pos;
         let x = ((local_pos.x - self.u_origin.value.x) / self.u_resolution_m.value).floor();
         let y = ((local_pos.y - self.u_origin.value.y) / self.u_resolution_m.value).floor();
-        (Vector2::new(x, y)).cast().unwrap() - self.texture_half_extent
+        assert!(
```
feuerste: Would it be possible to reuse the asserts of `convert_terrain_pos`?

src/math/aabb.rs
```diff
@@ -0,0 +1,133 @@
...
+    /// having to reference a specific Aabb instance.
+    pub fn axes() -> [Unit<Vector3<S>>; 3] {
+        [Vector3::x_axis(), Vector3::y_axis(), Vector3::z_axis()]
+    }
+}
+
+impl From<&proto::AxisAlignedCuboid> for Aabb<f64> {
+    fn from(aac: &proto::AxisAlignedCuboid) -> Self {
+        let aac_min = aac.min.clone().unwrap_or_else(|| {
+            let deprecated_min = aac.deprecated_min.clone().unwrap(); // Version 9
```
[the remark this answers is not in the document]
nnmm: I tried changing it, but it's actually not an Option. There's probably a way to do it, but it's copy-pasted from the conversion code in point_viewer_proto_rust, so we could do it all in one PR later if we choose to do so.

src/geometry/aabb.rs
feuerste: This would in particular be useful when sharding neighboring bounding boxes.

src/math/aabb.rs
```diff
@@ -0,0 +1,133 @@
...
+    }
+
+    pub fn grow(&mut self, p: Point3<S>) {
+        self.mins = nalgebra::inf(&self.mins, &p);
+        self.maxs = nalgebra::sup(&self.maxs, &p);
+    }
+
+    pub fn corners(&self) -> [Point3<S>; 8] {
+        [
+            self.mins,
```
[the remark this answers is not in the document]
nnmm: Done

src/geometry/frustum.rs
```diff
@@ -1,97 +1,201 @@
...
 
 impl<S> ConvexPolyhedron<S> for Frustum<S>
 where
-    S: BaseFloat,
+    S: RealField,
 {
+    #[rustfmt::skip]
     fn compute_corners(&self) -> [Point3<S>; 8] {
-        let corner_from = |x, y, z| self.query_from_clip.transform_point(Point3::new(x, y, z));
+        let corner_from = |x, y, z| self.world_from_clip.transform_point(&Point3::new(x, y, z));
```
feuerste: I would keep `query_from_clip`

src/math/frustum.rs
```diff
@@ -0,0 +1,219 @@
...
+    use nalgebra::{Matrix4, RealField};
+    use serde::{Deserialize, Serialize};
+
+    #[derive(Debug, Clone, Serialize, Deserialize)]
+    pub struct Perspective<S: RealField> {
+        matrix: Matrix4<S>,
+    }
+
+    impl<S: RealField> Perspective<S> {
+        pub fn new(left: S, right: S, bottom: S, top: S, near: S, far: S) -> Self {
```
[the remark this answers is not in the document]
nnmm: Done

src/math/base.rs
```diff
@@ -0,0 +1,9 @@
+use nalgebra::{Point3, RealField};
+use std::fmt;
+
+pub trait PointCulling<S>: fmt::Debug + Sync + Send
```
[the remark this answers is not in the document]
nnmm: That was carried over from master. I know Debug isn't nedeed, and it was already removed since you reviewed, I also removed the other two.

src/geometry/frustum.rs
nnmm: Oops, again a merge gone wrong.

src/geometry/aabb.rs
```diff
@@ -1,72 +1,163 @@
...
-use cgmath::{BaseFloat, Point3, Vector3};
-use collision::{Aabb, Contains};
-use num_traits::Bounded;
+use nalgebra::{Isometry3, Point3, RealField, Unit, Vector3};
+use serde::{Deserialize, Serialize};
 
-pub use collision::Aabb3;
+/// An axis-aligned bounding box.
+#[derive(Debug, PartialEq, Clone, Serialize, Deserialize)]
+pub struct Aabb<S: RealField> {
```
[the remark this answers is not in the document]
nnmm: Would you prefer it if I do that?

src/geometry/aabb.rs
```diff
@@ -1,72 +1,163 @@
...
 use arrayvec::ArrayVec;
-use cgmath::{BaseFloat, Point3, Vector3};
-use collision::{Aabb, Contains};
-use num_traits::Bounded;
+use nalgebra::{Isometry3, Point3, RealField, Unit, Vector3};
+use serde::{Deserialize, Serialize};
 
-pub use collision::Aabb3;
+/// An axis-aligned bounding box.
+#[derive(Debug, PartialEq, Clone, Serialize, Deserialize)]
```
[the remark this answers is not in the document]
nnmm: Hmm, all other PointLocation types (Obb, Frustum) are not Copy as well, and the same applies to them. I think it's nice to encourage passing by reference since for an Aabb3 it's probably more efficient (48 bytes).

sdl_viewer/src/box_drawer.rs
nnmm: Done

xray/src/generation.rs
```diff
@@ -526,10 +527,11 @@ pub fn xray_from_points(
     true
 }
 
-fn find_quadtree_bounding_rect_and_levels(bbox: &Aabb3<f64>, tile_size_m: f64) -> (Rect, u8) {
+fn find_quadtree_bounding_rect_and_levels(bbox: &Aabb<f64>, tile_size_m: f64) -> (Rect, u8) {
     let mut levels = 0;
     let mut cur_size = tile_size_m;
-    while cur_size < bbox.dim().x || cur_size < bbox.dim().y {
+    let dim = bbox.max() - bbox.min();
```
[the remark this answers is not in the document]
nnmm: Refactored for clarity and efficiency. @feuerste fyi

src/geometry/aabb.rs
nnmm: Thanks, changed it.

nnmm:
ptal @krzesi-mir @feuerste

src/geometry/aabb.rs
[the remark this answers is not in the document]
nnmm: Implemented it as diag()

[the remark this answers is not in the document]
feuerste: Sorry to chime in so late. When I first read `diag()` I thought it gives the diagonal. I think `dimensions`, or, even better, `extents` would fit better, what do you think?

nnmm:
Just added some benchmarks to the description, performance improves a lot with this!

src/geometry/aabb.rs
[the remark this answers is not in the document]
nnmm: It does give the diagonal (you probably expected the diagonal's length?). I'm fine with the other options too.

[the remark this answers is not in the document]
feuerste: Ah, I see, I thought you wanted to have a method giving the dimensions for each axis.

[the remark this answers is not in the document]
feuerste: Then `diag` is fine

src/geometry/frustum.rs
nnmm: Resolved.

src/geometry/aabb.rs
```diff
@@ -1,72 +1,170 @@
...
 
     fn intersector(&self) -> Intersector<S> {
         let mut unit_axes = ArrayVec::new();
-        unit_axes.push(Vector3::unit_x());
-        unit_axes.push(Vector3::unit_y());
-        unit_axes.push(Vector3::unit_z());
-        let corners = self.compute_corners();
+        unit_axes.push(Vector3::x_axis());
+        unit_axes.push(Vector3::y_axis());
+        unit_axes.push(Vector3::z_axis());
```
krzesi-mir: how about:
```suggestion
unit_axes
            .try_extend_from_slice(&[Vector3::x_axis(), Vector3::y_axis(), Vector3::z_axis()])
            .unwrap();
```

feuerste:
Do you know, why the frustum performance became worse?

src/math/base.rs
```diff
@@ -0,0 +1,11 @@
+use crate::geometry::Aabb;
+use nalgebra::{Point3, RealField};
+
+pub trait PointCulling<S>
+where
+    S: RealField,
+{
+    fn contains(&self, point: &Point3<S>) -> bool;
+
```
krzesi-mir: nit: remove empty line?

src/octree/mod.rs
```diff
@@ -99,8 +100,8 @@ pub fn to_meta_proto(octree_meta: &OctreeMeta, nodes: Vec<proto::OctreeNode>) ->
 // TODO(hrapp): something is funky here. "r" is smaller on screen than "r4" in many cases, though
 // that is impossible.
 fn project(m: &Matrix4<f64>, p: &Point3<f64>) -> Point3<f64> {
-    let q = m * Point3::to_homogeneous(*p);
-    Point3::from_homogeneous(q / q.w)
+    let q = m * p.to_homogeneous();
+    Point3::from_homogeneous(q).unwrap()
```
krzesi-mir: we don't divide by `q.w` anymore?

src/octree/mod.rs
```diff
@@ -130,7 +131,7 @@ fn relative_size_on_screen(bounding_cube: &Cube, matrix: &Matrix4<f64>) -> f64 {
         Point3::new(max.x, min.y, max.z),
         Point3::new(min.x, max.y, max.z),
     ] {
-        rv = rv.grow(clip_point_to_hemicube(&project(matrix, p)));
+        rv.grow(clip_point_to_hemicube(&project(matrix, p)));
     }
     (rv.max().x - rv.min().x) * (rv.max().y - rv.min().y)
```
krzesi-mir: how about using `Aabb::diag` here?
```suggestion
    let diag = rv.diag();
    diag.x * diag.y
```

src/read_write/s2.rs
feuerste: I don't understand this change. It just always returns `Some` and never `None`?

xray/src/lib.rs
```diff
@@ -101,25 +101,32 @@ impl Meta {
...
+        };
         let frustum =
             Frustum::from_matrix4(matrix).ok_or("Unable to create frustum from matrix")?;
+        let frustum_isec = frustum
+            .intersector()
+            .cache_separating_axes(&Aabb::axes(), &Aabb::axes());
         let mut result = Vec::new();
         let mut open = vec![Node::root_with_bounding_rect(self.bounding_rect.clone())];
         while !open.is_empty() {
             let node = open.pop().unwrap();
```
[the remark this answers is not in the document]
feuerste: I already merged this in another PR.

src/math/mod.rs
```diff
@@ -124,170 +111,81 @@ where
...
-            self.rotation.conjugate(),
-            -(self.rotation.conjugate() * self.translation),
-        )
+    let aabb_corner_cells = aabb
+        .compute_corners()
+        .iter()
+        .map(|p| CellID::from_point(p))
+        .collect();
+    let mut aabb_cell_union = CellUnion(aabb_corner_cells);
+    aabb_cell_union.normalize();
```
krzesi-mir: nit: combine these two lines into one?
```suggestion
    let aabb_cell_union = CellUnion(aabb_corner_cells).normalize();
```

src/math/mod.rs
```diff
@@ -296,76 +194,57 @@ where
...
-        assert_eq!(zero_obb.intersects_aabb3(&bbox), true);
-        assert_eq!(fourty_five_deg_obb.separating_axes.axes.len(), 5);
-        assert_eq!(fourty_five_deg_obb.intersects_aabb3(&bbox), false);
-        assert_eq!(arbitrary_obb.separating_axes.axes.len(), 15);
+    fn test_inverse() {
+        let persp = Perspective::new(-0.123, 0.45, 0.04, 0.75, 1.0, 4.0);
+        let reference_inverse = persp.as_matrix().try_inverse().unwrap();
+        let inverse = persp.inverse();
+        let diff = (reference_inverse - inverse).abs();
+        assert!(diff.max() < 0.000001, "diff.max() is {}", diff.max());
```
krzesi-mir: nit: `1e-6` instead of `0.000001`?

sdl_viewer/src/terrain_drawer/layer.rs
```diff
@@ -178,6 +169,19 @@ impl TerrainLayer {
             color: self.color_tiles.load(min_x, min_y, width, height),
         }
     }
+
+    // Helper function because OpenGL doesn't like i64
+    fn convert_terrain_pos(v: Vector2<i64>) -> Vector2<f64> {
```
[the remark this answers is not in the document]
nnmm: Good idea, done

sdl_viewer/src/terrain_drawer/layer.rs
nnmm: No, I don't think so. One is comparing floats and one ints, and they have different safe ranges.

point_viewer_grpc/src/bin/query_frustum.rs
```diff
@@ -122,9 +122,9 @@ fn main() {
             let last_num_points = points.len();
             for (position, color) in reply.positions.iter().zip(reply.colors.iter()) {
                 let p = Point3::new(position.x, position.y, position.z);
-                bounding_box = bounding_box.grow(p);
+                bounding_box.grow(p);
                 points.push(Point {
-                    position: p.to_vec(),
+                    position: p,
```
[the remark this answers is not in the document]
nnmm: Done.

point_cloud_test/src/synthetic_data.rs
```diff
@@ -67,7 +63,7 @@ impl Iterator for SyntheticData {
         }
         let pos = self.next_pos();
         let point = Point {
-            position: pos.to_vec(),
+            position: pos,
```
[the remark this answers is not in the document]
nnmm: Done.

src/geometry/aabb.rs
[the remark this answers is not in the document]
nnmm: I know, but you're fine with not doing that now?

src/geometry/frustum.rs
```diff
@@ -1,97 +1,209 @@
...
+        self.separating_axes.intersect(&aabb.compute_corners()) != Relation::Out
     }
 }
 
 impl<S> ConvexPolyhedron<S> for Frustum<S>
 where
-    S: BaseFloat,
+    S: RealField,
 {
+    #[rustfmt::skip]
```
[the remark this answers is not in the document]
nnmm: I tried that first, but `attributes on expressions are experimental`.

nnmm:
@feuerste re: Frustum performance, collision used an AABB intersection test based on [intersection testing the box with all six faces](https://docs.rs/collision/0.20.1/src/collision/frustum.rs.html#79). While this is maybe a bit faster, it's also [inaccurate](https://www.gamedev.net/forums/topic/672043-perfect-aabb-frustum-intersection-test/5254266/). I'd rather have a correct intersection test than the fastest one, and we can reduce the number of intersection tests anyway by returning the full Relation instead of a bool.

src/math/mod.rs
nnmm: The function is unfortunately written as a mutation, i.e. it returns `()`.

src/geometry/obb.rs
```diff
@@ -36,38 +35,41 @@ impl<S: BaseFloat + Bounded> CachedAxesObb<S> {
...
 
-impl<S: BaseFloat> From<&Aabb3<S>> for Obb<S> {
-    fn from(aabb: &Aabb3<S>) -> Self {
+impl<S: RealField> From<&Aabb<S>> for Obb<S> {
+    fn from(aabb: &Aabb<S>) -> Self {
         Obb::new(
-            Isometry3::new(Quaternion::one(), EuclideanSpace::to_vec(aabb.center())),
-            aabb.dim() / (S::one() + S::one()),
+            Isometry3::from_parts(aabb.center().coords.into(), UnitQuaternion::identity()),
+            aabb.max() - aabb.center(),
```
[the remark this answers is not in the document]
nnmm: Done.

src/math/mod.rs
nnmm: Done.

src/read_write/s2.rs
nnmm: There's a `?` after bounding_box.

src/math/base.rs
nnmm: Done.

src/geometry/aabb.rs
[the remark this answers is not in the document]
nnmm: Not completely sure what the difference is between "a method giving the dimensions for each axis" and `diag()`, but it seems we're in agreement :)

src/iterator.rs
nnmm: I'm resolving those – if we implement `Copy`, which I wouldn't unless you have a clear preference, clippy will remind me to do that anyway.

src/geometry/frustum.rs
```diff
@@ -1,97 +1,209 @@
...
+            fovy > S::zero() && fovy < S::pi(),
+            "`fovy` must be a number between 0 and π, found: {:?}",
+            fovy
+        );
+        assert!(
+            aspect > S::zero(),
+            "`aspect` must be a positive number, found: {:?}",
+            aspect
+        );
+        let angle = nalgebra::convert::<f64, S>(0.5) * fovy;
```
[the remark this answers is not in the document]

sdl_viewer/src/terrain_drawer/layer.rs
```diff
@@ -218,10 +219,18 @@ impl GridCoordinateFrame {
     /// Returns the terrain pos (i.e. the coordinate of the lower corner of the terrain) for
     /// a given camera position (in the world coordinate system).
     fn terrain_pos_for_camera_pos(&self, world_pos: Point3<f64>) -> Vector2<i64> {
```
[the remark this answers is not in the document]
nnmm: Done.

sdl_viewer/src/terrain_drawer/layer.rs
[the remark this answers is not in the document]
nnmm: Done. I think it's an open question what's better in general (https://www.forrestthewoods.com/blog/should-small-rust-structs-be-passed-by-copy-or-by-borrow/ is a pretty incomplete attempt to answer it), but I checked and rustc does pass by value where it's sensible even if you write a function as taking a reference.

sdl_viewer/src/terrain_drawer/layer.rs
```diff
@@ -49,7 +48,8 @@ impl TerrainLayer {
         // Initial terrain pos
         let terrain_pos: Vector2<i64> =
             grid_coordinates.terrain_pos_for_camera_pos(Point3::new(0.0, 0.0, 0.0));
-        let u_terrain_pos = GlUniform::new(&program, "terrain_pos", terrain_pos.cast().unwrap());
+        let float_terrain_pos: Vector2<f64> = Self::convert_terrain_pos(terrain_pos);
```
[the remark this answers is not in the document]
nnmm: Done

xray/src/lib.rs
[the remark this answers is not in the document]
nnmm: Merged master.

src/geometry/frustum.rs
```diff
@@ -1,97 +1,209 @@
...
+
+impl<S: RealField> Frustum<S> {
     pub fn new(query_from_eye: Isometry3<S>, clip_from_eye: Perspective<S>) -> Self {
-        let eye_from_query: Decomposed<Vector3<S>, Quaternion<S>> = query_from_eye.inverse().into();
-        let clip_from_query =
-            Matrix4::<S>::from(clip_from_eye) * Matrix4::<S>::from(eye_from_query);
-        let query_from_clip = clip_from_query.inverse_transform().unwrap();
-        let frustum = collision::Frustum::from_matrix4(clip_from_query).unwrap();
+        let clip_from_query = clip_from_eye.as_matrix() * query_from_eye.inverse().to_homogeneous();
+        let query_from_clip = query_from_eye.to_homogeneous() * clip_from_eye.inverse();
```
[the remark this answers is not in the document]
nnmm: This should be more efficient.

src/math/base.rs
nnmm: Done.

src/octree/mod.rs
nnmm: Done.

src/geometry/aabb.rs
```diff
@@ -1,72 +1,170 @@
...
+            u.grow(transform.transform_point(&corner));
+            u
+        })
+    }
+
+    /// It's convenient to have this associated function for intersection testing.
+    /// To be precise, it's nice if we can give this to cache_separating_axes() without
+    /// having to reference a specific Aabb instance.
+    pub fn axes() -> [Unit<Vector3<S>>; 3] {
+        [Vector3::x_axis(), Vector3::y_axis(), Vector3::z_axis()]
```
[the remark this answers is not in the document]
nnmm: I might have a better idea: The only place where this function was used was in a call to `cache_separating_axes` for AABBs. I specialized that function for AABBs, `cache_separating_axes_for_aabb`, which now doesn't take an argument anymore and is easier to use.

src/geometry/aabb.rs
nnmm: Tbh I'd prefer what we have now since it's easier to understand. Let me know if you see a reason why it should use the slice method.

src/geometry/frustum.rs
nnmm: For my own understanding and yours, let's take a step back and look at where PerspectiveFOV was used to understand how to replace it:
* In `point_cloud_test` where it's immediately converted to the more general `Perspective`
* In `point_viewer_grpc` where it's immediately converted to the more general `Perspective`
* In `sdl_viewer` where it's immediately converted to a matrix
* (avmapping-geometry) In `point_cloud_client_cc` where it's immediately converted to a matrix

We can replace the latter two with `nalgebra::Perspective3`, but for the former two we need conversion from the special case (symmetric frustum)  to the general case (asymmetric frustum, i.e. `Perspective`). I did that via a "convenience constructor" but thinking about it again, I could implement a conversion from `nalgebra::Perspective3`. We don't need to define a new struct whose only purpose is to convert to `Perspective` at any rate.

nnmm: Done, I added a `From<nalgebra::Perspective3>` and added a test that checks that it's equal to `new_fov` :)

src/geometry/frustum.rs
```diff
@@ -1,97 +1,209 @@
...
 {
     fn contains(&self, point: &Point3<S>) -> bool {
-        match self.frustum.contains(point) {
-            Relation::Cross => true,
-            Relation::In => true,
-            Relation::Out => false,
-        }
+        let p_clip = self.frustum.clip_from_query.transform_point(point);
+        p_clip.coords.min() > nalgebra::convert(-1.0)
+            && p_clip.coords.max() < nalgebra::convert(1.0)
```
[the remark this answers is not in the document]
nnmm: I'm fine with leaving it like this, we'll never need to tile space with frustums.

src/octree/mod.rs
nnmm: That's done by nalgebra, see https://docs.rs/nalgebra/0.20.0/nalgebra/geometry/struct.Point.html#method.from_homogeneous

sdl_viewer/src/camera.rs
```diff
@@ -96,10 +95,7 @@ impl Camera {
         height: i32,
         local_from_global: Option<Isometry3<f64>>,
     ) -> Self {
-        let local_from_global = local_from_global.map_or_else(One::one, |local_from_global| {
-            let decomposed: Decomposed<Vector3<f64>, Quaternion<f64>> = local_from_global.into();
-            decomposed.into()
-        });
+        let local_from_global = local_from_global.unwrap_or_else(nalgebra::one);
```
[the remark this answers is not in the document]
nnmm: Done.

src/geometry/frustum.rs
```diff
@@ -1,97 +1,209 @@
...
+pub struct Perspective<S: RealField> {
+    matrix: Matrix4<S>,
+}
+
+impl<S: RealField> Perspective<S> {
+    /// Left, right, bottom, and top are in radians.
+    pub fn new(left: S, right: S, bottom: S, top: S, near: S, far: S) -> Self {
+        assert!(
+            left < right,
+            "`left` cannot be greater than `right`, found: left: {:?} right: {:?}",
```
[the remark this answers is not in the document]

point_cloud_client/src/lib.rs
```diff
@@ -98,10 +98,11 @@ impl<'a> PointCloudClientBuilder<'a> {
...
             .map(|location| self.data_provider_factory.generate_data_provider(location))
             .collect::<Result<Vec<Box<dyn DataProvider>>>>()?;
-        let mut aabb: Option<Aabb3<f64>> = None;
-        let unite = |bbox: &Aabb3<f64>, with: &mut Option<Aabb3<f64>>| {
-            let b = with.get_or_insert(*bbox);
-            *b = b.union(bbox);
+        let mut aabb: Option<Aabb<f64>> = None;
+        let unite = |bbox: &Aabb<f64>, with: &mut Option<Aabb<f64>>| {
+            let b = with.get_or_insert(bbox.clone());
+            b.grow(*bbox.min());
```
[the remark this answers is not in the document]
nnmm: I don't think it makes this code any easier to understand, since this function mutates, but I'd expect a `union` method to have an immutable signature, i.e. `fn(self, other: &Self) -> Self`.

I'm fine with implementing it if you prefer though.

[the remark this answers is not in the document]
feuerste: It's fine with me, you can also leave it..

xray/src/generation.rs
```diff
@@ -542,15 +554,11 @@ pub fn build_xray_quadtree(
     let _ = fs::create_dir(output_directory);
 
     let bounding_box = match &parameters.query_from_global {
```
feuerste: Nit: How about doing
```rust
let mut bounding_box = parameters.point_cloud_client.bounding_box().clone();
if let Some(query_from_global) = &parameters.query_from_global {
    bounding_box = bounding_box.transform(&query_from_global);
}
```
instead?

nnmm:
I checked the number of points returned for all queries in point_cloud_test, they are the same before and after this change. Btw, check_box_query_equality returns all of the 1000000 points, so it's not a very good test.

nnmm:
I built an octree from the same ply on master and this branch and compared the output.
* The number of files are the same
* The file names and their sizes are the same according to `diff <(ls -1s octree | sort -k 2) <(ls -1s octree_nalgebra | sort -k 2)`
* The metas are not the same

I'll look further into it.

feuerste:
> * The metas are not the same

One reason for this can be that the order of items in the proto repeated fields are rarely the same.

nnmm:
> One reason for this can be that the orders of items in the proto repeated fields are rarely the same.

Yep, I'll check that next.

nnmm:
So, they really are just permuted. And besides this, I also tested sdl_viewer, which works exactly like before (even feels a tad smoother).

I'll merge soon after I resolve the merge conflicts unless I'm hearing concerns.

krzesi-mir:
🎉

nnmm:
I tested xrays as well, found a small bug that is fixed with the latest commit.

PR #416, opened 2020-03-03
Streamline tests

nnmm:
* `build_big_test_octree()` was building an abnormal octree where more than `MAX_POINTS_PER_NODE` all had the same coordinate, basically splitting infinitely often (in practice only until a certain max depth, but enough to slow down the tests a lot).
* The test that used `build_big_test_octree()` was functionally equivalent to `test_batch_iterator()` as far as I understand, so there's no added value there.
* Factor out repeated code, remove redundant checks
* Add comments, use `assert_eq!` instead of `assert!` where applicable


nnmm:
meow @krzesi-mir

PR #417, opened 2020-03-03
Remove superfluous module hierarchy

nnmm:


nnmm:
ptal @krzesi-mir

src/octree/tests.rs
```diff
@@ -0,0 +1,235 @@
+
```
krzesi-mir: nit: remove empty line

nnmm:
I guess it doesn't pick up on the move because one layer of indentation is gone

src/octree/tests.rs
nnmm: Done

krzesi-mir:
> I guess it doesn't pick up on the move because one layer of indentation is gone

probably

PR #418, opened 2020-03-04
Delete smoke test in src/octree/generation.rs

nnmm:
This test doesn't even have an `assert!`, it is simply a smoke test that is subsumed by the `point_cloud_test` crate, here: https://github.com/googlecartographer/point_cloud_viewer/blob/123557a2e11a326aad63a2ae95b4a3dddfcd26b7/point_cloud_test/src/lib.rs#L55-L60


nnmm:
meow @krzesi-mir

PR #419, opened 2020-03-05
Tests for codec.rs

nnmm:
* Test that encoding and decoding a scalar results approximately in the original value.
* We cannot test vector encoding/decoding, because there are no vec3_(fixpoint)_decode methods
* `encode()` is unused, do not export it to hide that fact.

### DIFF
--- a/src/read_write/codec.rs
+++ b/src/read_write/codec.rs
@@ -87,7 +87,7 @@ where
     num::cast(value).unwrap()
 }
 
-pub fn encode<T>(value: f64, min: f64, edge_length: f64) -> T
+pub fn _encode<T>(value: f64, min: f64, edge_length: f64) -> T
 where
     T: num_traits::NumCast,
 {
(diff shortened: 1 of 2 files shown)


nnmm:
meow @catevita

src/read_write/codec.rs
```diff
@@ -87,7 +87,7 @@ where
     num::cast(value).unwrap()
 }
 
-pub fn encode<T>(value: f64, min: f64, edge_length: f64) -> T
+pub fn _encode<T>(value: f64, min: f64, edge_length: f64) -> T
```
krzesi-mir: You could also do `#[allow(dead_code)]` here. Not sure which one is nicer.

src/read_write/codec.rs
```diff
@@ -140,3 +140,67 @@ where
...
+
+    #[test]
+    fn plain_scalar() {
+        let value = 41.33333;
+        let min = 40.0;
+        let edge_length = 2.0;
+
+        let value_f32 = _encode::<f32>(value, min, edge_length);
+        let value_f32_decoded = decode(value_f32, min, edge_length);
+        assert!(
```
krzesi-mir: nit: how about we use [approx](https://docs.rs/approx/0.3.2/approx/) crate here and possibly other places?

src/read_write/codec.rs
```diff
@@ -140,3 +140,67 @@ where
...
+            (value_f32_decoded - value).abs() < 0.0000001,
```
krzesi-mir: nit: use `1e-7` instead of `0.0000001`? Similarly below.

src/read_write/codec.rs
nnmm: Probably a good idea, but I'd do that separately since I don't want to do it in this one place only and otherwise this PR becomes an "Introduce approx" PR. Leaving a comment in the code.

src/read_write/codec.rs
nnmm: I feel like _encode calls attention to this function being unused, which is nice

src/read_write/codec.rs
nnmm: Done

PR #420, opened 2020-03-05
Split off geometry & sat testing from math module

nnmm:
Things that represent geometric objects (I.e. `Frustum`, `Obb` and `Cube`) have been moved into a "geometry" module.
No functional changes at all.


src/math.rs
```diff
@@ -29,6 +29,9 @@ use std::fmt;
 use std::ops::Mul;
 use std::str::FromStr;
 
+pub mod sat;
+pub use sat::*;
```
catevita: nit: here there is sat, is it already in the dependencies?

src/math/sat.rs
```diff
@@ -0,0 +1,34 @@
+use cgmath::{BaseFloat, EuclideanSpace, Point3, Vector3};
```
catevita: found the missing file of the nit

nnmm: You were right, it was missing. I pushed it.

PR #421, opened 2020-03-06
Factor out filtering code into a provided method

nnmm:
The code for fetching filtered points batches in a node currently only differs in the part that gets us a NodeIterator. We can factor out the part that filters the points. This makes it easier to see where the two point cloud types actually differ, and reduces code duplication.

### DIFF
--- a/src/iterator.rs
+++ b/src/iterator.rs
@@ -135,37 +135,12 @@ where
         (self.func)(res)
     }
 
-    fn push_points_and_callback<I>(&mut self, node_iterator: I) -> Result<()>
-    where
-        I: Iterator<Item = PointsBatch>,
-    {
-        for mut batch in node_iterator {
-            self.buf.append(&mut batch)?;
-            while self.buf.position.len() >= self.batch_size {
-                self.callback()?;
-            }
+    fn push_points_and_callback(&mut self, mut batch: PointsBatch) -> Result<()> {
+        self.buf.append(&mut batch)?;
+        while self.buf.position.len() >= self.batch_size {
+            self.callback()?;
         }
         Ok(())
-    }
-
-    /// Why only a single node? Because the nodes are distributed to several `PointStream`s working
-    /// in parallel by the `ParallelIterator`.
-    fn stream_points_for_query_in_node<P: PointCloud>(
-        &mut self,
-        point_cloud: &P,
-        query: &'a PointQuery,
-        node_id: P::Id,
-        batch_size: usize,
-    ) -> Result<()> {
-        let culling = query.location.get_point_culling();
-        let filter_intervals = &query.filter_intervals;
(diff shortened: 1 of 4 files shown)


nnmm:
meow @catevita

src/iterator.rs
```diff
@@ -240,7 +256,7 @@ where
                     }) {
                         // TODO(nnmm): This crashes on error. We should bubble up an error.
```
catevita: while reformatting, should we get rid also of this todo as well?

src/iterator.rs
```diff
@@ -154,13 +149,38 @@ pub trait PointCloud: Sync {
...
+        &self,
+        attributes: &[&str],
         node_id: Self::Id,
         batch_size: usize,
-    ) -> Result<FilteredIterator<'a>>;
+    ) -> Result<NodeIterator>;
     fn bounding_box(&self) -> &Aabb3<f64>;
+
+    /// Return the points matching the query in the selected node.
+    /// Why only a single node? Because the nodes are distributed to several `PointStream`s working
```
catevita: i think you meant `Pointstreams`

src/iterator.rs
nnmm: It's already gone, this is solved by that change :)

src/iterator.rs
```diff
@@ -202,11 +222,12 @@ where
         let mut number_of_jobs = 0;
         self.point_clouds
             .iter()
-            .flat_map(|octree| {
-                std::iter::repeat(octree).zip(octree.nodes_in_location(&self.point_query.location))
+            .flat_map(|point_cloud| {
```
catevita: nice

src/iterator.rs
catevita: now i get it...

catevita: maybe  Pointstream instances would be nicer to read thoug

nnmm: It was kind of intentional – to highlight the name of the struct, without the plural. Do you prefer to change it?

nnmm: Changed to PointStream instances

src/iterator.rs
```diff
@@ -154,13 +149,38 @@ pub trait PointCloud: Sync {
     type Id: ToString + Send + Copy;
     fn nodes_in_location(&self, location: &PointLocation) -> Vec<Self::Id>;
     fn encoding_for_node(&self, id: Self::Id) -> Encoding;
-    fn points_in_node<'a>(
-        &'a self,
-        query: &'a PointQuery,
+    /// Return all points in the selected node.
```
catevita: shall the Point Cloud trait be moved also somewhere else while reformatting this part?

nnmm: I'd keep the PR small for now, but I can do it soon :)

nnmm:
Thanks for reviewing!

PR #424, opened 2020-03-07
Generalize Cuboid trait to ConvexPolyhedron, generalize SAT intersection testing

nnmm:
* Define a `ConvexPolyhedron` trait that provides everything we need to do intersection tests with other convex polyhedra.
* AABBs, OBBs and frustums are all convex polyhedra and so they are instances of that trait.
* We are especially interested in intersection tests with AABBs, since that's how we determine octree nodes for a query. Via SAT intersection testing, we can deal with the removal of `collision::Frustum` in https://github.com/googlecartographer/point_cloud_viewer/pull/415 because there is no equivalent in nalgebra, and as a nice side effect have one less way of how we do intersection testing.
  * Currently, `Frustum`'s `intersect_aabb3()` method still uses `collision::Frustum`, but will be switched to SAT in https://github.com/googlecartographer/point_cloud_viewer/pull/415
  * OBB intersection testing with AABBs was already implemented by SAT, but we separate out the precomputation result from the core struct
* The SAT intersection test is now capable of not only determining whether two objects intersect, but also whether the second is completely contained within the first. That is useful for e.g. skipping intersection tests for children of octree nodes that are completely within the query geometry.
* Better documentation for SAT intersection testing module.

### DIFF
--- a/src/geometry/cube.rs
+++ b/src/geometry/cube.rs
@@ -1,7 +1,51 @@
 //! An axis-aligned cube.
 
-use cgmath::{Point3, Vector3};
-use collision::{Aabb, Aabb3};
+use crate::math::sat::{self, ConvexPolyhedron, Relation};
+use crate::math::PointCulling;
+use arrayvec::ArrayVec;
+use cgmath::{BaseFloat, Point3, Vector3};
+use collision::{Aabb, Contains};
+use num_traits::Bounded;
+
+pub use collision::Aabb3;
+
+impl<S> PointCulling<S> for Aabb3<S>
+where
+    S: 'static + BaseFloat + Sync + Send + Bounded,
+{
+    fn contains(&self, p: &Point3<S>) -> bool {
+        Contains::contains(self, p)
+    }
+
+    fn intersects_aabb3(&self, aabb: &Aabb3<S>) -> bool {
+        sat::sat(
+            ArrayVec::from([Vector3::unit_x(), Vector3::unit_y(), Vector3::unit_z()]),
+            &self.to_corners(),
+            &aabb.to_corners(),
+        ) != Relation::Out
+    }
+}
+
+impl<S> ConvexPolyhedron<S> for Aabb3<S>
+where
+    S: BaseFloat,
+{
+    fn compute_corners(&self) -> [Point3<S>; 8] {
+        self.to_corners()
+    }
+
+    fn compute_edges(&self) -> ArrayVec<[Vector3<S>; 6]> {
+        let mut edges = ArrayVec::new();
+        edges.push(Vector3::unit_x());
+        edges.push(Vector3::unit_y());
+        edges.push(Vector3::unit_z());
+        edges
+    }
+
+    fn compute_face_normals(&self) -> ArrayVec<[Vector3<S>; 6]> {
+        self.compute_edges()
+    }
+}
 
 #[derive(Debug, Clone)]
 pub struct Cube {
--- a/src/geometry/cube.rs
+++ b/src/geometry/cube.rs
@@ -1,6 +1,6 @@
 //! An axis-aligned cube.
 
-use crate::math::sat::{self, ConvexPolyhedron, Relation};
+use crate::math::sat::{self, ConvexPolyhedron, Intersector, Relation};
 use crate::math::PointCulling;
 use arrayvec::ArrayVec;
 use cgmath::{BaseFloat, Point3, Vector3};
@@ -34,16 +34,17 @@ where
         self.to_corners()
     }
 
-    fn compute_edges(&self) -> ArrayVec<[Vector3<S>; 6]> {
-        let mut edges = ArrayVec::new();
-        edges.push(Vector3::unit_x());
-        edges.push(Vector3::unit_y());
-        edges.push(Vector3::unit_z());
-        edges
-    }
-
-    fn compute_face_normals(&self) -> ArrayVec<[Vector3<S>; 6]> {
-        self.compute_edges()
+    fn intersector(&self) -> Intersector<S> {
+        let mut unit_axes = ArrayVec::new();
+        unit_axes.push(Vector3::unit_x());
+        unit_axes.push(Vector3::unit_y());
+        unit_axes.push(Vector3::unit_z());
+        let corners = self.compute_corners();
+        Intersector {
+            corners,
+            edges: unit_axes.clone(),
+            face_normals: unit_axes,
+        }
     }
 }
 
--- a/src/geometry/frustum.rs
+++ b/src/geometry/frustum.rs
@@ -113,7 +113,7 @@ where
         face_normals.push((edges[0].cross(edges[2])).normalize()); // Lower side
         face_normals.push((edges[0].cross(edges[3])).normalize()); // Upper side
         face_normals.push((edges[1].cross(edges[2])).normalize()); // Left side
-        face_normals.push((edges[1].cross(edges[2])).normalize()); // right side
+        face_normals.push((edges[1].cross(edges[4])).normalize()); // right side
 
         Intersector {
             corners,
--- a/src/geometry/frustum.rs
+++ b/src/geometry/frustum.rs
@@ -1,7 +1,11 @@
 //! An asymmetric frustum with an arbitrary 3D pose.
 
-use crate::math::{Cuboid, Isometry3, PointCulling};
-use cgmath::{BaseFloat, Decomposed, Matrix4, Perspective, Point3, Quaternion, Transform, Vector3};
+use crate::math::sat::{ConvexPolyhedron, Intersector};
+use crate::math::{Isometry3, PointCulling};
+use arrayvec::ArrayVec;
+use cgmath::{
+    BaseFloat, Decomposed, InnerSpace, Matrix4, Perspective, Point3, Quaternion, Transform, Vector3,
+};
 use collision::{Aabb3, Relation};
 use serde::{Deserialize, Serialize};
 
@@ -58,11 +62,11 @@ where
     }
 }
 
-impl<S> Cuboid<S> for Frustum<S>
+impl<S> ConvexPolyhedron<S> for Frustum<S>
 where
     S: BaseFloat,
 {
-    fn corners(&self) -> [Point3<S>; 8] {
+    fn compute_corners(&self) -> [Point3<S>; 8] {
         let corner_from = |x, y, z| self.query_from_clip.transform_point(Point3::new(x, y, z));
         [
             corner_from(-S::one(), -S::one(), -S::one()),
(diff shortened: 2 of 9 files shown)


nnmm:
ptal @feuerste

src/geometry/cube.rs
```diff
@@ -1,7 +1,51 @@
...
+
+impl<S> ConvexPolyhedron<S> for Aabb3<S>
+where
+    S: BaseFloat,
+{
+    fn compute_corners(&self) -> [Point3<S>; 8] {
+        self.to_corners()
+    }
+
+    fn compute_edges(&self) -> ArrayVec<[Vector3<S>; 6]> {
```
feuerste: I'm not sure about a better name, but my first understanding when reading the function name was to get the 12 edges of the box... how about something like `compute_unique_edge_vectors`?
Actually, thinking about the whole concept, can't we just force the user to implement `compute_corners` and determine the unique edges and face normals internally?

nnmm: We don't know which points are connected by an edge, so we might get more edges than exist. You'll also get quite many face normals this way.

nnmm: Would you be fine with naming it `compute_unique_edges()`?

src/geometry/obb.rs
```diff
@@ -34,95 +55,59 @@ impl<S: BaseFloat> Obb<S> {
...
+            corner_from(-self.half_extent.x, self.half_extent.y, -self.half_extent.z),
+            corner_from(self.half_extent.x, self.half_extent.y, -self.half_extent.z),
+            corner_from(-self.half_extent.x, -self.half_extent.y, self.half_extent.z),
+            corner_from(self.half_extent.x, -self.half_extent.y, self.half_extent.z),
+            corner_from(-self.half_extent.x, self.half_extent.y, self.half_extent.z),
+            corner_from(self.half_extent.x, self.half_extent.y, self.half_extent.z),
         ]
     }
+    fn compute_edges(&self) -> ArrayVec<[Vector3<S>; 6]> {
+        let mut edges = ArrayVec::new();
```
feuerste: Nit: Change to `ArrayVec::from(...)`?

src/geometry/obb.rs
```diff
@@ -1,18 +1,39 @@
...
 #[derive(Debug, Clone, Serialize, Deserialize)]
 pub struct Obb<S> {
     query_from_obb: Isometry3<S>,
     obb_from_query: Isometry3<S>,
     half_extent: Vector3<S>,
-    corners: [Point3<S>; 8],
-    pub separating_axes: Vec<Vector3<S>>,
+}
+
+/// Will be removed in the next PR.
```
feuerste: Nit: Replace with a TODO

src/geometry/cube.rs
```diff
@@ -1,7 +1,51 @@
...
+        let mut edges = ArrayVec::new();
+        edges.push(Vector3::unit_x());
+        edges.push(Vector3::unit_y());
+        edges.push(Vector3::unit_z());
+        edges
```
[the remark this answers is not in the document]
nnmm: We have three elements, but the backing array is 6 elements long. To use `from()`, we'd have to give it 6 elements. We can use `collect()` if you prefer.

src/geometry/frustum.rs
```diff
@@ -75,4 +79,46 @@ where
             corner_from(S::one(), S::one(), S::one()),
         ]
     }
+
+    fn compute_edges(&self) -> ArrayVec<[Vector3<S>; 6]> {
+        // To compute the edges, we need the points, so it's more efficient to implement
+        // intersector() directly and compute the points only once. We still provide this
```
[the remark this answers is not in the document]
nnmm: So, the dependency chain is corners -> edges -> normals. The default implementation would call:
* compute_corners(): Computes corners
* compute_edges(): Computes corners and edges
* compute_face_normals(): Computes corners, edges and face normals
so we'd compute corners three times and edges twice.

We could pass the corners to `compute_edges()`, and the edges to `compute_face_normals()`, would that be preferable? That forces an ordering between the calls, even when the previous results are not needed (like for AABBs).

src/geometry/cube.rs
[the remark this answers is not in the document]
feuerste: Ok, I see. No, then I would just leave it as it is...

src/geometry/cube.rs
feuerste: Yes

src/geometry/frustum.rs
```diff
@@ -75,4 +79,46 @@ where
...
+            (corners[5] - corners[4]).normalize(), // z lower right
+            (corners[7] - corners[6]).normalize(), // z upper right
+        ]);
+
+        let mut face_normals = ArrayVec::new();
+        face_normals.push((edges[0].cross(edges[1])).normalize()); // Front and back sides
+        face_normals.push((edges[0].cross(edges[2])).normalize()); // Lower side
+        face_normals.push((edges[0].cross(edges[3])).normalize()); // Upper side
+        face_normals.push((edges[1].cross(edges[2])).normalize()); // Left side
+        face_normals.push((edges[1].cross(edges[2])).normalize()); // right side
```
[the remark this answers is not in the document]
nnmm: Thanks, good catch!

src/geometry/frustum.rs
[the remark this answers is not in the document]
feuerste: This may be a solution. Or, alternatively, how about getting rid of `compute_edges()` and `compute_face_normals()` as well as the default implementation of `intersector()` and let just everyone implementing `ConvexPolyhedron` do the edge/face normal computation inside `intersector()`?

src/geometry/obb.rs
```diff
@@ -34,95 +55,59 @@ impl<S: BaseFloat> Obb<S> {
...
         [
-            corner_from(-half_extent.x, -half_extent.y, -half_extent.z),
-            corner_from(half_extent.x, -half_extent.y, -half_extent.z),
-            corner_from(-half_extent.x, half_extent.y, -half_extent.z),
-            corner_from(half_extent.x, half_extent.y, -half_extent.z),
-            corner_from(-half_extent.x, -half_extent.y, half_extent.z),
-            corner_from(half_extent.x, -half_extent.y, half_extent.z),
-            corner_from(-half_extent.x, half_extent.y, half_extent.z),
-            corner_from(half_extent.x, half_extent.y, half_extent.z),
+            corner_from(
```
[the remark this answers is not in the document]
nnmm: cargo fmt always splits this up, even with a single-letter name.

src/geometry/obb.rs
nnmm: See above

src/geometry/frustum.rs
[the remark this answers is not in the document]
nnmm: Sounds good.

src/geometry/obb.rs
nnmm: Done.

src/geometry/frustum.rs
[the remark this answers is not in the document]

src/math/sat.rs
```diff
@@ -1,36 +1,252 @@
...
+    let mut rel = Relation::In;
+    for sep_axis in separating_axes {
+        // Project corners of A onto that axis
+        let mut a_min_proj: S = Bounded::max_value();
+        let mut a_max_proj: S = Bounded::min_value();
+        for corner in corners_a {
+            let corner_proj = corner.to_vec().dot(sep_axis);
+            a_min_proj = a_min_proj.min(corner_proj);
+            a_max_proj = a_max_proj.max(corner_proj);
+        }
```
feuerste: How about moving this into a helper function or closure taking the corners and separation axis, so it can be reused below?

nnmm: Ok

src/geometry/obb.rs
```diff
@@ -34,95 +55,66 @@ impl<S: BaseFloat> Obb<S> {
...
-                && separating_axes.iter().all(|elem| {
-                    (elem - axis).magnitude() > S::default_epsilon()
-                        && (elem + axis).magnitude() > S::default_epsilon()
-                });
-            if is_finite_and_non_parallel {
-                separating_axes.push(*axis);
-            }
+    fn intersector(&self) -> Intersector<S> {
+        let corners = self.compute_corners();
+        let edges = self.compute_edges();
```
feuerste: Nit: It's a matter of taste, but I would just get rid of `compute_edges`, as it is only used once...

src/math/sat.rs
nnmm: Actually, I don't think that buys us readability, one updates a_min/max_proj, the other b_min/max_proj

nnmm: Ah, I see, they could be initialized inside and returned

nnmm: Done.

src/geometry/obb.rs
nnmm: Done.

src/geometry/obb.rs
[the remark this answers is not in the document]
feuerste: Ok, that's unfortunate.

src/geometry/cube.rs
```diff
@@ -1,7 +1,52 @@
...
+
+impl<S> PointCulling<S> for Aabb3<S>
+where
+    S: 'static + BaseFloat + Sync + Send + Bounded,
+{
+    fn contains(&self, p: &Point3<S>) -> bool {
+        Contains::contains(self, p)
+    }
+
+    fn intersects_aabb3(&self, aabb: &Aabb3<S>) -> bool {
```
feuerste: Would it make sense to return `Relation` here?

src/geometry/frustum.rs
```diff
@@ -75,4 +79,30 @@ where
...
+            (corners[5] - corners[4]).normalize(), // z lower right
+            (corners[7] - corners[6]).normalize(), // z upper right
+        ]);
+
+        let mut face_normals = ArrayVec::new();
+        face_normals.push((edges[0].cross(edges[1])).normalize()); // Front and back sides
+        face_normals.push((edges[0].cross(edges[2])).normalize()); // Lower side
+        face_normals.push((edges[0].cross(edges[3])).normalize()); // Upper side
+        face_normals.push((edges[1].cross(edges[2])).normalize()); // Left side
+        face_normals.push((edges[1].cross(edges[4])).normalize()); // right side
```
feuerste: Nit: right -> Right

src/geometry/cube.rs
nnmm: Yes, absolutely! That's planned for a future PR, I wanted to keep them small.

feuerste: Ok, great!

nnmm:
Thank you for reviewing!

PR #426, opened 2020-03-09
Remove transform() method on Frustum

nnmm:
It's an unused method that causes the Frustum to be more complex than it needs to be.
And also rename the `query` coordinate system to `world`.

### DIFF
--- a/src/geometry/frustum.rs
+++ b/src/geometry/frustum.rs
@@ -12,29 +12,21 @@ use serde::{Deserialize, Serialize};
 /// creating the perspective projection, see also the frustum unit test below.
 #[derive(Debug, Clone, Serialize, Deserialize)]
 pub struct Frustum<S: BaseFloat> {
-    query_from_eye: Isometry3<S>,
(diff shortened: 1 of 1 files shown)


nnmm:
ptal @feuerste @catevita

src/geometry/frustum.rs
```diff
@@ -12,30 +12,22 @@ use serde::{Deserialize, Serialize};
 /// creating the perspective projection, see also the frustum unit test below.
 #[derive(Debug, Clone, Serialize, Deserialize)]
 pub struct Frustum<S: BaseFloat> {
-    query_from_eye: Isometry3<S>,
-    clip_from_eye: Perspective<S>,
-    query_from_clip: Matrix4<S>,
+    world_from_clip: Matrix4<S>,
```
feuerste: I'm not 100% sure, whether the renaming here makes it clearer, in particular in situations where the Frustum is not defined in world coordinates, but some local coordinates...

nnmm: Ok, I'll undo it

PR #429, opened 2020-03-09
Use criterion for benchmarks

nnmm:
With bencher, it was hard to get meaningful results:
* I couldn't find a way to increase the number of times a function is iterated, to get narrower error bars
* But that wouldn't help anyway because bencher reports the difference between the min and max as the error bar, so the error bar would only grow.

Criterion can be used like bencher (`cargo bench`) but has much nicer reporting. Sample output:

> Benchmarking bench_s2_building_singlethreaded: Warming up for 3.0000 s
> Warning: Unable to complete 100 samples in 5.0s. You may wish to increase target time to 828.6s or reduce sample count to 10.
> bench_s2_building_singlethreaded
>                         time:   [148.67 ms 149.33 ms 150.27 ms]
>                         change: [-6.5982% -3.3366% +0.3522%] (p = 0.06 > 0.05)
>                         No change in performance detected.
> Found 16 outliers among 100 measurements (16.00%)
>   14 (14.00%) high mild
>   2 (2.00%) high severe

As you can see, if you've run it before, it will compare the durations. It also does warmup runs, which helps decrease variance from disk caches.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -461,12 +461,6 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "b41b7ea54a0c9d92199de89e20e58d49f02f8e699814ef3fdf266f6f748d15c7"
 
 [[package]]
-name = "bencher"
-version = "0.1.5"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "7dfdb4953a096c551ce9ace855a604d702e6e62d77fac690575ae347571717f5"
-
-[[package]]
 name = "bigdecimal"
 version = "0.1.0"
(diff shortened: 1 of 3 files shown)


nnmm:
ptal @feuerste

feuerste:
... and of course formatting needs to be fixed first :)

nnmm:
True, it's heavy, but at least it's a dev-dependency. So if you do `cargo build` in any of the crates including `point_cloud_test`, it won't get built, only when doing `cargo test/bench` in that crate (or all crates via `--workspace`). Do you agree it's worth it?

feuerste:
It's fine with me.

PR #430, opened 2020-03-11
Simplify bucket insertion for standard X-rays.

feuerste:


xray/src/build_quadtree.rs
```diff
@@ -142,7 +142,7 @@ pub fn run<T: Extension>(data_provider_factory: DataProviderFactory) {
             xray => ColoringStrategyKind::XRay,
             colored => ColoringStrategyKind::Colored(binning),
             colored_with_intensity => ColoringStrategyKind::ColoredWithIntensity(
-                value_t!(args, "min_intensity", f32).unwrap_or(1.),
+                value_t!(args, "min_intensity", f32).unwrap_or(0.),
                 value_t!(args, "max_intensity", f32).unwrap_or(1.),
```
krzesi-mir: Wouldn't it make more sense to panic if provided arguments cannot be parsed?

xray/src/generation.rs
```diff
@@ -180,16 +180,8 @@ impl ColoringStrategy for XRayColoringStrategy {
...
-                    e.get_mut().insert(d_loc.z);
-                }
-                Entry::Vacant(v) => {
-                    let mut set = FnvHashSet::default();
-                    set.insert(d_loc.z);
-                    v.insert(set);
-                }
-            }
+            let z_buckets = self.z_buckets.entry((d_loc.x, d_loc.y)).or_default();
+            z_buckets.insert(d_loc.z);
```
krzesi-mir: nit: use one-liner?
```suggestion
            self.z_buckets.entry((d_loc.x, d_loc.y)).or_default().insert(d_loc.z);
```

xray/src/build_quadtree.rs
krzesi-mir: WOuldn't it make more sense to panic when the provided arguments cannot be parsed?

xray/src/generation.rs
feuerste: Tried this before, cargo fmt doesn't make it a one-liner, but a three- or four-liner :(

xray/src/build_quadtree.rs
feuerste: Problably, this was just some random stuff I observed... how about just removing the `_or()` part  and make it an `unwrap()`?

feuerste: Hm, actually it makes most sense to provide default values above and change it here to `expect`s

krzesi-mir: :+1: that makes sense.

xray/src/generation.rs
krzesi-mir: :+1:

xray/src/build_quadtree.rs
feuerste: Done, PTAL!

feuerste:
💨 
Thanks for your review!

xray/src/build_quadtree.rs
krzesi-mir: lgtm

PR #431, opened 2020-03-11
Remove OctreeMeta's Default instance

nnmm:
This is to make https://github.com/googlecartographer/point_cloud_viewer/pull/415 smaller.

This makes the two dangerous/unexpected parts more explicit and doesn't allow them to hide in `..Default::default()`, namely:
* You shouldn't use a zero AABB as the bounding box for generation, but the Default instance gave you such an AABB.
* The two magic attributes "color" and "intensity".

The attribute data types are duplicated in two places now, but that's necessary for moving towards a state where octrees can have more attributes: The occurence in `build_octree()` will be removed, while the one in `from_data_provider()` will stay.

In the end, it's a matter of taste.

### DIFF
--- a/src/octree/generation.rs
+++ b/src/octree/generation.rs
@@ -60,7 +60,7 @@ impl RawNodeWriter {
     }
 }
 
-// Return a list a leaf nodes and a list of nodes to be splitted further.
+// Return a list of leaf nodes and a list of nodes to be split further.
 fn split<P>(
     octree_data_provider: &OnDiskDataProvider,
     octree_meta: &octree::OctreeMeta,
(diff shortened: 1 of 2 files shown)


nnmm:
meow @feuerste

feuerste:
Having the `Default` implementation actually served the purpose of getting rid of redundant code. How about providing a `new(resolution: f64, bounding_box: Aabb3<f64>)` method instead?

nnmm:
Would it be fine to make that HashMap a constant as well?

nnmm:
Ok, we don't have lazy_static as a dependency, I'll go with the new method.

feuerste:
Using lazy_static and a const hash map is fine with me, too, whatever you prefer...

PR #433, opened 2020-03-17
Expose quadtree level generation [AVMAPPING-8129]

krzesi-mir:
# What changed
* Created a new function `generation::build_level` which generates a new level based on nodes from level above;
* For simplification reasons removed `parents_to_create_{tx,rx}`, `all_nodes_{tx,rx}` crossbeam channels. The performance shouldn't be affected;
* Some minor refactorings.
# Tests
Produced xrays from both master and this branch, computed md5 checksums for each of the produced files and verified that they are consistent (except for `meta.pb`), and visualized the xrays.

### DIFF
--- a/xray/src/generation.rs
+++ b/xray/src/generation.rs
@@ -557,8 +557,7 @@ pub fn build_xray_quadtree(
     );
 
     // Create the deepest level of the quadtree.
-    //let (parents_to_create_tx, mut parents_to_create_rx) = crossbeam::channel::unbounded();
-    let (all_nodes_tx, all_nodes_rx) = crossbeam::channel::unbounded();
+    let mut all_nodes = FnvHashSet::default();
     let (created_leaf_node_ids_tx, created_leaf_node_ids_rx) = crossbeam::channel::unbounded();
 
     let mut leaf_nodes = Vec::with_capacity(4usize.pow(deepest_level.into()));
@@ -579,7 +578,6 @@ pub fn build_xray_quadtree(
     );
     rayon::scope(|scope| {
         while let Some(node) = leaf_nodes.pop() {
-            let all_nodes_tx_clone = all_nodes_tx.clone();
             let created_leaf_node_ids_tx_clone = created_leaf_node_ids_tx.clone();
             let strategy: Box<dyn ColoringStrategy> = coloring_strategy_kind.new_strategy();
             let progress_bar = Arc::clone(&progress_bar);
@@ -596,7 +594,6 @@ pub fn build_xray_quadtree(
                     strategy,
                     parameters,
                 ) {
-                    all_nodes_tx_clone.send(node.id).unwrap();
                     created_leaf_node_ids_tx_clone.send(node.id).unwrap();
                 }
                 progress_bar.lock().unwrap().inc();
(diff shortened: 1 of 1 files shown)


xray/src/generation.rs
```diff
@@ -635,55 +626,23 @@ pub fn build_xray_quadtree(
     });
     progress_bar.lock().unwrap().finish_println("");
 
+    let mut all_nodes = FnvHashSet::default();
+    let mut previous_level_nodes = created_leaf_node_ids;
+    for node in &previous_level_nodes {
+        all_nodes.insert(*node);
+    }
```
feuerste: How about shortening this to
```rust
let mut previous_level_nodes = created_leaf_node_ids;
let mut all_nodes = previous_level_nodes;
```
?

xray/src/generation.rs
```diff
@@ -635,55 +626,23 @@ pub fn build_xray_quadtree(
...
-                    parents_to_create_tx.send(id).unwrap()
-                }
-            }
-            progress_bar.lock().unwrap().inc();
-        });
-        progress_bar.lock().unwrap().finish_println("");
-        drop(parents_to_create_tx);
+        for node in &previous_level_nodes {
+            all_nodes.insert(*node);
+        }
```
feuerste: How about shortening this to
```rust
all_nodes.extend(&previous_level_nodes);
```
?

xray/src/generation.rs
```diff
@@ -715,3 +674,56 @@ pub fn build_xray_quadtree(
...
+    output_directory: &Path,
+    tile: &Tile,
+    current_level: u8,
+    previous_level_nodes: &FnvHashSet<NodeId>,
+    parameters: &XrayParameters,
+) -> FnvHashSet<NodeId> {
+    let current_level_nodes: FnvHashSet<NodeId> = previous_level_nodes
+        .iter()
+        .filter_map(|node| node.parent_id())
+        .collect();
```
feuerste: I wonder whether it would be easier to understand, if we pulled out the computation of `current_level_nodes` into the for loop and passed this instead of `previous_level_nodes`. Then the function wouldn't need to return anything.

xray/src/generation.rs
krzesi-mir: :+1:

xray/src/generation.rs
krzesi-mir: ah, thanks, this is exactly what i was looking for!

xray/src/generation.rs
krzesi-mir: I pulled it out and I think it looks better?

xray/src/generation.rs
```diff
@@ -635,55 +626,24 @@ pub fn build_xray_quadtree(
...
-                if let Some(id) = node_id.parent_id() {
-                    parents_to_create_tx.send(id).unwrap()
-                }
-            }
-            progress_bar.lock().unwrap().inc();
-        });
-        progress_bar.lock().unwrap().finish_println("");
-        drop(parents_to_create_tx);
+        all_nodes.extend(&current_level_nodes);
+        previous_level_nodes = current_level_nodes;
```
feuerste: Nit: How about getting rid of this line and `previous_level_nodes` and just use `current_level_nodes` everywhere?

xray/src/generation.rs
```diff
@@ -635,55 +626,23 @@ pub fn build_xray_quadtree(
...
-                );
-                image
-                    .as_rgba8()
-                    .unwrap()
-                    .save(&get_image_path(output_directory, node_id))
-                    .unwrap();
-                if let Some(id) = node_id.parent_id() {
-                    parents_to_create_tx.send(id).unwrap()
-                }
-            }
```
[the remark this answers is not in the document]
feuerste: Nice

xray/src/generation.rs
```diff
@@ -715,3 +674,56 @@ pub fn build_xray_quadtree(
...
+    // space must be at the bottom left. Since images have their origin at the top
+    // left, we need actually have to invert y and go from the bottom of the image.
+    for id in 0..4 {
+        let png = get_image_path(
+            output_directory,
+            node_id.get_child_id(&ChildIndex::from_u8(id)),
+        );
+        if png.exists() {
+            children[id as usize] = Some(image::open(&png).unwrap().to_rgba());
+        }
```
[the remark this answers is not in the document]
feuerste: True, I also like it more this way.

xray/src/generation.rs
krzesi-mir: Good idea, thanks.

PR #434, opened 2020-03-17
Add option to specify node id for X-ray quadtree generation

feuerste:
This enables the generation of sub-quadtrees, which later can be merged. This is a useful feature for sharding quadtree generation.

### DIFF
--- a/quadtree/src/lib.rs
+++ b/quadtree/src/lib.rs
@@ -55,13 +55,6 @@ pub struct Node {
 }
 
 impl Node {
-    pub fn root_with_bounding_rect(rect: Rect) -> Self {
-        Node {
-            id: NodeId::root(),
-            bounding_rect: rect,
-        }
-    }
-
     pub fn from_node_id_and_root_bounding_rect(id: NodeId, rect: Rect) -> Self {
         let mut ids = vec![id];
         while let Some(parent_id) = ids.last().and_then(|last| last.parent_id()) {
--- a/quadtree/src/lib.rs
+++ b/quadtree/src/lib.rs
@@ -62,6 +62,21 @@ impl Node {
         }
     }
 
+    pub fn from_node_id_and_root_bounding_rect(id: NodeId, rect: Rect) -> Self {
+        let mut ids = vec![id];
+        while let Some(parent_id) = ids.last().and_then(|last| last.parent_id()) {
+            ids.push(parent_id);
+        }
+        let mut node = Self {
+            id: ids.pop().unwrap(),
+            bounding_rect: rect,
+        };
+        while let Some(child_id) = ids.pop().and_then(|id| id.child_index()) {
+            node = node.get_child(&child_id);
+        }
+        node
+    }
+
     pub fn get_child(&self, child_index: &ChildIndex) -> Node {
         let child_bounding_rect = {
             let half_edge_length = self.bounding_rect.edge_length() / 2.;
--- a/quadtree/src/lib.rs
+++ b/quadtree/src/lib.rs
@@ -55,15 +55,21 @@ pub struct Node {
 }
 
 impl Node {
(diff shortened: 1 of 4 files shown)


quadtree/src/lib.rs
```diff
@@ -55,11 +55,19 @@ pub struct Node {
 }
 
 impl Node {
-    pub fn root_with_bounding_rect(rect: Rect) -> Self {
-        Node {
-            id: NodeId::root(),
+    pub fn from_node_id_and_root_bounding_rect(id: NodeId, rect: Rect) -> Self {
+        let mut ids = vec![id];
```
krzesi-mir: Maybe it would make sense to store only `child_indices` here instead of the whole `node_ids`? And store only the ids/child_indices of non-root nodes.

quadtree/src/lib.rs
```diff
@@ -55,11 +55,19 @@ pub struct Node {
...
+        while let Some(parent_id) = ids.last().and_then(|last| last.parent_id()) {
+            ids.push(parent_id);
+        }
+        let mut node = Self {
+            id: ids.pop().unwrap(),
```
krzesi-mir: Just making sure I understand: at this point `ids.pop().unwrap()` returns the root node, or`id: NodeId::root()`?

quadtree/src/lib.rs
```diff
@@ -55,11 +55,19 @@ pub struct Node {
...
             bounding_rect: rect,
+        };
+        while let Some(child_id) = ids.pop().and_then(|id| id.child_index()) {
```
krzesi-mir: nit: `child_id -> child_index`?

krzesi-mir: (assuming that root is not in `ids` so `child_id.child_index().unwrap()` should be safe) We could use iterators here:
`ids.iter().rev().fold(node, |node, child_id| node.get_child(&child_id.child_index().unwrap()))`
not sure if it makes it cleaner though.

xray/src/generation.rs
```diff
@@ -708,7 +716,8 @@ pub fn build_xray_quadtree(
         meta
     };
 
-    let mut buf_writer = BufWriter::new(File::create(output_directory.join("meta.pb")).unwrap());
+    let meta_pb_name = format!("{}.pb", root_node_id).replace("r", "meta");
```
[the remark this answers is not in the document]
feuerste: Yes.

xray/src/lib.rs
```diff
@@ -111,7 +111,10 @@ impl Meta {
         let frustum =
             Frustum::from_matrix4(matrix).ok_or("Unable to create frustum from matrix")?;
         let mut result = Vec::new();
-        let mut open = vec![Node::root_with_bounding_rect(self.bounding_rect.clone())];
+        let mut open = vec![Node::from_node_id_and_root_bounding_rect(
+            NodeId::root(),
+            self.bounding_rect.clone(),
+        )];
         while !open.is_empty() {
             let node = open.pop().unwrap();
```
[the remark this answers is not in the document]
feuerste: Done

quadtree/src/lib.rs
feuerste: Done

quadtree/src/lib.rs
feuerste: Yes, I implemented it this way that it also works when just passing the actual root node.

quadtree/src/lib.rs
feuerste: Hm, I'm not sure whether this works. Currently, if the sub-root node is somewhere in the middle of the tree, we first need to find all nodes on the way to the root node. After finding these, we can use the `get_child()` method starting with the root node down to the sub-root node to compute the adjusted bounding box internally...

PR #435, opened 2020-03-18
xray merger interface 

krzesi-mir:


xray/src/bin/merge_xray_quadtrees.rs
```diff
@@ -0,0 +1,22 @@
...
+/// of each octree belongs to the same level of the final
+/// octree. 
+struct CommandlineArguments {
+    /// Directory with partial xray quadtrees.
+    #[structopt(parse(from_os_str))]
+    directory: PathBuf,
+    /*
+    /// Paths to each quadtree's metadata protobuf file.
+    metadata_proto_filenames: Vector<Pathbuf>
+    */
```
krzesi-mir: What do you think: should this binary scan the directory and look for metadata protos on its own, or should the user pass  as an argument the paths to those protos?

feuerste: I think passing the directories is enough, with something like https://docs.rs/globwalk/0.8.0/globwalk/ we can easily identify the *.pb files. Going a step further, once we have dozens or hundreds of partial quadtrees, we may instead pass one single file containing all the paths...

xray/src/bin/merge_xray_quadtrees.rs
```diff
@@ -0,0 +1,22 @@
+use std::path::{PathBuf};
+use structopt::StructOpt;
+
+#[derive(StructOpt, Debug)]
+#[structopt(name = "merge_xray_quadtrees")]
+/// Merge partial xray octrees. We assume that the root
+/// of each octree belongs to the same level of the final
+/// octree.
```
feuerste: octree(s) -> quadtree(s)

krzesi-mir: oups, thanks.

krzesi-mir:
@googlebot I fixed it.

PR #436, opened 2020-03-18
xray merger interface

krzesi-mir:
```
$ cargo run --bin merge_xray_quadtrees -- --output_directory /mnt/disk . ../ a/ a/b/c/
[xray/src/bin/merge_xray_quadtrees.rs:21] args = CommandlineArguments {
    output_directory: "/mnt/disk",
    input_directories: [
        ".",
        "../",
        "a/",
        "a/b/c/",
    ],
}
thread 'main' panicked at 'not implemented', xray/src/bin/merge_xray_quadtrees.rs:22:5
```

```
$ cargo run --bin merge_xray_quadtrees -- --help
merge_xray_quadtrees 0.1.0
Merge partial xray quadtrees. We assume that the root of each quadtree belongs to the same level of the final quadtree

USAGE:
    merge_xray_quadtrees --output_directory <output-directory> [input-directories]...

FLAGS:
    -h, --help       Prints help information
    -V, --version    Prints version information

OPTIONS:
        --output_directory <output-directory>
            Directory where to write the merged quadtree. Does *not* have to be disjoint from input_directories


ARGS:
    <input-directories>...    Directories with, possibly multiple, partial xray quadtrees
```

### DIFF
--- a/xray/src/bin/merge_xray_quadtrees.rs
+++ b/xray/src/bin/merge_xray_quadtrees.rs
@@ -7,12 +7,17 @@ use structopt::StructOpt;
 /// of each quadtree belongs to the same level of the final
 /// quadtree.
 struct CommandlineArguments {
-    /// Directory with partial xray quadtrees.
+    /// Directory where to write the merged quadtree. Does *not*
+    /// have to be disjoint from input_directories.
+    #[structopt(parse(from_os_str), long = "output_directory")]
+    output_directory: PathBuf,
+    /// Directories with, possibly multiple, partial xray quadtrees.
     #[structopt(parse(from_os_str))]
-    directory: PathBuf,
+    input_directories: Vec<PathBuf>,
 }
 
 fn main() {
-    let _args = CommandlineArguments::from_args();
+    let args = CommandlineArguments::from_args();
+    dbg!(args);
     unimplemented!();
 }
--- a/xray/src/bin/merge_xray_quadtrees.rs
+++ b/xray/src/bin/merge_xray_quadtrees.rs
@@ -1,5 +1,6 @@
 use std::path::PathBuf;
 use structopt::StructOpt;
+use xray::generation;
 
 #[derive(StructOpt, Debug)]
 #[structopt(name = "merge_xray_quadtrees")]
@@ -11,6 +12,9 @@ struct CommandlineArguments {
     /// have to be disjoint from input_directories.
     #[structopt(parse(from_os_str), long = "output_directory")]
     output_directory: PathBuf,
+    #[structopt(parse(from_os_str), default_value = "white")]
+    /// ...
+    tile_background_color: generation::TileBackgroundColorArgument,
     /// Directories with, possibly multiple, partial xray quadtrees.
     #[structopt(parse(from_os_str))]
     input_directories: Vec<PathBuf>,
--- a/xray/src/bin/merge_xray_quadtrees.rs
+++ b/xray/src/bin/merge_xray_quadtrees.rs
@@ -1,0 +1,18 @@
+use std::path::PathBuf;
+use structopt::StructOpt;
+
+#[derive(StructOpt, Debug)]
+#[structopt(name = "merge_xray_quadtrees")]
+/// Merge partial xray octrees. We assume that the root
+/// of each quadtree belongs to the same level of the final
+/// quadtree.
+struct CommandlineArguments {
+    /// Directory with partial xray quadtrees.
+    #[structopt(parse(from_os_str))]
+    directory: PathBuf,
+}
+
+fn main() {
+    let _args = CommandlineArguments::from_args();
+    unimplemented!();
+}
--- a/xray/src/bin/merge_xray_quadtrees.rs
+++ b/xray/src/bin/merge_xray_quadtrees.rs
@@ -3,7 +3,7 @@ use structopt::StructOpt;
 
 #[derive(StructOpt, Debug)]
 #[structopt(name = "merge_xray_quadtrees")]
-/// Merge partial xray octrees. We assume that the root
+/// Merge partial xray quadtrees. We assume that the root
 /// of each quadtree belongs to the same level of the final
 /// quadtree.
 struct CommandlineArguments {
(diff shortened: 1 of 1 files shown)


xray/src/bin/merge_xray_quadtrees.rs
```diff
@@ -0,0 +1,18 @@
+use std::path::PathBuf;
+use structopt::StructOpt;
+
+#[derive(StructOpt, Debug)]
+#[structopt(name = "merge_xray_quadtrees")]
+/// Merge partial xray octrees. We assume that the root
```
feuerste: octrees -> quadtrees

xray/src/bin/merge_xray_quadtrees.rs
```diff
@@ -0,0 +1,18 @@
...
+/// of each quadtree belongs to the same level of the final
+/// quadtree.
+struct CommandlineArguments {
+    /// Directory with partial xray quadtrees.
+    #[structopt(parse(from_os_str))]
+    directory: PathBuf,
```
feuerste: How about a Vec<PathBuf>, so we could potentially have inputs in several places?
And how about also specifying one dedicated output directory (which can be equal to one of the input directories)?

xray/src/bin/merge_xray_quadtrees.rs
krzesi-mir: oups...

xray/src/bin/merge_xray_quadtrees.rs
krzesi-mir: Agree with `Vec<PathBuf>` suggestion.
About output directory: what would be output there? Only the newly generated nodes/images or the whole quadtree?

xray/src/bin/merge_xray_quadtrees.rs
```diff
@@ -0,0 +1,23 @@
...
+#[derive(StructOpt, Debug)]
+#[structopt(name = "merge_xray_quadtrees")]
+/// Merge partial xray quadtrees. We assume that the root
+/// of each quadtree belongs to the same level of the final
+/// quadtree.
+struct CommandlineArguments {
+    /// Directory where to write the merged quadtree. Does *not*
+    /// have to be disjoint from input_directories.
+    #[structopt(parse(from_os_str), long = "output_directory")]
+    output_directory: PathBuf,
```
feuerste: You can remove the `long`, as default in our Rust code is kebap case and structopt automatically converts to it.

xray/src/bin/merge_xray_quadtrees.rs
```diff
@@ -0,0 +1,23 @@
...
+    /// Directories with, possibly multiple, partial xray quadtrees.
+    #[structopt(parse(from_os_str))]
```
feuerste: We may explicitly add a `long = "input-directory"` here, as this argument can be specified multiple times and then IMHO it reads nicer to have
```
merge_xray_quadtree --input-directory bla --input-directory blub ...
```

xray/src/bin/merge_xray_quadtrees.rs
feuerste: I think it should be the whole quadtree

xray/src/bin/merge_xray_quadtrees.rs
krzesi-mir: Do you find `merge_xray_quadtree --output-directory OUT --input-directory IN1 --input-directory IN2` nicer to read than `merge_xray_quadtree --output-directory OUT IN1 IN2`>

xray/src/bin/merge_xray_quadtrees.rs
krzesi-mir: Do you mean:
`#[structopt(parse(from_os_str), long)]`
or
`#[structopt(parse(from_os_str))]`?
The second option would make the argument positional.

krzesi-mir:
With new arguments:
```
$ cargo run --bin merge_xray_quadtrees -- --help
merge_xray_quadtrees 0.1.0
Merge partial xray quadtrees. We assume that the root of each quadtree belongs to the same level of the final quadtree

USAGE:
    merge_xray_quadtrees [OPTIONS] --output-directory <output-directory> [input-directories]...

FLAGS:
    -h, --help       Prints help information
    -V, --version    Prints version information

OPTIONS:
        --output-directory <output-directory>
            Directory where to write the merged quadtree. Does *not* have to be disjoint from input_directories

        --tile-background-color <tile-background-color>    Tile background color [default: white]
        --tile-size <tile-size>                            Tile size [default: 256]

ARGS:
    <input-directories>...    Directories with, possibly multiple, partial xray quadtrees
```

krzesi-mir:
Or do you prefer to have both `output_directory` and `input_directories` as positional arguments?
i.e.

krzesi-mir:
Or
```
$ cargo run --bin merge_xray_quadtrees -- --help
merge_xray_quadtrees 0.1.0
Merge partial xray quadtrees. We assume that the root of each quadtree belongs to the same level of the final quadtree

USAGE:
    merge_xray_quadtrees [OPTIONS] --output-directory <output-directory>

FLAGS:
    -h, --help       Prints help information
    -V, --version    Prints version information

OPTIONS:
        --input-directory <input-directories>...           Directories with, possibly multiple, partial xray quadtrees
        --output-directory <output-directory>
            Directory where to write the merged quadtree. Does *not* have to be disjoint from input_directories

        --tile-background-color <tile-background-color>    Tile background color [default: white]
        --tile-size <tile-size>                            Tile size [default: 256]
```

xray/src/bin/merge_xray_quadtrees.rs
feuerste: Actually your second version looks nice, too, so we can use this as well

xray/src/bin/merge_xray_quadtrees.rs
feuerste: Oh yes, sorry for the confusion, I meant `long`

krzesi-mir: :+1:

xray/src/bin/merge_xray_quadtrees.rs
krzesi-mir: :+1:
I'll go with the second version then :)

krzesi-mir:
Thanks for the review!

PR #440, opened 2020-03-19
a binary for merging subquadtrees into a complete quadtree.

krzesi-mir:


xray/src/lib.rs
```diff
@@ -11,6 +11,7 @@ use std::path::Path;
 // Version 2 -> 3: Change in Rect proto from Vector2f to Vector2d min and float to double edge_length.
 // We are able to convert the proto on read, so the tools can still read version 2.
 pub const CURRENT_VERSION: i32 = 3;
+pub const IMAGE_FILE_EXTENSION: &str = "png";
```
krzesi-mir: do you think it makes sense to have this constant? Or should we just hardcode "png" in several places?

xray/src/bin/merge_xray_quadtrees.rs
```diff
@@ -20,8 +27,179 @@ struct CommandlineArguments {
...
+    if input_directory == output_directory {
+        return;
+    }
+    globwalk::GlobWalkerBuilder::new(
+        input_directory,
+        &format!("*.{}", xray::IMAGE_FILE_EXTENSION),
+    )
+    .build()
+    .expect("Failed to build GlobWalker")
+    .filter_map(Result::ok)
```
krzesi-mir: I'm filtering out errors here (i'm not sure when we can get errors, tbh, i coulnd find it in the documentation of glob_walker). Do you think that instead of filtering we should panic on error?

xray/src/bin/merge_xray_quadtrees.rs
```diff
@@ -20,8 +27,179 @@ struct CommandlineArguments {
...
+    let level = check_all_the_same(root_nodes.iter().map(|node| node.level()))
+        .expect("Not all roots have the same level.");
+    let version = check_all_the_same(metadata.iter().map(|meta| meta.get_version()))
+        .expect("Not all meta files have the same version.");
+    assert_eq!(version, xray::CURRENT_VERSION);
+    let deepest_level = check_all_the_same(metadata.iter().map(|meta| meta.get_deepest_level()))
+        .expect("Not all meta files have the same deepest level.") as u8;
+    let tile_size = check_all_the_same(metadata.iter().map(|meta| meta.get_tile_size()))
+        .expect("Not all meta files have the same tile size.");
+    // TODO: check whether all root nodes have the same bound_rect.
```
feuerste: I think it would definitely make sense to implement this

xray/src/bin/merge_xray_quadtrees.rs
```diff
@@ -20,8 +27,179 @@ struct CommandlineArguments {
...
+        .iter()
+        .map(NodeId::from)
+        .min_by_key(|node| node.level())
+}
+
+fn get_root_nodes(meta: &Vec<proto::Meta>) -> Vec<Option<NodeId>> {
+    meta.iter().map(get_root_node).collect()
+}
+
+struct Metadata {
```
feuerste: How about reusing/extending `Meta` from `xray/src/lib` here?

xray/src/bin/merge_xray_quadtrees.rs
feuerste: No, I think it's ok as it is.

xray/src/bin/merge_xray_quadtrees.rs
```diff
@@ -20,8 +27,179 @@ struct CommandlineArguments {
...
+    )
+    .build()
+    .expect("Failed to build GlobWalker")
+    .filter_map(Result::ok)
+    .for_each(|dir_entry| {
+        std::fs::copy(
+            dir_entry.path(),
+            output_directory.join(dir_entry.file_name()),
+        )
+        .expect("Failed to copy the file.");
```
feuerste: Nit: State which file failed

xray/src/bin/merge_xray_quadtrees.rs
```diff
@@ -20,8 +27,179 @@ struct CommandlineArguments {
...
+}
+
+fn get_root_node(meta: &proto::Meta) -> Option<NodeId> {
+    meta.get_nodes()
+        .iter()
+        .map(NodeId::from)
+        .min_by_key(|node| node.level())
+}
+
+fn get_root_nodes(meta: &Vec<proto::Meta>) -> Vec<Option<NodeId>> {
```
feuerste: Would it make sense to already return an `Option<Vec<NodeId>>` here? This can easily be done by just changing the return type, `collect()` will do the magic behind...
Then the logic inside `validate_metadata()` would become a little easier...

xray/src/bin/merge_xray_quadtrees.rs
```diff
@@ -20,8 +27,179 @@ struct CommandlineArguments {
...
+    nodes: FnvHashSet<NodeId>,
+    level: u8,
+    deepest_level: u8,
+    tile_size: u32,
+    bounding_rect: proto::Rect,
+}
+
+// Checks wheter all the elements in the iterator have the same value and returns it.
+// There must be at least one element in the iterator.
+fn check_all_the_same<I, V>(mut iterator: I) -> Option<V>
```
feuerste: How about using https://docs.rs/itertools/0.9.0/itertools/trait.Itertools.html#method.all_equal instead? We already use `itertools` anyway.

xray/src/bin/merge_xray_quadtrees.rs
```diff
@@ -20,8 +27,179 @@ struct CommandlineArguments {
...
+        root_nodes,
+        level,
+        deepest_level,
+        tile_size,
+        bounding_rect,
+        nodes,
+    }
+}
+
+fn write_metadata(metadata: Metadata, output_directory: &Path) {
```
feuerste: How about having a `From` implementation for `Meta` instead? Then we could also reuse it for replacing the meta generation code at the end of `build_xray_quadtree` in `xray/src/generation.rs`.

xray/src/lib.rs
feuerste: I think a const makes sense.

xray/src/bin/merge_xray_quadtrees.rs
krzesi-mir: ah, didn't know it existed, thanks!

xray/src/bin/merge_xray_quadtrees.rs
krzesi-mir: I like this idea

xray/src/bin/merge_xray_quadtrees.rs
krzesi-mir: done

xray/src/bin/merge_xray_quadtrees.rs
krzesi-mir: Returning Vec<NodeId> while skipping metas with no nodes.

xray/src/bin/merge_xray_quadtrees.rs
krzesi-mir: Since there exists `Meta::from_proto`, for symmetry I added `Meta::to_proto`

xray/src/bin/merge_xray_quadtrees.rs
```diff
@@ -20,8 +24,154 @@ struct CommandlineArguments {
...
+    }
+}
+
+fn read_metadata_from_directory(directory: &Path) -> Vec<Meta> {
+    globwalk::GlobWalkerBuilder::new(directory, "meta*.pb")
+        .build()
+        .expect("Failed to build GlobWalker")
+        .filter_map(Result::ok)
+        .map(|dir_entry| PathBuf::from(dir_entry.path()))
+        .map(|path| Meta::from_disk(&path).expect(&format!("Failed to load meta from {:?}.", path)))
```
feuerste: Nit: These two lines can be merged, right?

xray/src/lib.rs
```diff
@@ -44,6 +47,18 @@ impl Meta {
         Ok(Self::from_proto(&proto))
     }
 
+    pub fn to_disk<P: AsRef<Path>>(&self, filename: P) -> io::Result<()> {
```
feuerste: Can we also use this method in `generation.rs`?

krzesi-mir: done

xray/src/bin/merge_xray_quadtrees.rs
```diff
@@ -20,8 +24,156 @@ struct CommandlineArguments {
...
+
+    let mut nodes = FnvHashSet::default();
+    for meta in metadata {
+        nodes.extend(&meta.nodes);
+    }
+
+    MergedMetadata {
+        root_nodes,
+        level,
+        merged_meta: Meta {
```
feuerste: Nit: I think we can remove the `merged_` prefix here or rename it to `root_meta`... What do you think?

xray/src/bin/merge_xray_quadtrees.rs
krzesi-mir: done

xray/src/bin/merge_xray_quadtrees.rs
krzesi-mir: :+1: sounds good

xray/Cargo.toml
```diff
@@ -10,14 +10,17 @@ authors = [
...
 cgmath = "0.17.0"
 clap = "2.33.0"
 collision = "0.20.1"
 crossbeam = "0.7.3"
 fnv = "1.0.6"
+globwalk = "0.8.0"
 image = "0.23.1"
 imageproc = "0.20.0"
 iron = "0.6.1"
+itertools = "0.9.0"
```
feuerste: Nit: remove again

xray/src/bin/merge_xray_quadtrees.rs
```diff
@@ -20,8 +24,170 @@ struct CommandlineArguments {
...
+    all_equal_by_func(iterator, |left, right| left == right)
+}
+
+fn all_equal_by_func<I, V, F>(mut iterator: I, comp: F) -> Option<V>
+where
+    I: Iterator<Item = V>,
+    F: Fn(&V, &V) -> bool,
+    V: Clone,
+{
+    let first_element = &iterator.next().expect("Iterator cannot be empty.");
```
feuerste: Would something like this work as well?
```rust
    iterator.next().and_then(|first_element| {
        if iterator.all(|element| comp(&element, first_element)) {
            Some(first_element)
        } else {
            None
        }
    })
```

krzesi-mir: :+1: done

xray/Cargo.toml
krzesi-mir: :+1:

xray/src/bin/merge_xray_quadtrees.rs
```diff
@@ -20,8 +24,171 @@ struct CommandlineArguments {
     input_directories: Vec<PathBuf>,
 }
 
+fn copy_images(input_directory: &Path, output_directory: &Path) {
+    if input_directory == output_directory {
```
feuerste: I think this equality only makes sense after `canonicalize()`ing the paths. Otherwise, relative and absolute paths of the same dir seem to be different.

krzesi-mir: :+1: fixed

krzesi-mir:
> LGTM
> The only remaining thing is that we have lots of `expect(...)`s and `unwrap()`s... we may consider removing them... But this could be in a separate PR, too.

Yeah, I agree with, you. Will do it in another PR.

krzesi-mir:
Thanks for the review!

PR #443, opened 2020-03-25
Set bounding rects to actual (sub) root node.

feuerste:


xray/src/bin/merge_xray_quadtrees.rs
```diff
@@ -77,12 +76,18 @@ fn read_metadata_from_directories(directories: &[PathBuf]) -> Vec<Meta> {
         .collect()
 }
 
-fn get_root_node(meta: &Meta) -> Option<NodeId> {
-    meta.nodes.iter().copied().min_by_key(|node| node.level())
+fn get_root_node(meta: &Meta) -> Option<Node> {
```
krzesi-mir: Let's make this function a public method of`Meta`?

feuerste: Done

krzesi-mir: :+1:

feuerste:
Thanks for your review!

PR #444, opened 2020-03-31
Export some of the functionality in `xray::generation`

krzesi-mir:


quadtree/src/lib.rs
```diff
@@ -217,9 +217,10 @@ impl FromStr for NodeId {
     }
 }
 
+pub const NODE_PREFIX: char = 'r';
 impl fmt::Display for NodeId {
     fn fmt(&self, formatter: &mut fmt::Formatter) -> fmt::Result {
-        formatter.write_char('r')?;
+        formatter.write_char(NODE_PREFIX)?;
```
feuerste: What do you think about changing this to
```rust
pub const NODE_PREFIX: &str = "r";
...
        formatter.write_str(NODE_PREFIX)?;
```
then we wouldn't need `&NODE_PREFIX.to_string()` in `utils.rs`?

xray/src/utils.rs
```diff
@@ -1,9 +1,24 @@
...
 
+const META_PREFIX: &str = "meta";
+
 pub fn get_meta_pb_path(directory: &Path, id: NodeId) -> PathBuf {
-    directory.join(&format!("{}.pb", id).replace("r", "meta"))
+    directory.join(&format!("{}.pb", id).replace(&NODE_PREFIX.to_string(), META_PREFIX))
+}
+
+pub fn get_root_node_id_from_meta_pb_path(meta_path: &Path) -> io::Result<NodeId> {
+    let error_message = format!("Invalid path {:?}.", meta_path);
```
feuerste: How about changing this into
```rust
let invalid_input_error = || {
    io::Error::new(io::ErrorKind::InvalidInput, format!("Invalid path {:?}.", meta_path))
};
```
and calling `invalid_input_error()` below twice?

quadtree/src/lib.rs
krzesi-mir: done :+1:

xray/src/utils.rs
krzesi-mir: Good idea! changed.

krzesi-mir:
Thanks for the review, Marco!

PR #445, opened 2020-04-02
Redirect progress and logs to stderr.

feuerste:
This is a first step towards proper logging. On `stdout` we only want program outputs.


feuerste:
Thanks for your review Kamil!

PR #446, opened 2020-04-03
Fix errors in local xray viewer console

nnmm:
ptal @dudv 
I just took the advice of the first [StackOverflow answer](https://stackoverflow.com/questions/42101723/unable-to-preventdefault-inside-passive-event-listener) for the error message ("Unable to preventDefault inside passive event listener"), but I'm not sure what the effect is. Can you let me know whether this is the right thing to do?

### DIFF
--- a/xray/client/control.ts
+++ b/xray/client/control.ts
@@ -44,22 +44,22 @@ export class Maps2DController {
     window.addEventListener(
       'keydown',
       (event) => this.onKeyDown(<KeyboardEvent>event),
-      false
+      { capture: false, passive: false }
     );
     window.addEventListener(
       'keyup',
       (event) => this.onKeyUp(<KeyboardEvent>event),
-      false
+      { capture: false, passive: false }
     );
     window.addEventListener(
       'mousewheel',
       (event) => this.onMouseWheel(<WheelEvent>event),
-      false
+      { capture: false, passive: false }
     );
     this.domElement.addEventListener(
       'mousedown',
       (event) => this.onMouseDown(<MouseEvent>event),
-      false
+      { capture: false, passive: false }
     );
   }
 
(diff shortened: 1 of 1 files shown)


dudv:
Yes, this is ok to do

nnmm:
Actually, I went with another route. This fixes the actual problem that I was seeing: That tiles did not get loaded.

ptal @feuerste

xray/client/index.html
```diff
@@ -18,7 +18,7 @@
    <title>XRay Viewer</title>
 </head>
 <body>
-  <div id="renderArea" style="position: absolute; width: 100%; height: 100%; overflow: hidden;"></div>
+  <div id="renderArea" style="position: absolute; width: 100%; height: 100%; overflow: hidden; touch-action: none;"></div>
```
feuerste, line 21: So what was the actual problem?

nnmm: When I zoomed in, the tiles would stay white.

nnmm: I implemented the fix for the error message in search for a fix for the problem, but it did not fix the problem. I still created a PR, but searched further and this fixes the problem.

nnmm: This is Yak shaving depth 3 for me by the way :P

feuerste: Ok, do we also need this fix in https://github.com/googlecartographer/point_cloud_viewer/blob/master/octree_web_viewer/client/index.html#L22?

nnmm: Probably. I added it.

feuerste: Thanks a lot! Can you please also test it?

nnmm: I tried ... I can't see any octree with or without the fix. Can we merge just the xray fix?

feuerste: I just tried your fix, for me the error still occurs.

nnmm: Have you deleted the yarn-generated files and regenerated them?

feuerste: Yes

nnmm: You're right, I have to check again.

feuerste: Is your original approach (https://github.com/cartographer-project/point_cloud_viewer/pull/446/commits/71259c21ea3bb891de936200c1408fdb7e35ba1e) working?

feuerste: #466 fixes the actual problem of tiles staying white...

PR #447, opened 2020-04-06
Move S2 cell union code to src/geometry/s2_cell_union.rs

nnmm:
This is to make src/math.rs less of a grab-bag. I think having this here makes sense because an S2 cell union is one of the query types, like Aabb, Obb and Frustum, and it is a geometric construct.

### DIFF
--- a/src/geometry/s2_cell_union.rs
+++ b/src/geometry/s2_cell_union.rs
@@ -8,6 +8,32 @@ use crate::math::FromPoint3;
 use nalgebra::{Point3, RealField};
 use s2::{cell::Cell, cellid::CellID, region::Region};
 
+/// Checks for an intersection between a cell_union and a polyhedron.
+///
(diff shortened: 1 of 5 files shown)


nnmm:
ptal @catevita

src/geometry/s2_cell_union.rs
```diff
@@ -0,0 +1,33 @@
...
+impl<S> PointCulling<S> for CellUnion
+where
+    S: RealField,
+    f64: From<S>,
+{
+    fn contains(&self, p: &Point3<S>) -> bool {
+        self.contains_cellid(&CellID::from_point(p))
+    }
+
+    fn intersects_aabb(&self, aabb: &Aabb<S>) -> bool {
```
catevita, line 20: maybe here it is worth to mention that the intersection is carried on the rect bound of the cell union

PR #449, opened 2020-04-07
Web Mercator conversion from and to Lat/Lng

nnmm:
First step towards a Web Mercator (+ margin) query


nnmm:
ptal @catevita

nnmm:
ptal @cmarschner

src/math/web_mercator.rs
```diff
@@ -0,0 +1,146 @@
...
+
+impl WebMercatorCoord {
+    /// Projects a lat/lng coordinate to Web Mercator.
+    ///
+    /// Equivalent to the formula on [Wikipedia](https://en.wikipedia.org/wiki/Web_Mercator_projection#Formulas).
+    /// If the latitude is outside `[-85.051129, 85.051129]`, it is clamped to that interval first.
+    pub fn from_lat_lng(lat_lng: &WGS84<f64>) -> Self {
+        // Implemented according to
+        // https://developers.google.com/maps/documentation/javascript/examples/map-coordinates?csw=1
+        // but clamping is done before the sin() operation.
```
catevita, line 39: clamping after the sin would allow latitude a slightly bigger interval, until 89. degrees, why this choice?

catevita: i got it, the projection becomes a square

nnmm: Exactly :)

src/math/web_mercator.rs
```diff
@@ -0,0 +1,146 @@
...
+/// LAT_BOUND_SIN = sin(LAT_BOUND_RAD)
+const LAT_BOUND_SIN: f64 = 0.99627207622074998027983383508399;
+
+const TWO_PI: f64 = 2.0 * PI;
+const FOUR_PI: f64 = 4.0 * PI;
+const FOUR_PI_INV: f64 = 0.25 * FRAC_1_PI;
+const TILE_SIZE: u32 = 256;
+
+/// The max zoom level is currently 23 because it makes implementation easier,
+/// but theoretically nothing stops us from going deeper.
```
catevita, line 20: the current max zoom is given by the TILE_SIZE<<zoom which is expressed in u32 in line 81. I'd say it is currently limited by the implementation choice, or make this constant bound to the integer bits that describe TILE_SIZE.

src/math/web_mercator.rs
```diff
@@ -0,0 +1,146 @@
...
+    fn projection_ground_truth() {
+        let test_coordinate = WGS84::new(37.407204, -122.147604, 0.0);
+        // This test coordinate is at approx. pixel (165, 18) on this OSM tile at level 19:
+        // https://a.tile.openstreetmap.org/19/84253/203324.png
+        // So the pixel coordinate in a map at zoom level 19 (resolution 256*2^19) is this:
+        let test_coordinate_web_mercator_zoomed_truth =
+            Vector2::new(84253.0 * 256.0 + 165.0, 203324.0 * 256.0 + 18.0);
+        let test_coordinate_web_mercator_zoomed =
+            WebMercatorCoord::from_lat_lng(&test_coordinate).to_zoomed_coordinate(19);
+        // This was from eyeballing, so we shouldn't expect more than 20px accuracy at zoom 19.
```
catevita, line 139: I am readin that level 19 has 0.3m per pixel accuracy at the equator, so 20 pixel would be kind of 6 meters. is it enough for the purpose of this conversion?

feuerste:
Did you consider testing https://github.com/Mange/googleprojection-rs for this?

src/math/web_mercator.rs
```diff
@@ -0,0 +1,146 @@
...
+
+    /// Convert the Web Mercator coordinate back to lat/lng.
+    ///
+    /// The altitude returned is always 0.
+    pub fn to_lat_lng(&self) -> WGS84<f64> {
+        let centered = self.xy - Vector2::new(0.5, 0.5);
+
+        let sin_term = (-centered.y * FOUR_PI).exp();
+
+        // Note that sin_term = -(2/(sin(y)-1)) - 1
```
catevita, line 58: is sin_term the sin_y computed in line 60? maybe unify the syntax

src/math/web_mercator.rs
nnmm: 6 m precision would be too low of course, but that's the precision to which I could measure with this test. I couldn't find a more precise location in the world, since OSM doesn't get deeper than that.

src/math/web_mercator.rs
nnmm: Oh true. Done

nnmm:
> Did you consider testing https://github.com/Mange/googleprojection-rs for this, in particular if [Mange/googleprojection-rs#3](https://github.com/Mange/googleprojection-rs/issues/3) is still a problem?

Yes, I saw that crate and was a bit turned off by it being a port of a port, which seems to be confirmed by that ticket (but I didn't try to reproduce that ticket). I'd also like to use the nalgebra vector types, which that crate doesn't provide.

src/math/web_mercator.rs
nnmm: I edited the comment.

src/math/web_mercator.rs
catevita: if you compared with the results of the common/data_structures/web_mercator of avsw and see if the results are comparable?

nnmm: Yes, they're even exactly the same for this test.

src/math/web_mercator.rs
```diff
@@ -0,0 +1,143 @@
+//! Calculations with Web Mercator coordinates and tiles.
+
+use nalgebra::Vector2;
+use nav_types::WGS84;
+use std::f64::consts::{FRAC_1_PI, PI};
+
+/// 2.0 * E.powf(PI).arctan() - FRAC_PI_2;
+/// In degrees, it's 85.051129 (cf. Wikipedia)
```
feuerste, line 8: Nit: Can you please link to the URL instead of cf. Wikipedia?

src/math/web_mercator.rs
```diff
@@ -0,0 +1,143 @@
...
+const LAT_BOUND_RAD: f64 = 1.484_422_229_745_332_4;
+
+/// LAT_BOUND_SIN = sin(LAT_BOUND_RAD)
+const LAT_BOUND_SIN: f64 = 0.996_272_076_220_75;
+
+const TWO_PI: f64 = 2.0 * PI;
+const FOUR_PI: f64 = 4.0 * PI;
+const FOUR_PI_INV: f64 = 0.25 * FRAC_1_PI;
```
feuerste, line 16: How about renaming this to `FRAC_1_FOUR_PI`?

src/math/web_mercator.rs
```diff
@@ -0,0 +1,143 @@
...
+/// stops us from going deeper.
+pub const MAX_ZOOM: u8 = 23;
+
+/// A Web Mercator coordinate. Essentially a position in a 2D map of the world.
+#[derive(Copy, Clone, Debug, PartialEq, PartialOrd)]
+pub struct WebMercatorCoord {
+    /// Implementation detail: This is normalized to [0, 1), so not zoom level 0.
+    /// This makes calculations a bit simpler.
+    xy: Vector2<f64>,
+}
```
feuerste, line 30: What do you think about just doing
```rust
pub struct WebMercatorCoord(Vector2<f64>);
```
or
```rust
pub struct WebMercatorCoord(pub Vector2<f64>);
```
depending on whether you want to make the inner vector public or not?

feuerste:
BTW, if we need stuff in `nav_types`, I can add it there, too

src/math/web_mercator.rs
nnmm: I don't have a preference worth mentioning here, but I like giving it a name as a means of additional documentation (I changed it to something more meaningful). If there's a reason a tuple struct would be a better fit I'm happy to change it though.

src/math/web_mercator.rs
nnmm: Done.

src/math/web_mercator.rs
nnmm: Done.

nnmm:
:dash:

PR #450, opened 2020-04-08
Replace ad-hoc "caching" structures for intersection testing with trait

nnmm:
As a motivation, let me summarize a few things about how octrees are queried:
* A query for data in a geometric primitive, like an Obb, is made
* The usual representation of that geometric primitive doesn't support efficient intersection testing, instead it needs to compute derived data (intersecting axes usually, but `Cell` for the `CellID`s in S2 queries)
* We do many intersection tests against different octree nodes, and the computed data should not get re-computed each time, instead it should get cached/reused.
* Then, once nodes are identified, the points in them are filtered using the `PointCulling` trait's `contains()` method.

Currently the caching is done in a pretty ad-hoc manner:
* Frustum and Obb currently define an extra struct that holds the separating axes.
* There is no caching for S2 and Aabb queries.
* AllPoints queries do not need caching.

With this change:
* The part about intersection testing is spilt off from point filtering, i.e. `PointCulling` no longer has an `intersect_aabb()` method.
* Instead, all the `PointLocation` variants implement a trait to _produce_ an intersection testing structure (it can be identical to itself) named `HasAabbIntersector`.
* The produced intersection testing structure or "intersector" can borrow from, or be identical to, the geometric primitive. It can be reused many times and implements the actual `IntersectAabb` trait.

Benefits:
* Better performance by caching S2 and Aabb queries
* We can now write generic code over all `PointLocation` variants which receives an `impl HasAabbIntersector`. Compare `nodes_in_location()` in octree/mod.rs.
* Capturing de-facto patterns like this with traits seems like a good idea because it forces new queries to follow them and it's cognitively easier: If you have something that implements `HasAabbIntersector`, you know you can search the octree nodes with it.

Drawbacks:
* We should be able to tell Rust that any `ConvexPolyhedron` implementor is also a `HasAabbIntersector` implementor, but I got some conflicting traits errors. Resolving this would save us a few boilerplate implementations.

### DIFF
--- a/src/geometry/s2_cell_union.rs
+++ b/src/geometry/s2_cell_union.rs
@@ -8,7 +8,7 @@ use crate::math::FromPoint3;
 use nalgebra::{Point3, RealField};
 use s2::{cell::Cell, cellid::CellID, region::Region};
 
-/// Checks for an intersection between a cell_union and a polyhedron.
+/// Checks for an intersection between a list of cells and a polyhedron.
 ///
 /// This is done by checking whether an cell in the cell union intersects
 /// a covering of the polyhedron with S2 cells.
--- a/src/geometry/s2_cell_union.rs
+++ b/src/geometry/s2_cell_union.rs
@@ -51,9 +51,10 @@ where
     }
 }
 
-impl<'a, S: RealField> HasAabbIntersector<'a, S> for CellUnion
+impl<'a, S> HasAabbIntersector<'a, S> for CellUnion
 where
     f64: From<S>,
+    S: RealField,
 {
     type Intersector = Vec<Cell>;
     fn aabb_intersector(&'a self) -> Self::Intersector {
--- a/src/geometry/s2_cell_union.rs
+++ b/src/geometry/s2_cell_union.rs
@@ -10,7 +10,7 @@ use s2::{cell::Cell, cellid::CellID, region::Region};
 
 /// Checks for an intersection between a list of cells and a polyhedron.
 ///
-/// This is done by checking whether an cell in the cell union intersects
+/// This is done by checking whether any cell in the list intersects
 /// a covering of the polyhedron with S2 cells.
 pub fn cells_intersecting_polyhedron<S>(
     cells: &[Cell],
--- a/src/geometry/s2_cell_union.rs
+++ b/src/geometry/s2_cell_union.rs
@@ -2,7 +2,7 @@
 pub use s2::cellunion::CellUnion;
 
 use crate::geometry::Aabb;
-use crate::math::base::PointCulling;
+use crate::math::base::{HasAabbIntersector, IntersectAabb, PointCulling};
 use crate::math::sat::ConvexPolyhedron;
 use crate::math::FromPoint3;
 use nalgebra::{Point3, RealField};
@@ -12,8 +12,8 @@ use s2::{cell::Cell, cellid::CellID, region::Region};
 ///
 /// This is done by checking whether an cell in the cell union intersects
 /// a covering of the polyhedron with S2 cells.
-pub fn cell_union_intersects_polyhedron<S>(
-    cell_union: &CellUnion,
+pub fn cells_intersecting_polyhedron<S>(
+    cells: &[Cell],
     polyhedron: &impl ConvexPolyhedron<S>,
 ) -> bool
 where
@@ -28,10 +28,7 @@ where
     let mut polyhedron_cell_union = CellUnion(polyhedron_corner_cells);
     polyhedron_cell_union.normalize();
     let rect = polyhedron_cell_union.rect_bound();
-    cell_union
-        .0
-        .iter()
-        .any(|cell_id| rect.intersects_cell(&Cell::from(cell_id)))
+    cells.iter().any(|cell| rect.intersects_cell(cell))
 }
 
 impl<S> PointCulling<S> for CellUnion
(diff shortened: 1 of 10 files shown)


src/iterator.rs
```diff
@@ -197,17 +169,38 @@ pub trait PointCloud: Sync {
...
-            (|culling| {
-                let mut filtered_iterator = FilteredIterator {
-                    culling,
-                    filter_intervals,
-                    node_iterator,
-                };
-                filtered_iterator.try_for_each(callback)
-            })
-        )
+        match &query.location {
```
catevita, line 172: i'd prefer the more compact use of the macro, why is it not anymore possible?

src/iterator.rs
```diff
@@ -30,41 +30,13 @@ impl PointLocation {
...
-            PointLocation::Frustum(frustum) => Box::new(CachedAxesFrustum::new(frustum.clone())),
-            PointLocation::Obb(obb) => Box::new(CachedAxesObb::new(obb.clone())),
+            PointLocation::Frustum(frustum) => Box::new(frustum.clone()),
+            PointLocation::Obb(obb) => Box::new(obb.clone()),
             PointLocation::S2Cells(cell_union) => Box::new(cell_union.clone()),
         }
     }
 }
 
-/// This macro is an alternative to `get_point_culling()`, to be used where
```
catevita, line 40: why has it been removed?

src/geometry/s2_cell_union.rs
```diff
@@ -12,8 +12,8 @@ use s2::{cell::Cell, cellid::CellID, region::Region};
 ///
```
catevita, line 12: nit cell_union -> cells

nnmm: Thanks, done.

src/iterator.rs
nnmm: It can be changed to make better use of the traits and be simpler. It's back now :)

src/geometry/s2_cell_union.rs
```diff
@@ -2,18 +2,18 @@
...
+use crate::math::base::{HasAabbIntersector, IntersectAabb, PointCulling};
 use crate::math::sat::ConvexPolyhedron;
 use crate::math::FromPoint3;
 use nalgebra::{Point3, RealField};
 use s2::{cell::Cell, cellid::CellID, region::Region};
 
-/// Checks for an intersection between a cell_union and a polyhedron.
+/// Checks for an intersection between a list of cells and a polyhedron.
 ///
 /// This is done by checking whether an cell in the cell union intersects
```
catevita, line 13: nit: cell_union -> list

src/iterator.rs
```diff
@@ -197,20 +192,34 @@ pub trait PointCloud: Sync {
...
 
+// TODO(nnmm): Instead of having this helper function, make stream_points_for_query_in_node
+// accept a T: PointCulling, so we can dispatch to this function directly
+fn stream<'a, T: PointCulling<f64> + Clone, F: FnMut(PointsBatch) -> Result<()>>(
+    intv: &'a HashMap<&'a str, ClosedInterval<f64>>,
+    itr: NodeIterator,
+    callback: F,
+    culling: &T,
+) -> Result<()> {
+    let culling: T = culling.clone();
```
catevita, line 214: this clone() would have been called in the macro in the previous implementation, what are the advantages of it? or was in the previous implementation due to the closure?

src/geometry/s2_cell_union.rs
nnmm: Thanks, done.

src/iterator.rs
nnmm: Good question. Previously we used the macro only in one place, where the culling was cloned. Now it's also used in another place, `nodes_in_location()` in `impl PointCloud for Octree`, because all the different match arms are now the same. And that place doesn't need to clone, so I thought I'd leave it up to the function to clone or not.

src/math/base.rs
```diff
@@ -1,10 +1,50 @@
...
+
+/// We use this trait to allow an indirection: The geometry itself does not need to be able to
+/// efficiently do intersection tests with AABBs, because the geometry (e.g. an OBB) should not need
+/// to store data to support intersection tests (like separating axes).
+/// The trait has a lifetime to support the intersector borrowing from self, or being identical to
+/// &Self.
+///
+/// Having `impl<'a, T: IntersectAabb<S>, S> HasAabbIntersector<'a, S> for T` and
+/// `impl<'a, S, T: ConvexPolyhedron<S>> HasAabbIntersector<'a, S> for T` would be nice, but results
+/// in conflicts.
```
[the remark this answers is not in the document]
nnmm: This one: https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=a30186c0de3429f05cbd3ad1249bcaf5
It's bogus since the traits are private to the crate.

src/math/base.rs
```diff
@@ -1,10 +1,50 @@
...
+pub trait HasAabbIntersector<'a, S: RealField> {
```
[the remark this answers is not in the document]
nnmm: My reasoning was that the implementor is not itself an AabbIntersector, it produces one.

src/math/base.rs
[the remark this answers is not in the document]
nnmm: We could avoid that if we got rid of the `S` generics and used `f64` everywhere.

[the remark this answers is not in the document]
feuerste: Ok, I see, thanks for the explanation.

PR #451, opened 2020-04-09
Add Web Mercator rect query

nnmm:
A Web Mercator rect is just a min and max corner on a Web Mercator map. The querying is implemented by extruding the rectangle's four corners along their altitude axis up and down, which results in a convex polyhedron.

Notes:
* The length of the edges in an intersector had to change from 6 to 12
* I only implemented the f64 version – using f32 for LatLng and ECEF is not that precise.
* This implementation is only really accurate up to a medium scale: If you define a Web Mercator rect that spans longitudes [-90°; 90°] it will totally break, but [even before then](https://app.diagrams.net?lightbox=1&highlight=0000ff&edit=_blank&layers=1&nav=1&title=Web%20Mercator%20Query#R5Vddb9sgFP01eUxkDPjjsUm7TpoqVcukbnuZqE1sOmIiTJqkv34Xh8SmdtVWS%2FrSlwQOcOGec%2B8Fj%2FBsub3WbFXeqJzLURjk2xG%2BHIUhQkkEfxbZ7ZE4dUChRe4mtcBcPHEHBg5di5zX3kSjlDRi5YOZqiqeGQ9jWquNP22hpL%2FrihW8B8wzJvvonchNuUcTGrT4Vy6K8rAzCtzIkh0mO6AuWa42HQhfjfBMK2X2reV2xqUl78DLft2XF0aPB9O8Mm9ZoHfrxZhMHy430c9fD%2Fj3w7c%2FbOysPDK5dg67w5rdgQGwAmRDZ7opheHzFcvsyAb0Bqw0Swk9BE1nimvDty%2BeER09h5DhasmN3sEUtwA7rlywoAPNm5b6KHVY2aGdJA5kTu7iaLplBBqOlHcQFL5OkNGCVYV8A0NaGWaEqqA7joLTUEbjxOeMDHA2QNnZGEteZwxSYWWbBhjiT8qam664FnAArrv4bQu%2Bh1sbN3i6EFLOlFS62RQvFoswywCvjVZ%2FeWckj%2B4jGsGIAuvCWCIbA%2FW%2BGgUThNOIxAnFNMZxQGJyIukImkTUVy9EkyTpCYjCAQURxhNyJhHTnog%2FSm6dU3JX8lwDz7aYarv3viTDr2S6sJPu%2BD383nCdMWOnRBJOPL23rcK2tC3UcCUoKOxhUCnbEVUm17ldbZqNFuvmkIfiA%2BEDi6y8HxFLz%2BLg%2FYHUCcbUi6QwJpTEYYoTuAgRJdFpImmMUjpJsRdKmLRQtxbQtB9KKZ7E9EyhhPp3zI1V%2FAo85rXpl1O%2BNb4cPs%2BVqvgzURzEpCgs5Rmw1shsORVwl1%2B4gaXI8xfrtFbrKufWk6AfAqfI9iTsZXtMJ7Sf7WQg2cNzlWuE3%2FAEqPIL%2B5ay3EpW1yIbuvl53ntKvcpNx2s64PUB01xCPj365oeocDvcKlGZlnpC%2FdSII99CrdY6425R9wn1zA4lvh0UJb4hYyug6RlqhDl6%2FR9akZ5W3xuy4ATNC%2BSz5BIJ%2FUQaeCd%2BbBLRnjB31%2FOEHG8wey99FnEw9cXBydnUgW77AbXPsvYzFF%2F9Aw%3D%3D). There are two ways to handle this: Restricting the size of the rects that we accept, and increasing the height of the polyhedron. As a compromise, we require rects to be smaller than one level 8 tile (1/256 of the total width or height of the map).


nnmm:
@feuerste please review!
Perhaps this query type could be beneficial for XRay sharding as well.

fyi @cmarschner @catevita

src/geometry/web_mercator_rect.rs
```diff
@@ -0,0 +1,140 @@
...
+            Unit::new_normalize(corners[7] - corners[3]), // SW edge
+        ]);
+
+        let mut face_normals = ArrayVec::new();
+        face_normals.push(Unit::new_normalize(edges[0].cross(&edges[8]))); // N face
+        face_normals.push(Unit::new_normalize(edges[1].cross(&edges[9]))); // E face
+        face_normals.push(Unit::new_normalize(edges[2].cross(&edges[10]))); // S face
+        face_normals.push(Unit::new_normalize(edges[3].cross(&edges[11]))); // W face
+        face_normals.push(Unit::new_normalize(edges[1].cross(&edges[0]))); // down face
+        face_normals.push(Unit::new_normalize(edges[5].cross(&edges[4]))); // up face
```
feuerste, line 86: As above, how about using `ArrayVec::from`?

src/geometry/web_mercator_rect.rs
```diff
@@ -0,0 +1,140 @@
...
+}
+
+impl WebMercatorRect {
+    /// Returns `None` when `z` is greater than [`MAX_ZOOM`](index.html#constant.max_zoom)
+    /// or when the coordinates are out of bounds for the zoom level `z`. It also returns `None`
+    /// when the rect is larger than
+    /// Rects crossing the ±180° longitude line are not supported at the moment.
+    pub fn new(min: Vector2<f64>, max: Vector2<f64>, z: u8) -> Option<Self> {
+        let north_west = WebMercatorCoord::from_zoomed_coordinate(nalgebra::inf(&min, &max), z)?;
+        let south_east = WebMercatorCoord::from_zoomed_coordinate(nalgebra::sup(&min, &max), z)?;
```
feuerste, line 30: Nit: Use `min.inf(&max)` and `min.sup(&max)` instead, as the `nalgebra::...` versions are deprecated?

src/math/base.rs
```diff
@@ -39,12 +39,20 @@ impl<S: RealField + Bounded> IntersectAabb<S> for CachedAxesIntersector<S> {
...
         }
     };
+    ($type:ty) => {
+        impl<'a> HasAabbIntersector<'a, f64> for $type {
+            type Intersector = CachedAxesIntersector<f64>;
+            fn aabb_intersector(&'a self) -> Self::Intersector {
+                self.intersector().cache_separating_axes_for_aabb()
+            }
+        }
+    };
```
feuerste, line 57: If we implemented `WebMercatorRect` for all float types, we wouldn't need this change, right? I think this would make the code a little cleaner....

src/math/sat.rs
```diff
@@ -233,8 +232,8 @@ mod tests {
                 Point3::new( 1.0,  1.0, -1.0),
                 Point3::new( 1.0,  1.0,  1.0),
             ],
-            edges: unit_vectors.clone(),
-            face_normals: unit_vectors.clone(),
+            edges: ArrayVec::from_iter(unit_vectors.clone()),
+            face_normals: ArrayVec::from_iter(unit_vectors.clone()),
```
feuerste, line 236: See above

src/math/sat.rs
```diff
@@ -248,8 +247,8 @@ mod tests {
                 Point3::new( 1.5,  1.5, -0.5),
                 Point3::new( 1.5,  1.5,  1.5),
             ],
-            edges: unit_vectors.clone(),
-            face_normals: unit_vectors.clone(),
+            edges: ArrayVec::from_iter(unit_vectors.clone()),
+            face_normals: ArrayVec::from_iter(unit_vectors.clone()),
```
feuerste, line 251: See above

src/math/sat.rs
```diff
@@ -263,8 +262,8 @@ mod tests {
                 Point3::new(-0.7, -0.7, -0.9),
                 Point3::new(-0.7, -0.7, -0.7),
             ],
-            edges: unit_vectors.clone(),
-            face_normals: unit_vectors.clone(),
+            edges: ArrayVec::from_iter(unit_vectors.clone()),
+            face_normals: ArrayVec::from_iter(unit_vectors),
```
feuerste, line 266: See above

nnmm:
The `from_iter()` and `push()` calls were introduced for the reason that edges and face_normals now have different types, see the change in sat.rs.

src/geometry/web_mercator_rect.rs
```diff
@@ -0,0 +1,140 @@
...
+use crate::math::base::{HasAabbIntersector, PointCulling};
+use crate::math::sat::{CachedAxesIntersector, ConvexPolyhedron, Intersector};
+use crate::math::web_mercator::WebMercatorCoord;
+use arrayvec::ArrayVec;
+use nalgebra::{Point3, RealField, Unit, Vector2};
+use nav_types::{ECEF, WGS84};
+use serde::{Deserialize, Serialize};
+
+/// The dead sea is at -413m, but we use a more generous minimum
+const MIN_ELEVATION_M: f64 = -1000.0;
```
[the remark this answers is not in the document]
nnmm: I don't see how it leads to overfetching, could you maybe explain with a drawing?

src/math/sat.rs
nnmm: `edges` and `face_normals` now differ in capacity (`ArrayVec<[T; 12]>` vs `ArrayVec<[T; 6]>`), and they both have more capacity than length (3, the number of elements in `unit_vectors`). So our options here are: 
* Have `unit_vectors` be an `ArrayVec<[T; 3]>` and use `from_iter()` for `edges` and  `face_normals`. That's pretty short and sweet, and has imo the best readability. Note we never need to mention an explicit capacity.
* Define two variables instead of one; `unit_vectors_edges` (an `ArrayVec<[T; 12]>`) and `unit_vectors_face_normals` (an `ArrayVec<[T; 6]>`), and `clone()` those. Maybe a little more efficient, but we shouldn't care about performance here obviously.
If there's a reason not to do it like this, I'm happy to reconsider, otherwise I prefer the present form. I can add comments as well of course.

src/geometry/web_mercator_rect.rs
nnmm: The inherent methods are not there yet in version 0.20.0. We can do that here and in Aabb when we switch to 0.21.0

src/geometry/web_mercator_rect.rs
```diff
@@ -0,0 +1,140 @@
...
+    north_west: WebMercatorCoord,
+    south_east: WebMercatorCoord,
+}
+
+impl WebMercatorRect {
+    /// Returns `None` when `z` is greater than [`MAX_ZOOM`](index.html#constant.max_zoom)
+    /// or when the coordinates are out of bounds for the zoom level `z`. It also returns `None`
+    /// when the rect is larger than
+    /// Rects crossing the ±180° longitude line are not supported at the moment.
+    pub fn new(min: Vector2<f64>, max: Vector2<f64>, z: u8) -> Option<Self> {
```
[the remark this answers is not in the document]
nnmm: Done.

src/geometry/web_mercator_rect.rs
nnmm: You're right, here the length matches the capacity. Done.

src/math/base.rs
nnmm: Yes.
Aside: In the long term, do you think having these be generic makes sense, or should we go towards making everything f64?

src/geometry/web_mercator_rect.rs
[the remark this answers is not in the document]
nnmm: Done.

nnmm:
@feuerste After taking a few steps towards implementing it and still getting a million error messages, I am unsure if making the rect generic is worth it:
* Having it be generic is not very useful at the moment, only `f64` is ever needed. We're only buying the ability to use `f128` in the future, not `f32` since there are `f64: From<S>` bounds at various points.
* Many things that would be constants in a `f64`-only version need to be calculated/converted at runtime.
* It makes the code more complex, with many `nalgebra::convert()` calls

What do you think?

src/geometry/web_mercator_rect.rs
feuerste: Ok, can you please leave a `TODO` here?

src/geometry/web_mercator_rect.rs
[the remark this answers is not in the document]
feuerste: When fetching points we use the corner points of the extended web mercator cell to determine s2 cells, right?
If the corner points extending towards the center of the earth are further away from the earth center by 500 m, we will cover less s2 cell area.

src/math/base.rs
feuerste: I think generics don't hurt here so much. Users can still decide on precision themselves... I would keep generic math stuff as generic as possible.

src/math/sat.rs
feuerste: Ah, I've overseen this before.

src/geometry/web_mercator_rect.rs
nnmm: It's already there.

src/math/base.rs
feuerste: As discussed in person, `f64` makes more sense.

src/geometry/web_mercator_rect.rs
```diff
@@ -0,0 +1,140 @@
...
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct WebMercatorRect {
+    north_west: WebMercatorCoord,
+    south_east: WebMercatorCoord,
+}
+
+impl WebMercatorRect {
+    /// Returns `None` when `z` is greater than [`MAX_ZOOM`](index.html#constant.max_zoom)
+    /// or when the coordinates are out of bounds for the zoom level `z`. It also returns `None`
+    /// when the rect is larger than
```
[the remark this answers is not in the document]
nnmm: Fixed the comment.

src/geometry/web_mercator_rect.rs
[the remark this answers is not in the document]
nnmm: You're right, changed to -500

nnmm:
I decided on a good bound for the rect size: a level 8 tile, i.e. 1/256 of the map, or 1.4°. Earth curvature has a not-too-strong effect there so that the polyhedron method still works – see the new test I added.

There was no reason not to support wraparound at the longitude boundary, so I added it.

I think all open feedback from the previous round has been addressed, please take another look!

src/geometry/web_mercator_rect.rs
```diff
@@ -151,4 +168,58 @@ mod tests {
...
+        // Use the local radius of curvature at the equator (i.e. semi_minor_axis²/semi_major_axis),
+        // see https://en.wikipedia.org/wiki/Ellipse#Curvature for the formula and
+        // https://en.wikipedia.org/wiki/World_Geodetic_System#A_new_World_Geodetic_System:_WGS_84
+        // for the values of the semi-major and semi-minor axes.
+        let lat_sagitta = 6335439.32 * (1.0 - (lat_diff / 2.0).cos());
+        // Here the equatorial radius can be used.
+        let lng_sagitta = 6378137.0 * (1.0 - (lng_diff / 2.0).cos());
+        // The result is about 480 for both.
+        assert!(lat_sagitta < 500.0);
+        assert!(lng_sagitta < 500.0);
```
feuerste, line 195: Would
```rust
        assert!(lat_sagitta < MIN_ELEVATION_M.abs());
        assert!(lng_sagitta < MIN_ELEVATION_M.abs());
```
make sense here?

nnmm: No, that is only coincidental.

PR #452, opened 2020-04-10
Add a test for PointCulling/ConvexPolyhedron consistency

nnmm:
We have two implementations of a containment predicate on points: One via the `PointCulling` trait, and one via the `ConvexPolyhedron` trait. They should behave the same.

This also fixes a bug in the tests where the Aabb was equal to the entire point cloud's bbox.

### DIFF
--- a/point_cloud_test/tests/main.rs
+++ b/point_cloud_test/tests/main.rs
@@ -80,7 +80,7 @@ where
     let (s2, oct, data) = setup_pointcloud(&args);
     let query = PointQuery {
         attributes: vec!["color"],
-        location: gen_location(data.clone()),
+        location: gen_location(data),
         ..Default::default()
     };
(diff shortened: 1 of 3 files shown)


point_cloud_test/tests/main.rs
```diff
@@ -66,14 +80,42 @@ where
     let (s2, oct, data) = setup_pointcloud(&args);
     let query = PointQuery {
         attributes: vec!["color"],
-        location: gen_location(data),
+        location: gen_location(data.clone()),
```
feuerste, line 83: Nit: Why do we need the `clone` now?

nnmm: Right, that was a leftover from another experiment

nnmm: Removed

nnmm:
Thanks for reviewing!

PR #455, opened 2020-04-14
Properly handle errors in `merge_xray_quadtrees`.

krzesi-mir:
Instead of using `unwraps/expects` use proper errors.


krzesi-mir:
Thanks, Marco!

PR #458, opened 2020-04-27
Monomorphize all geometry primitives to use f64

nnmm:
Why is this better?
* The generics buy the ability to instantiate code with f32 (f32 and f64 are the only float types in Rust). However, queries in global coordinates (anything in ECEF, Web Mercator tile, S2 cell) require f64 precision anyway.
* De facto everything is always f64 (see iterator.rs, including the query and point coordinate type), and changing it would require a bunch of refactoring anyway. This just makes it official :) 
* Code becomes nicer to read and maintain: Fewer trait bounds everywhere, no more `nalgebra::convert()` etc.
* More things can be constants instead of run-time values, like `std::f64::consts::PI` instead of `S::pi()`.
* Compile times should improve a bit.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2129,7 +2129,6 @@ dependencies = [
  "serde_derive",
  "structopt",
  "tempdir",
- "walkdir",
 ]
 
 [[package]]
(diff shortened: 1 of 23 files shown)


nnmm:
meow @feuerste

nnmm:
Btw, should I monomorphize the SAT code as well? Probably yes, right?

feuerste:
Yes

nnmm:
Thanks for reviewing!

PR #463, opened 2020-06-03
Revert to stable gl_generator.

feuerste:
#462 fixed clippy problems with the gl_generator by a temporary branch. However, since the branch is [not going to be merged into the gl repo](https://github.com/brendanzab/gl-rs/pull/521), we switch back to the original version and wrap it with `allow(clippy::manual_non_exhaustive)`.


feuerste:
Thanks for your review!

PR #466, opened 2020-10-01
Fix matrix initialization.

feuerste:
https://github.com/cartographer-project/point_cloud_viewer/pull/415/files#diff-fedc635549c49a1867cf6b18f648e6b8R158-R165 introduced a bug when initializing a matrix from all values. cgmath takes column major values, while nalgebra takes row major ones.


pdubroy (approved):

feuerste:
Thanks for your review!

PR #467, opened 2020-10-01
Fix console errors/warnings

feuerste:
This replaces #446 and fixes the errors in the debug console:
* Unable to preventDefault inside passive event listener
* Some three js warnings

### DIFF
--- a/octree_web_viewer/client/control.ts
+++ b/octree_web_viewer/client/control.ts
@@ -49,7 +49,8 @@ export class FirstPersonController {
       this.onKeyUp(<KeyboardEvent>event)
     );
     window.addEventListener('mousewheel', (event) =>
-      this.onMouseWheel(<WheelEvent>event)
+      this.onMouseWheel(<WheelEvent>event),
+      { capture: false, passive: false }
     );
     this.domElement.addEventListener('mousedown', (event) =>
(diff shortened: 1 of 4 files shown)


feuerste:
fyi @dudv

octree_web_viewer/client/control.ts
```diff
@@ -49,7 +49,8 @@ export class FirstPersonController {
       this.onKeyUp(<KeyboardEvent>event)
     );
     window.addEventListener('mousewheel', (event) =>
-      this.onMouseWheel(<WheelEvent>event)
+      this.onMouseWheel(<WheelEvent>event),
+      { capture: false, passive: false }
```
pdubroy, line 53: I would omit `capture: false` here if you can, since that is the default.

xray/client/control.ts
```diff
@@ -54,7 +54,7 @@ export class Maps2DController {
     window.addEventListener(
       'mousewheel',
       (event) => this.onMouseWheel(<WheelEvent>event),
-      false
+      { capture: false, passive: false }
```
pdubroy, line 57: Same, you can omit `capture` I'm pretty sure.

pdubroy (approved):

feuerste:
Thanks for your review!

PR #471, opened 2020-10-12
Update versions, remove grpc support.

feuerste:
* Updates all versions of third party dependencies to the latest ones.
* Also removes grpc support, as it is hard to maintain and isn't used anywhere to the best of my knowledge.


tblaette (approved):
LGTM, great cleanup!

feuerste:
Thanks for your review!

